PR #1, opened 2018-01-06
Add multiple build targets, for umd and es5.

Hinton:
Based upon https://github.com/alexjoverm/typescript-library-starter

VS Code and Visual Studio uses the commonjs library for their IntelliSense, while webpack prefers the es5 version.

As an added bonus, documentation gets generated using typedoc.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,0 @@
-tsconfig.json
-src
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -1,0 +1,35 @@
+// Based upon https://github.com/alexjoverm/typescript-library-starter/blob/master/rollup.config.ts
+
+import camelCase from 'lodash.camelcase';
+import commonjs from 'rollup-plugin-commonjs';
+import resolve from 'rollup-plugin-node-resolve';
+import sourceMaps from 'rollup-plugin-sourcemaps';
+
+const pkg = require('./package.json');
+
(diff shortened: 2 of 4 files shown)


.npmignore
```diff
@@ -1,2 +0,0 @@
-tsconfig.json
-src
```
kspearrin: we want to publish the src files to npm too?

.npmignore
```diff
@@ -1,7 +0,0 @@
-.vs
```
kspearrin: Why do we not want to ignore this stuff?

rollup.config.ts
```diff
@@ -0,0 +1,35 @@
...
+import sourceMaps from 'rollup-plugin-sourcemaps';
+
+const pkg = require('./package.json');
+
+const libraryName = 'index';
+
+export default {
+  input: `dist/es/${libraryName}.js`,
+  output: [
+    { file: pkg.main, name: camelCase(libraryName), format: 'umd' },
```
kspearrin: commelCase lib seems unnecessary?

.npmignore
Hinton: The `files` attribute in `package.json` tells npm to only include the `dist` directory.

rollup.config.ts
Hinton: Probably, I just copied the example.

PR #2, opened 2018-02-25
AuditService

Hinton:
Companion PR to https://github.com/bitwarden/browser/pull/526.

### DIFF
--- a/src/services/audit.service.ts
+++ b/src/services/audit.service.ts
@@ -7,7 +7,7 @@ export class AuditService {
     constructor(private cryptoService: CryptoService) {
     }
 
-    async passwordLeaked(password: string) {
+    async passwordLeaked(password: string): Promise<number> {
         const hash = (await this.cryptoService.sha1(password)).toUpperCase();
 
         const response = await fetch(this.url + hash.substr(0, 5));
@@ -21,6 +21,6 @@ export class AuditService {
                 return v.split(':')[0] === hashEnding;
             });
 
-        return match ? match.split(':')[1] : 0;
+        return match ? parseInt(match.split(':')[1], 10) : 0;
     }
 }
--- a/src/services/audit.service.ts
+++ b/src/services/audit.service.ts
@@ -1,0 +1,27 @@
+import { CryptoService } from 'jslib/abstractions/crypto.service';
+
+export class AuditService {
+
+    private url = 'https://api.pwnedpasswords.com/range/';
+
+    constructor(private cryptoService: CryptoService) {
+    }
+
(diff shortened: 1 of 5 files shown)


src/services/crypto.service.ts
```diff
@@ -452,6 +452,12 @@ export class CryptoService implements CryptoServiceAbstraction {
         return b64DecValue;
     }
 
+    sha1(password: string) {
```
kspearrin: I'd rather use web crypto for this to further reduce our dependency on Forge. Web crypto will turn this into a promise-based return. https://github.com/diafygi/webcrypto-examples#sha-1---digest

src/services/audit.service.ts
```diff
@@ -0,0 +1,27 @@
...
+    constructor(private cryptoService: CryptoService) {
+    }
+
+    async passwordLeaked(password: string) {
+        const hash = this.cryptoService.sha1(password);
+
+        const response = await fetch(this.url + hash.substr(0, 5));
+        const leakedHashes = await response.text();
+
+        const hashEnding = hash.substr(5);
```
kspearrin: use this higher up so you dont have to substring twice

src/services/audit.service.ts
```diff
@@ -0,0 +1,27 @@
+import { CryptoService } from 'jslib/abstractions/crypto.service';
+
+export class AuditService {
+
+    private url = 'https://api.pwnedpasswords.com/range/';
```
kspearrin: Should this be a constant outside the class?

src/services/audit.service.ts
Hinton: The `substr` calls affect different portions of the hash. The first call returns the first 5 characters, while the other call returns the remaining portion of the hash.

kspearrin: I see

src/services/crypto.service.ts
```diff
@@ -452,6 +452,17 @@ export class CryptoService implements CryptoServiceAbstraction {
         return b64DecValue;
     }
 
+    async sha1(password: string): Promise<string> {
+        const hash = await Crypto.subtle.digest(
+            {
+                name: 'SHA-1',
+            },
+            UtilsService.fromB64ToArray(window.btoa(password)),
```
kspearrin: Why not the utf-8 to array util?

Hinton: Whops, nice catch. Was experimenting why the result from `fromBufferToUtf8` threw errors. (Needed to convert it to Hex, not Utf8)

PR #3, opened 2018-04-15
Mix entropy with RNG for making floor on entropy.

dabura667:
https://www.reddit.com/r/Bitwarden/comments/8caa9d/whitepaper/dxe07m8/

/u/kinoshitajona has a good idea.

## Scenario 1: CSPRNG is bad (like Android bug in 2013) but Master Password is good
* Master Password and email are XOR so entropy is at least as good as master password on initial registration.

## Scenario 2: Master Password is bad
* Nothing changes, as XOR weak entropy with strong entropy will keep the strongest entropy.
* Many more problems outside this PR.

Notes: Perhaps quietly doing this is less good. Maybe saying "your initial registration master password will help protect the RNG in case of bug"... is better?

Also, Maybe Instead of This we should make the user move mouse / tap screen / etc. instead to create the mixin.

I'm sorry this is my first Typescript edit. Thanks.

### DIFF
--- a/src/angular/components/register.component.ts
+++ b/src/angular/components/register.component.ts
@@ -1,4 +1,5 @@
 import { Router } from '@angular/router';
+import * as crypto from 'crypto';
 
 import { ToasterService } from 'angular2-toaster';
 import { Angulartics2 } from 'angulartics2';
@@ -70,7 +71,8 @@ export class RegisterComponent {
     private async register() {
         this.email = this.email.toLowerCase();
         const key = this.cryptoService.makeKey(this.masterPassword, this.email);
-        const encKey = await this.cryptoService.makeEncKey(key);
+        const mixin = crypto.createHmac('sha512', this.email).update(this.masterPassword).digest();
+        const encKey = await this.cryptoService.makeEncKey(key, mixin);
         const hashedPassword = await this.cryptoService.hashPassword(this.masterPassword, key);
         const request = new RegisterRequest(this.email, hashedPassword, this.hint, encKey.encryptedString);
         await this.apiService.postRegister(request);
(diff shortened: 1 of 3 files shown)


src/angular/components/register.component.ts
```diff
@@ -1,4 +1,5 @@
 import { Router } from '@angular/router';
+import * as crypto from 'crypto';
```
kspearrin: this breaks in web clients like browser extension. node is not available

dabura667: What would be an alternative?

I had trouble following the Typescript. If there is a module with HMAC abilities somewhere in the library where would it be?

kspearrin: You would need to use an abstraction of crypto functions, which currently isn't available (will be pulled out of the crypto service where they are currently). It's something I started working on since it will be needed for our CLI and we can also switch the desktop app to.

dabura667: https://stackoverflow.com/a/47332317

Found this... hmmm

kspearrin: Yes, there are two WIP cryptoFunction services: https://github.com/bitwarden/jslib/blob/master/src/services/nodeCryptoFunction.service.ts and https://github.com/bitwarden/jslib/blob/master/src/services/webCryptoFunction.service.ts

mcfedr:
Seems like you are just adding the email address into the key, not adding any entropy as its completely predictable. All the (good) advice I've ever read is to use the system CSPRNG.

eliykat:
There hasn't been any further activity on this PR in a few years, so I'm closing it as stale.

@dabura667 if you are still interested in progressing this PR, please:

Otherwise, the [community forums](https://community.bitwarden.com/) might be a better place for general discussion.

PR #4, opened 2018-04-27
make src/services/api.service.ts work with https servers requiring client certificates

yxzzx:
Tested with firefox 52.7.3ESR

### DIFF
--- a/src/services/api.service.ts
+++ b/src/services/api.service.ts
@@ -76,6 +76,7 @@ export class ApiService implements ApiServiceAbstraction {
     async postIdentityToken(request: TokenRequest): Promise<IdentityTokenResponse | IdentityTwoFactorResponse> {
         const response = await fetch(new Request(this.identityBaseUrl + '/connect/token', {
             body: this.qsStringify(request.toIdentityToken(this.platformUtilsService.identityClientId)),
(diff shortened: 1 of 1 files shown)


kspearrin:
I am a little confused as to what this allows? What kind of credentials are being sent with these requests?

yxzzx:
My apache webserver is configured with:
`
SSLCACertificateFile ".../client-ca.pem"
SSLVerifyClient require
`
With this configuration the browser has to present a user certificate from its certificate store issued by a specific CA requested by the webserver. 
Without the proposed changes bitwarden will present the certificate only with chrome but not with firefox. Because of this the initial SSL-handshake will fail with firefox and access to the vault will not be granted.

kspearrin:
IS there a reason you closed this?

yxzzx:
Sorry, pressed wrong button :-( and didn't notice.

kspearrin:
Is this still needed now that ESR is @ Firefox 60?

yxzzx:
Doesn't work with firefox 60. BTW, it also doesn't work with edge (I just tried). Probably it doesn't work for anything but chrome. From what I have read, chrome does things different to the standard.

PR #5, opened 2018-06-13
Adding mask and toggle for Card Security Code

neilsb:
Required for [bitwarden/desktop issue #78](/bitwarden/desktop/issues/78)

### DIFF
--- a/src/models/view/cardView.ts
+++ b/src/models/view/cardView.ts
@@ -18,6 +18,10 @@ export class CardView implements View {
         // ctor
     }
 
+    get maskedCode(): string {
+        return this.code != null ? '••••' : null;
+    }
+
     get brand(): string {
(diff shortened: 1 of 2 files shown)


src/models/view/cardView.ts
```diff
@@ -18,6 +18,10 @@ export class CardView implements View {
         // ctor
     }
 
+    get maskedCode(): string {
+        return this.code != null ? '••••' : null;
```
kspearrin: Should we make the number of dots equal to `this.code.length`? On password we purposely don't do that to mask the length as well, but I think it's probably ok here since it's probably either 3 or 4.

neilsb: Yup - updated that to show same number of dots, whatever the length.

PR #7, opened 2018-07-18
Add sequentialize to prevent parralel loading of cipher keys

mcfedr:
Massively improves start up performance of extensions. Fixes https://github.com/bitwarden/browser/issues/655 for me. 

I see that `CipherString.decrypt` is being called 1000s of times for the same keys. This changed my start up time from ~10 minutes of 100% CPU to about 5 seconds.

### DIFF
--- a/spec/common/misc/sequentialize.spec.ts
+++ b/spec/common/misc/sequentialize.spec.ts
@@ -12,6 +12,20 @@ describe('sequentialize decorator', () => {
         expect(foo.calls).toBe(1);
     });
 
+    it('should call the function once for each instance of the object', async () => {
+        const foo = new Foo();
+        const foo2 = new Foo();
+        const promises = [];
+        for (let i = 0; i < 10; i++) {
+            promises.push(foo.bar(1));
+            promises.push(foo2.bar(1));
+        }
+        await Promise.all(promises);
+
+        expect(foo.calls).toBe(1);
+        expect(foo2.calls).toBe(1);
+    });
+
     it('should call the function once with key function', async () => {
         const foo = new Foo();
         const promises = [];
--- a/spec/common/misc/sequentialize.spec.ts
+++ b/spec/common/misc/sequentialize.spec.ts
@@ -1,0 +1,127 @@
+import { sequentialize } from '../../../src/misc/sequentialize';
+
+describe('sequentialize decorator', () => {
+    it('should call the function once', async () => {
+        const foo = new Foo();
+        const promises = [];
+        for (let i = 0; i < 10; i++) {
+            promises.push(foo.bar(1));
+        }
+        await Promise.all(promises);
+
+        expect(foo.calls).toBe(1);
+    });
+
+    it('should call the function once with key function', async () => {
+        const foo = new Foo();
+        const promises = [];
+        for (let i = 0; i < 10; i++) {
+            promises.push(foo.baz(1));
+        }
+        await Promise.all(promises);
+
+        expect(foo.calls).toBe(1);
+    });
+
+    it('should call the function again when already resolved', async () => {
+        const foo = new Foo();
+        await foo.bar(1);
+        expect(foo.calls).toBe(1);
+        await foo.bar(1);
+        expect(foo.calls).toBe(2);
+    });
+
+    it('should call the function again when already resolved with a key function', async () => {
+        const foo = new Foo();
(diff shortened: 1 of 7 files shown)


kspearrin:
Hey, thanks for the PR. I've never done anything with decorators before so I am trying to understand what is going on here that is making things so much better. It looks like you're caching the args for this function or something? Can you explain a bit more what is going on here? This function already caches the resulting `decryptedValue`.

kspearrin:
I studied the current code a bit more after reading your comments again and I understand the problem now. Decrypting the vault after initial login results in many calls to get decryption keys. These keys are fetched from storage and are cached after their first get request, however, since the cipher decryption after unlocking the app is being done asynchronously this results in multiple "first calls" to get these keys from disk.

I still don't really understand your code here, but I gave it a try and it resulted my vault items to be incorrect. The same 2-3 items were being repeated multiple times in my vault listing (of 50 items total). I assume this is because it was fetching the same decryption cache multiple times.

Finally, I moved the vault decryption methods to be synchronous and I confirmed that there was only 1 "first call" to each get key method. I committed it here https://github.com/bitwarden/jslib/commit/51ee0b065abb965980e8fe398b568c015e557a2f

Can you try this and see how much better it is? Though it would still be nice if we could somehow make the decryption loop asynchronous, but I am unsure how we can do that and not still have the original issue. Maybe that's what you were attempting with this PR?

mcfedr:
so this usage of a [decorator](https://www.typescriptlang.org/docs/handbook/decorators.html) is to 'wrap' the function that its applied to with some other code. The API is a bit confusing, but in the end, the returned function `value` replaces the `decrypt` function whenever its called.

The decorator seemed the best way to implement this as it seemed likely that there are other places in the code where it would be useful, its similar to how you keep the result of the text in a property, but in a generic way that can be applied to any function.

Now you mention it, I'm also seeing repeated items with this change, I hadn't noticed this. I'll try and see if I can see whats causing it.

mcfedr:
I've worked out what the problem was, it was sharing cache for all instances of CipherString in one cache, not per instance. I've fixed this now and it correctly loads my vault.

I've also added sequentialize to functions that have internal caches, e.g. `CipherService.getAllDecrypted` - I see that this can be happening multiple times at the same time if I open and close the bitwarden extension popover whilst loading.

kspearrin:
The tests help, thanks. The problem I see with this is that we won't have control over when the cached data is cleared from memory. In the current implementation you can see that we clear caches when a vault is locked or a user logs out. This is important since the cache contains sensitive information.

mcfedr:
So I've purposely made this so that it doesn't do any caching and doesnt change the existing caching models.

Whilst the promise is being resolved, it is shared, but once its resolved its removed from the cache and future calls to the function will result in the function itself being called again.

That why its different from other decorators, like [`@memoize`](https://github.com/developit/decko#memoize) you might see that are caching the result of the function.

kspearrin:
@mcfedr I implemented this in the CryptoService  and it seems to work well.

One thing I am noticing is that the scope of the decorator seems to be a singleton, which means the decorator's cache could lead to memory leaks if used on objects that are not singleton as well. For example, when you had it on the CipherString decrypt, each time you lock and unlock the vault, the size of the cache doubles, since old CipherStrings are never released from that cache. I've removed it from the CipherString's decrypt function since that wasn't really helping us anyways. The problem was in the CryptoService with getting the keys.

mcfedr:
Awesome to see this merged. I think its worth looking at the `getAllDecrypted` calls, I think at least one of these gets called everything the extension icon is clicked, and these are functions that are caching the results anyway.

mcfedr:
The cache should be per instance of the object, thats what https://github.com/bitwarden/jslib/pull/7/files#diff-ff8ce314f7ebfa66f5c3a10afa065c7aR15 shows.

kspearrin:
@mcfedr That test only shows that the cached results are returned properly per instance of the object, but the cache map itself is bound to the scope of the decorator, which from my testing show it as a singleton to the lifetime of the application.

I don't think that there are any areas where `getAllDecrypted()` is called in an asynchronous manner, so they would not benefit.

mcfedr:
i see what you mean, the map will itself doesn't get cleared up. It would be possible to have a clear up method, that just checks if the map is empty after removing items from it

PR #8, opened 2018-08-02
feature/cli-18-import

subdavis:
argely duplicate code from the web vault, which will be re-factored.

Part 1 of bitwarden/cli#18 implementation.

### DIFF
--- a/src/services/import.service.ts
+++ b/src/services/import.service.ts
@@ -1,0 +1,229 @@
+import { ApiService } from '../abstractions/api.service';
+import { CipherService } from '../abstractions/cipher.service';
+import { FolderService } from '../abstractions/folder.service';
+import { I18nService } from '../abstractions/i18n.service';
+import { ImportService as ImportServiceAbstraction } from '../abstractions/import.service';
+
+import { ImportResult } from '../models/domain/importResult';
+
+import { CipherRequest } from '../models/request/cipherRequest';
+import { FolderRequest } from '../models/request/folderRequest';
+import { ImportCiphersRequest } from '../models/request/importCiphersRequest';
+import { KvpRequest } from '../models/request/kvpRequest';
+
+import { CipherView } from '../models/view/cipherView';
+
+import { AscendoCsvImporter } from '../importers/ascendoCsvImporter';
+import { AviraCsvImporter } from '../importers/aviraCsvImporter';
+import { BitwardenCsvImporter } from '../importers/bitwardenCsvImporter';
+import { BlurCsvImporter } from '../importers/blurCsvImporter';
+import { ChromeCsvImporter } from '../importers/chromeCsvImporter';
+import { ClipperzHtmlImporter } from '../importers/clipperzHtmlImporter';
+import { DashlaneCsvImporter } from '../importers/dashlaneCsvImporter';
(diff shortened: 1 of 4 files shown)


src/services/import.service.ts
```diff
@@ -0,0 +1,229 @@
...
+    message: string;
+    shouldReport: boolean;
+}
+
+export class ImportService implements ImportServiceAbstraction {
+    format: string = null;
+    fileContents: string;
+    formPromise: Promise<any>;
+
+    protected successNavigate: any[] = ['vault'];
```
kspearrin: not used. remove

src/services/import.service.ts
```diff
@@ -0,0 +1,229 @@
...
+export interface ISubmitResult {
+    success: boolean;
+    message: string;
+    shouldReport: boolean;
+}
+
+export class ImportService implements ImportServiceAbstraction {
+    format: string = null;
+    fileContents: string;
+    formPromise: Promise<any>;
```
kspearrin: Can be removed. Only used on projects that implement apiAction directive. just await

src/services/import.service.ts
```diff
@@ -0,0 +1,229 @@
...
+import { RoboFormCsvImporter } from '../importers/roboformCsvImporter';
+import { SafeInCloudXmlImporter } from '../importers/safeInCloudXmlImporter';
+import { SaferPassCsvImporter } from '../importers/saferpassCsvImport';
+import { SplashIdCsvImporter } from '../importers/splashIdCsvImporter';
+import { StickyPasswordXmlImporter } from '../importers/stickyPasswordXmlImporter';
+import { TrueKeyCsvImporter } from '../importers/truekeyCsvImporter';
+import { UpmCsvImporter } from '../importers/upmCsvImporter';
+import { ZohoVaultCsvImporter } from '../importers/zohoVaultCsvImporter';
+
+export interface ISubmitResult {
```
kspearrin: Is this really needed? Could you just return a string for success and reject with message on failure? Not sure what `shouldReport` is for though.

subdavis: I wasn't sure what to do here, actually.  In bitwarden/web angularlytics is called inline for certain scenarios.  In order to preserve that behavior, I really needed to pass an object back.

If you have a suggestion for how to more appropriately do this (or if you'd be ok with `any` here) let me know.  I'm pretty new to Typescript.

Thanks for the very quick review.

kspearrin: I would consider maybe just leaving some of those "precheck" errors (should report = false) in each client application rather than keeping them here.

subdavis:
Responded to comments.  Moved importOptions into this library as well, so that the command line can display a list of the supported import types.

subdavis:
Any other work needed to merge this?

kspearrin:
I was waiting for your PR on web. Are you planning to do that after this merges?

subdavis:
Yes, I was planning to do that after this and CLI merge.  We can do it all at once if you prefer.

kspearrin:
@subdavis I added import service to web vault here: https://github.com/bitwarden/web/commit/c37359cdfd8d64ea508412140b9d1ef3b55dda89

So no need to do that still from your end.

PR #10, opened 2018-09-24
update cipher getLastUsedForUrl to prioritize exact domain match

ijjk:
Currently, when using autofill and having multiple logins for different subdomains for the same base domain it defaults to the last used instead of the one that matches the current subdomain.

For example, if you have two logins app.domain.com and mail.domain.com and you last used mail.domain.com then visit app.domain.com it will autofill mail.domain.com's login details instead of app.domain.com's

This just adds a small check to see if there is an exact match and prioritizes it over the last used one. By the way, super awesome service it works great!

### DIFF
--- a/src/services/cipher.service.ts
+++ b/src/services/cipher.service.ts
@@ -45,6 +45,7 @@ import { UserService } from '../abstractions/user.service';
 
 import { sequentialize } from '../misc/sequentialize';
 import { Utils } from '../misc/utils';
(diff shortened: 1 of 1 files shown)


kspearrin:
This would be expected if you are using the default "base domain" matching rule. You could solve this by using the "host" matching rule. See https://help.bitwarden.com/article/uri-match-detection/

PR #12, opened 2018-10-06
Add a passphrase generation mechanism

mart-e:
This is an implementation of the feature request [Add an ability to generate diceware passphrases](https://community.bitwarden.com/t/add-an-ability-to-generate-diceware-passphrases/229/6).

This PR adds the required changes in jslib while bitwarden/web#267 implements the change in the web interface. I wait to see if this is accepted before working on the other implementations but the good news is that most of the changes should be in jslib.

It should also be backward compatible with cli that have not implemented the passphrase yet.

The word list used to generate the passphrase comes from the [diceware page](http://world.std.com/~reinhold/diceware.html) and is currently hardcoded in a `misc/ ` file. This is a first step but in a future contribution, I would like to, either have multiple lists (e.g. per language) or allow an import of a word list.

@kspearrin as said on the community website, this is a first draft. I am not an expert in typescript so don't hesitate to tell if anything is wrong in the code.
Let me know if you would like another architecture too.

Thanks for your project


mart-e:
Thanks for the quick review !

>  I'd like to explore lazy loading it when a user visits the password generator page itself. This is probably more of a task for me since you are probably not familiar with how the applications are compiled.

I agree, lazy loading seems a good option if we multiply the wordlists. If you give me some pointers I can give it a try but it may be easier for you indeed :)

mart-e:
Indeed, I have implemented it only in the web so far. I see @kspearrin has implemented it for the command line. But I don't think it has been implemented for the mobile app yet.

Extarys:
We could add the ability to insert numbers and uppercase letters (check Enpass).

Don't try that password, I didn't use that one :wink: 
Screenshot from 2019-06-29 00-12-52

mart-e:
@kspearrin what do you think of this proposal? I tend to prefer "simple" words but it is a way to add a bit of entropy. If you want, I can make a PR for that

kspearrin:
We added uppercase and numbers to passphrase generation in https://github.com/bitwarden/jslib/pull/43

Extarys:
:open_mouth: Thanks @kspearrin (and @mart-e ) I guess we just need to wait for the web and desktop app to add the GUI for it.

*excitement*

PR #13, opened 2018-10-07
Add importer for Passman

ServiusHack:
The importer reads JSON files generated by the Nextcloud app Passman.

The first tag is used as the folder name. This works well if passwords
have no or only one tag. If no username is set then the email address is
used as the username.

Files are not being imported.

### DIFF
--- a/src/importers/passmanJsonImporter.ts
+++ b/src/importers/passmanJsonImporter.ts
@@ -13,38 +13,38 @@ export class PassmanJsonImporter extends BaseImporter implements Importer {
         }
 
         results.forEach((credential: any) => {
-          if (credential.tags.length > 0) {
+            if (credential.tags.length > 0) {
                 const folderName = credential.tags[0].text;
                 this.processFolder(result, folderName);
-          }
+            }
 
-                const cipher = this.initLoginCipher();
-                cipher.name = credential.label;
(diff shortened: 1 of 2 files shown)


ServiusHack:
I've incorporated your suggestions and also ran `tslint` which found some minor things.

Here's the export of my test vault from passman: [passman-export.json.txt](https://github.com/bitwarden/jslib/files/2457283/passman-export.json.txt) It's `.txt` to make GitHub happy.

kspearrin:
Thanks. Is it possible to have a username and email in the same entry?

ServiusHack:
> Thanks. Is it possible to have a username and email in the same entry?

Yes, it is. Should I add the email as a note in that case?

kspearrin:
If both are set you can add the one that isnt used for username to the notes or custom fields.

ServiusHack:
Wow that merge was fast. :-) I accidentally already pushed, actually wanted to test this first. I'll let you know with a new pull request if there is any issue.

kspearrin:
There was one issue. I have a fix about to be pushed for it

kspearrin:
I committed my changes.

ServiusHack:
Awesome. Thanks a lot for your fast response to this.

PR #15, opened 2018-10-26
More prominent password exposed warning

Hinton:
Companion PR to https://github.com/bitwarden/desktop/pull/155

PR #16, opened 2018-10-26
Cache passwordLeaked

Hinton:
While working on #15, I noticed that we never cache the calls to HIBP which could cause some unnecessary network traffic. Did it as a separate PR since it's a good feature to have even without #15.


src/services/audit.service.ts
```diff
@@ -18,9 +21,13 @@ export class AuditService implements AuditServiceAbstraction {
         const hashStart = hash.substr(0, 5);
         const hashEnding = hash.substr(5);
 
-        const response = await fetch(new Request(PwnedPasswordsApi + hashStart));
-        const leakedHashes = await response.text();
-        const match = leakedHashes.split(/\r?\n/).find((v) => {
+        if (this.cachedPasswordLeaked[hashStart] == null) {
```
kspearrin: Should you use hash start + end for the key?

Hinton: This caches the whole response for hashStart which would prevent request starting with the same hash from requesting the same block. We could use the whole hash and just cache the number of leaks which might be a bit more memory efficient at the "remote" chance of doing some additional network request should the first 5 characters be the same for different hashes.

Hinton: I primarily did it this way to avoid storing a sha hash of the password in memory, which might be silly since the whole password will be loaded in memory either way.

kspearrin: That brings up a good point though. Since you are now holding a cache , that would need to be cleared on lock/logout like the others.

Hinton: Does it though? The only thing it currently stores is the beginning of the hash and the response back from HIBP. Which shouldn't be sensitive information.

Hinton: @kspearrin Added 24h max limit for caching. Wasn't certain if it should be time based or just invalidate on lock like the other caches. But since HIBP updates fairly rarely time-based should be fine, if slightly more complicated.

Hinton:
Closing old PRs.

PR #20, opened 2018-11-25
Add support for Steam TOTP

h44z:
Steam uses a different format for printing their TOTP codes.
This pr allows to use this format by prefixing the OTP key with steam://


kspearrin:
@h44z How do you get the steam secret key? Looking in my steam account settings now and don't see anything.

PR #21, opened 2018-11-25
Close to tray

h44z:
Allow the app to be minimized to the tray icon if the window gets closed. This is disabled by default.

Tested (and therefore enabled for linux) on Archlinux.

See desktop pull request: https://github.com/bitwarden/desktop/pull/168

### DIFF
--- a/src/electron/window.main.ts
+++ b/src/electron/window.main.ts
@@ -13,6 +13,7 @@ const Keys = {
 
 export class WindowMain {
     win: BrowserWindow;
+    isQuitting: boolean = false;
 
     private windowStateChangeTimer: NodeJS.Timer;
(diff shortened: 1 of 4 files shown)


src/electron/window.main.ts
```diff
@@ -221,4 +228,8 @@ export class WindowMain {
         return state != null && Number.isInteger(state.x) && Number.isInteger(state.y) &&
             Number.isInteger(state.width) && state.width > 0 && Number.isInteger(state.height) && state.height > 0;
     }
+
+    public quit() {
```
kspearrin: Is this function necessary? Looks like `isQuitting` is already public.

h44z: Function removed.

PR #25, opened 2019-01-28
 Throttle calls to HIBP api

mcfedr:
Randomly failing to check by passwords, I'm pretty sure its because ~2000 connections are made at the same time.

### DIFF
--- a/spec/common/misc/throttle.spec.ts
+++ b/spec/common/misc/throttle.spec.ts
@@ -1,0 +1,110 @@
+import { throttle } from '../../../src/misc/throttle';
+import { sequentialize } from '../../../src/misc/sequentialize';
+
+describe('throttle decorator', () => {
+    it('should call the function once at a time', async () => {
+        const foo = new Foo();
+        const promises = [];
+        for (let i = 0; i < 10; i++) {
+            promises.push(foo.bar(1));
+        }
+        await Promise.all(promises);
+
+        expect(foo.calls).toBe(10);
+    });
+
+    it('should call the function once at a time for each object', async () => {
+        const foo = new Foo();
+        const foo2 = new Foo();
+        const promises = [];
+        for (let i = 0; i < 10; i++) {
+            promises.push(foo.bar(1));
+            promises.push(foo2.bar(1));
+        }
+        await Promise.all(promises);
+
+        expect(foo.calls).toBe(10);
+        expect(foo2.calls).toBe(10);
+    });
+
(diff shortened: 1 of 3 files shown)


kspearrin:
Are the checks failing from the client or from the server?

Also, why are there changes to `sequentialize` mixed in here? Seems unrelated.

mcfedr:
The error shown in the console is `GET https://api.pwnedpasswords.com/range/94B83 net::ERR_CONNECTION_RESET 200`

The changes to `sequentialize` are slightly related :), because they are minor improvements to the decorator that I realised I could make whilst writing the new one.

mcfedr:
I also looked at catching the errors in the web project, but this just results in showing some of the results, and not the rest, which isnt very helpful

Editing the function `setCiphers` in `exposed-passwords-report.component.ts`

```typescript 
async setCiphers() {
        const allCiphers = await this.getAllCiphers();
        const exposedPasswordCiphers: CipherView[] = [];
        const promises: Array<Promise<void>> = [];
        const errs: Error[] = [];
        allCiphers.forEach((c) => {
            if (c.type !== CipherType.Login || c.login.password == null || c.login.password === '') {
                return;
            }
            const promise = this.auditService.passwordLeaked(c.login.password).then((exposedCount) => {
                if (exposedCount > 0) {
                    exposedPasswordCiphers.push(c);
                    this.exposedPasswordMap.set(c.id, exposedCount);
                }
            }).catch((err) => {
                errs.push(err);
                return;
            });
            promises.push(promise);
        });
        await Promise.all(promises);
        this.ciphers = exposedPasswordCiphers;
        if (errs.length) {
            // tslint:disable-next-line
            console.error(errs);
        }
    }
```

kspearrin:
I wouldn't worry about caching the results. Could you please move the changes to sequentialize into a new PR?

kspearrin:
Also, I tried this in Firefox on my vault of around 500 items and am not seeing any connection issues. I am guessing we can probably adjust the number higher than 50.

mcfedr:
> Also, I tried this in Firefox on my vault of around 500 items and am not seeing any connection issues. I am guessing we can probably adjust the number higher than 50.

Yes, not really sure what a reasonable number is, I have ~2000 items, and it sometimes runs successfully, but more often than not I get a failure, with this change it still runs in just a couple of seconds, given the browser limits for inflight requests, there probably isn't too much to be gained by having a really high limit.

PR #26, opened 2019-01-28
Clear sequentialize cache when empty to remove change of memory leaks

mcfedr:
Also use `finally` to prevent code duplication.


kspearrin:
@mcfedr This update just went out and seems to have broken users on Firefox pre-58 since `finally()` is not supported there. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

Is there a workaround here? Perhaps revert, or move to `.then(onFinally, onFinally)`?

```
const onFinally = () => {
    cache.delete(argsCacheKey);
    if (cache.size === 0) {
        caches.delete(this);
    }
};
response = originalMethod.apply(this, args).then(onFinally, onFinally);
```

Same issue for `throttle` decorator I imagine as well.

mcfedr:
Wow, sorry! Did not consider that case properly.

PR #29, opened 2019-02-19
Making base menu extensible

rafadc:
Changing base edit menu following the same pattern as in the view menu so we can add elements there.

See https://github.com/bitwarden/desktop/pull/216 for description of the PR


kspearrin:
I am nto sure why this had to be changed. Can't you just splice in the new menu items to the `editMenuItemOptions.subMenu` array?

rafadc:
Yes. I will close this PR and do that.

PR #30, opened 2019-03-11
feat(unpwnpassword): generate unpwned password and sync with server

raddatzk:
This pull request is a followup to [452](https://github.com/bitwarden/server/pull/452) on the server

With this pull request the process of generating a password changes a little.
By default the new password is validated against [haveibeenpwned](https://haveibeenpwned.com/) in order to only generate passwords, that are not currently exposed. 
Submitting the new cipher, independent of the previous result, also forces a check against haveibeenpwned and afterwards includes the information pwned and the pwnedCheckDate to the storing process in the database. 
There is also a new group _pwnedPasswords_ that gives a direct overview over all pwned passwords

### DIFF
--- a/src/angular/components/add-edit.component.ts
+++ b/src/angular/components/add-edit.component.ts
@@ -208,7 +208,16 @@ export class AddEditComponent implements OnInit {
                 this.collections.filter((c) => (c as any).checked).map((c) => c.id);
         }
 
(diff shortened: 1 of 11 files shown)


eliykat:
Please see my comment on the main PR on the server repo. Thanks!

PR #31, opened 2019-03-21
Fix appFlexCopy improperly trimming newlines and spaces

kleinfreund:
Related to https://github.com/bitwarden/browser/pull/897.

Triple-clicking a password to select it completely and invoking a platform-specific copy action (e.g. pressing Ctrl-C) currently copies the password with unrelated newlines and whitespaces.

This change assumes that a password always consists of only one line and removes everything after a newline (`\n` or `\r\n`).

### DIFF
--- a/src/angular/directives/flex-copy.directive.ts
+++ b/src/angular/directives/flex-copy.directive.ts
@@ -21,7 +21,13 @@ export class FlexCopyDirective {
         for (let i = 0; i < selection.rangeCount; i++) {
             const range = selection.getRangeAt(i);
             const text = range.toString();
-            copyText += text;
+
+            // The selection should only contain one line of text. In some cases however, the
+            // selection contains newlines and space characters from the identation of following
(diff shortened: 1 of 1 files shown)


src/angular/directives/flex-copy.directive.ts
```diff
@@ -21,7 +21,7 @@ export class FlexCopyDirective {
         for (let i = 0; i < selection.rangeCount; i++) {
             const range = selection.getRangeAt(i);
             const text = range.toString();
-            copyText += text;
+            copyText += text.substring(0, text.search(/\r?\n/));
```
kspearrin: How does this work if new line isn't found? Seems like it would substring(0, -1) which would be nothing. Did you test this?

kleinfreund: @kspearrin I’m sorry, of course. I’ll push an updated version of that commit to maintain the original behavior for lines not containing newlines at all.

PR #32, opened 2019-03-23
1password 1pif importer: create hidden fields

spiderpug:
When I migrated my 1Password archive into Bitwarden I noticed that concealed fields in 1Password are imported as regular "text" fields in Bitwarden. While the content itself is imported correctly it can do better by setting the FieldType to 'hidden'.

This PR adds this feature and related specs.

### DIFF
--- a/src/importers/onepassword1PifImporter.ts
+++ b/src/importers/onepassword1PifImporter.ts
@@ -169,11 +169,7 @@ export class OnePassword1PifImporter extends BaseImporter implements Importer {
                 }
             }
 
-            let fieldType = FieldType.Text;
-            if (field.k === 'concealed') {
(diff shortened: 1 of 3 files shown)


src/importers/onepassword1PifImporter.ts
```diff
@@ -168,8 +169,13 @@ export class OnePassword1PifImporter extends BaseImporter implements Importer {
                 }
             }
 
+            let fieldType = FieldType.Text;
```
kspearrin: i think you can make this a simple 1 liner

kspearrin:
Thanks!

kspearrin:
@spiderpug I deployed it to the production web vault if you want to use it now.

Issue #18, opened 2018-11-22
1password import misses a bunch of things


theontho:
**1password import does not import 1p identities as bitwarden identities**

I imported a 1pif file which contained several identities. They were imported as secure notes, even though bitwarden has an identity type. You can tell they are identities by looking this field : "typeName":"identities.Identity"

**1password imports trashed items with no differentiation**

The 1pif file does tag trashed items with a "trashed":true, tag. Maybe bitwarden can automatically put them in one category for easy categorization, such as a '1password import trashed items'. Another alternative would be to implement a trashbin feature and put those imported items into the new trashbin.

**1 password import does not import the password part of password items**

In a 1pif import "typeName":"passwords.Password" items do not have their passwords fields imported, but instead show up as weird stub items

"typeName":"webforms.WebForm do seem to get imported properly

This came from a 1password 6.8.9 export

**1password import does not import password history**
There are password history fields in 1pif items ("passwordHistory":), so it's something bitwarden could import too.

**1password import doesn't import attachments**
For example, I have pictures attached to my credit cards and id card items. 1 password export does export the attachments, but bitwarden import does not.

kspearrin:
identities and history are now imported per the above referenced PRs

saphirblanc:
What would be really, really useful is to import the tags as collections (if imported in an organization) or Folders if imported in a personal vault. Obviously, if the tags is written like "IT/Client/XXX", this should be the same in Bitwarden. Please :-), currently doing it by hand.

djsmith85:
Closing this now.

- Improvements to the existing 1pif importer have been made as mentioned by Kyle and can still be used.

- 1password has since replaced the `.1pif` with the new `.1pux` format. To also support this format, we have created a new importer. Details which cases it handles can be found at #594

PR #34, opened 2019-03-23
1password 1pif importer: create identity records

spiderpug:
Solves part of #18 

My export file is create by 1Password 6.

> 1password import does not import 1p identities as bitwarden identities
>
> I imported a 1pif file which contained several identities. They were imported as secure notes, even though bitwarden has an identity type. You can tell they are identities by looking this field : "typeName":"identities.Identity"

When no address is specified in 1Password, the address field is still exported, with blank strings in every subfield.

### DIFF
--- a/src/importers/onepassword1PifImporter.ts
+++ b/src/importers/onepassword1PifImporter.ts
@@ -5,6 +5,7 @@ import { ImportResult } from '../models/domain/importResult';
 
 import { CardView } from '../models/view/cardView';
 import { CipherView } from '../models/view/cipherView';
+import { IdentityView } from '../models/view/identityView';
 import { SecureNoteView } from '../models/view/secureNoteView';
 
 import { CipherType } from '../enums/cipherType';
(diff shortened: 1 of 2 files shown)


src/importers/onepassword1PifImporter.ts
```diff
@@ -167,6 +171,36 @@ export class OnePassword1PifImporter extends BaseImporter implements Importer {
...
+                } else if (this.isNullOrWhitespace(identity.company) && fieldDesignation === 'company') {
+                    identity.company = fieldValue;
+                    return;
+                } else if (this.isNullOrWhitespace(identity.email) && fieldDesignation === 'email') {
+                    identity.email = fieldValue;
+                    return;
+                } else if (fieldDesignation === 'address') {
+                    // fieldValue is an object casted into a string, so access the plain value instead
+                    const { street, city, country, zip } = field[valueKey];
+                    identity.address1 = street;
```
kspearrin: You said these are exported as blank strings if not provided. Should we perform `isNullOrWhitespace` checks on each here? ex: `identity.city = this.isNullOrWhitespace(city) ? null : city`. I'd hate to store an encrypted string of an empty field if that is all it is.

spiderpug: Sure. I assumed that's handled sometime later when preparing the payload as I couldn't "see" these properties in my vault's test record. I'll change it.

kspearrin: It might be, can’t remember.  but we can just handle it here to be sure.

kspearrin:
I'll need to check the opvault format on Windows to see how we can support this there too.

kspearrin:
I added windows opvault support and also the identity username field. Looks good.

PR #40, opened 2019-05-14
MacOS: Closing with red button won't open window again through tray icon

mchonan1996:
Fixed issue on MacOS where closing BW via the red button then reopening using tray icon wouldn't work.

See the issue on `bitwarden/desktop` here: [Issue 143](https://github.com/bitwarden/desktop/issues/143)

The issue was that when closing the window via the red circle, the `win` property in `windowMain` becomes null, and the window instance needs to be recreated.

### DIFF
--- a/src/electron/tray.main.ts
+++ b/src/electron/tray.main.ts
@@ -121,14 +121,17 @@ export class TrayMain {
     }
 
     private toggleWindow() {
-        if (this.windowMain.win == null) {
-            return;
-        }
-
-        if (this.windowMain.win.isVisible()) {
-            this.windowMain.win.hide();
+        if (this.windowMain.win === null) {
+            // On MacOS, closing the window via the red button destroys the BrowserWindow instance.
+            this.windowMain.recreateWindow().then(() => {
+                this.windowMain.win.show();
+            });
         } else {
-            this.windowMain.win.show();
+            if (this.windowMain.win.isVisible()) {
+                this.windowMain.win.hide();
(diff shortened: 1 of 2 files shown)


src/electron/window.main.ts
```diff
@@ -81,7 +81,13 @@ export class WindowMain {
...
     }
 
-    private async createWindow() {
+    async recreateWindow(): Promise<void> {
+        if (this.win === null) {
+            await this.createWindow();
+        }
+    }
+
+    private async createWindow(): Promise<void> {
```
kspearrin: Couldn't you just make this public so that you can call it?

src/electron/tray.main.ts
```diff
@@ -121,14 +121,17 @@ export class TrayMain {
     }
 
     private toggleWindow() {
-        if (this.windowMain.win == null) {
-            return;
-        }
-
-        if (this.windowMain.win.isVisible()) {
-            this.windowMain.win.hide();
+        if (this.windowMain.win === null) {
```
kspearrin: Should we make this condition only for macos then?

mchonan1996: Will do, thanks for the feedback. :)

src/electron/window.main.ts
mchonan1996: Will do, thanks for the feedback. :)

mchonan1996:
Implemented some feedback by @kspearrin.

PR #41, opened 2019-05-31
add window option: always on top of other windows

Tobirexy:
Adds a checkbox option to the window menu which makes Bitwarden always stay on top of other windows. This is really useful when copying more than one thing to an other program because the window doesn't get minimized. This option is disabled by default.

See desktop pull request: [https://github.com/bitwarden/desktop/pull/257](https://github.com/bitwarden/desktop/pull/257)

image

### DIFF
--- a/src/electron/window.main.ts
+++ b/src/electron/window.main.ts
@@ -154,11 +154,9 @@ export class WindowMain {
 
     }
 
-    async alwaysOnTop(){
-
-        this.enableAlwaysOnTop = this.win.isAlwaysOnTop() ? false : true;
-        this.win.setAlwaysOnTop(this.enableAlwaysOnTop ? true : false);
-
+    async toggleAlwaysOnTop(){
+        this.enableAlwaysOnTop = !this.win.isAlwaysOnTop();
+        this.win.setAlwaysOnTop(this.enableAlwaysOnTop);
         await this.storageService.save(ElectronConstants.enableAlwaysOnTopKey, this.enableAlwaysOnTop);
     }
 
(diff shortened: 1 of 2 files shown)


src/electron/window.main.ts
```diff
@@ -147,6 +151,15 @@ export class WindowMain {
         this.win.on('move', () => {
             this.windowStateChangeHandler(Keys.mainWindowSize, this.win);
         });
+
+    }
+
+    async alwaysOnTop(){
+
+        this.enableAlwaysOnTop = this.win.isAlwaysOnTop() ? false : true;
```
kspearrin: I dont think you need the ternary operators here. Just use `this.enableAlwaysOnTop = !this.win.isAlwaysOnTop()`?

src/electron/window.main.ts
```diff
@@ -147,6 +151,15 @@ export class WindowMain {
...
+        this.win.setAlwaysOnTop(this.enableAlwaysOnTop ? true : false);
```
kspearrin: Same. Just use `this.win.setAlwaysOnTop(this.enableAlwaysOnTop)`, right?

src/electron/window.main.ts
```diff
@@ -147,6 +151,15 @@ export class WindowMain {
         this.win.on('move', () => {
             this.windowStateChangeHandler(Keys.mainWindowSize, this.win);
         });
+
+    }
+
+    async alwaysOnTop(){
+
```
kspearrin: remove empty line

src/electron/window.main.ts
```diff
@@ -147,6 +151,15 @@ export class WindowMain {
         this.win.on('move', () => {
             this.windowStateChangeHandler(Keys.mainWindowSize, this.win);
         });
+
+    }
+
+    async alwaysOnTop(){
```
kspearrin: Should this be called `toggleAlwaysOnTop()`?

PR #43, opened 2019-06-06
Add common password requirements option to passphrase generator

SvenvDam:
While the generated passphrases are generally secure enough without the need for uppercase characters and numbers, I ran into the issue that a lot of websites still require these to be present when creating/updating your password.
I've added an option to the passphrase generator which does the following:
- Capitalize the first letter of each word in a generated passphrase
- Insert a digit at a random place in the the passphrase

Passphrases generated when this option is selected often do meet all the requirements on passwords imposed by websites without any manual changes, and the broader character set does not hurt I suppose.

Should I also submit PR's to the other Bitwarden repos to add this option in the browser extension, apps, etc?

### DIFF
--- a/src/services/passwordGeneration.service.ts
+++ b/src/services/passwordGeneration.service.ts
@@ -25,7 +25,8 @@ const DefaultOptions = {
     type: 'password',
     numWords: 3,
     wordSeparator: '-',
-    commonRequirements: false,
+    capitalize: false,
+    includeNumber: false,
 };
 
 const Keys = {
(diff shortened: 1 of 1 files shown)


src/services/passwordGeneration.service.ts
```diff
@@ -25,6 +25,7 @@ const DefaultOptions = {
     type: 'password',
     numWords: 3,
     wordSeparator: '-',
+    commonRequirements: false,
```
kspearrin: I would prefer these two be independent options that a user could set from the UI. `capitalize` and `includeNumber` or something.

src/services/passwordGeneration.service.ts
```diff
@@ -171,12 +172,23 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
...
+            o.addCommonRequirements = false;
+        }
+
         const listLength = EEFLongWordList.length - 1;
-        const wordList = new Array(o.numWords);
+        let wordList = new Array(o.numWords);
         for (let i = 0; i < o.numWords; i++) {
             const wordIndex = await this.cryptoService.randomNumber(0, listLength);
             wordList[i] = EEFLongWordList[wordIndex];
         }
```
kspearrin: I would do capitalize as part of this existing loop. That way the words do not have to be iterated again.

src/services/passwordGeneration.service.ts
```diff
@@ -256,6 +268,16 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
         return result;
     }
 
+    private capitalizeFirstLetter(str: string) {
```
kspearrin: Just call it `capitalize`, which already means uppercasing the first letter in a string.

kspearrin:
I made one adjustment here. Instead of inserting the number in the middle of a readable word, I adjusted it to append the number to a random word in the list. See https://github.com/bitwarden/jslib/commit/8375f7381ad1d113fbf2fe2bd65e90ec4a1631b7

PR #46, opened 2019-07-03
fix bitwarden/desktop#186 App not responding when NOT opening from tray

Tobirexy:
fixes [https://github.com/bitwarden/desktop/issues/186](https://github.com/bitwarden/desktop/issues/186)


kspearrin:
What's the difference here? Does `show` work the same as `restore` in both cases where opening from tray vs taskbar?

Tobirexy:
> What's the difference here? 

I don't know but it does seem to fix the linked issue.

> Does `show` work the same as `restore` in both cases where opening from tray vs taskbar?

Yes, everything worked in my testing. Maybe you can confirm that.

PR #47, opened 2019-07-15
support up to 256 char passwords

graingert:
MS accounts now support up to 256 character passwords, and it's possible for companies to set a policy to raise the minimum length to 256 also


eliykat:
Hi @graingert, thanks for your submission! If you're still interested in progressing this PR, please:

As this PR is more than 12 months old, I'll close it for now. However, this is just housekeeping. We encourage our community members to get involved and make contributions, so you can comment here even after it's closed and I'll reopen it if required.

PR #48, opened 2019-09-03
Only show dev tool options while in development mode

DarkArc:
This if pulled in for bitwarden/desktop hides what I would consider developer only options when the client is not in dev mode.


kspearrin:
These are important for debugging issues in production builds of the app. We don't plan on removing them from the menu.

DarkArc:
Is there a reason keyboard shortcuts, or a launch flag would be insufficient? I feel like these options are confusing for non-technical people and simply clutter the menus for 99% of users.

PR #49, opened 2019-09-23
Make library consumable the public

jschatz1:
This PR makes the library itself consumable by others so they can use

```
npm install @bitwarden/jslib
```
and

```javascript
const CryptoService = require('@bitwarden/jslib').Services.CryptoService;
```
or the equivalent import statement. Specifically I would like to use the `CryptoService` .

I am not sure what version you would want this to be or if you want the changes to the package-lock.json.

For now I have published my fork https://www.npmjs.com/package/jschatz-bitwarden-jslib so I can use it myself. 😄

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1534,7 +1534,8 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true
+              "dev": true,
+              "optional": true
             },
             "aproba": {
               "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


kspearrin:
Currently, I don't have any plans to publish releases for this. All of our internal projects that consume this library do so through git submodules. I would recommend that you do the same.

jschatz1:
Ok sounds good. Thanks.

PR #50, opened 2019-10-13
Allow special characters selection for pwd generation

GeorgySerga:
Supporting part for the https://github.com/bitwarden/desktop/pull/327.
Should hot break existing code. If modification of `specialCharacterSet` is not supported by the client regular string with all the characters would be created.


Hinton:
Hi,

PR #53, opened 2019-12-20
Global history

ajayyy:
This adds global edit history to notes and passwords.

Today I decided to try switching to Bitwarden from LastPass. It worked out well, but I was really missing this one feature.

This method only requires changing this library's code and does not require any server changes. It simply adds the item type to the saved item. I can prepare a client side PR as well to change the text to say "Edit History" instead of "Password History". 

image

Resolves https://github.com/bitwarden/server/issues/33

### DIFF
--- a/src/models/view/passwordHistoryView.ts
+++ b/src/models/view/passwordHistoryView.ts
@@ -4,6 +4,7 @@ import { Password } from '../domain/password';
 
 export class PasswordHistoryView implements View {
     password: string = null;
+    type: string = null;
     lastUsedDate: Date = null;
 
     constructor(ph?: Password) {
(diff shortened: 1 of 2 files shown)


kspearrin:
This feature was built to store password history. It's a bit of a hack to change it as-is to store other type of data.

ajayyy:
I did it this way due to not being super familiar with the code-base yet, but I can make a full update if you would like. Which way do you think would be best:

1. Add a `type` variable to the current `PasswordHistory` endpoint
2. Make a new endpoint for this with a `type` variable

ajayyy:
@DanStevens Password history should work fine, this one just extends the password history to work with all fields (notes, etc.).

I personally use my fork with this change and it has been working great for me, but I understand why Bitwarden would want a complete rewrite that used a new history mechanism instead of prepending the `Field name: ` in front of the values that aren't passwords inside the normal password history.

Hinton:
Hi,

PR #57, opened 2020-01-15
Fix import of expiration date from LastPass

felipecassiors:
resolves https://github.com/bitwarden/web/issues/443

### DIFF
--- a/src/importers/lastpassCsvImporter.ts
+++ b/src/importers/lastpassCsvImporter.ts
@@ -166,11 +166,25 @@ export class LastPassCsvImporter extends BaseImporter implements Importer {
             if (typeParts.length > 1 && typeParts[0] === 'NoteType' &&
                 (typeParts[1] === 'Credit Card' || typeParts[1] === 'Address')) {
                 if (typeParts[1] === 'Credit Card') {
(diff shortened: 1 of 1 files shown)


felipecassiors:
This still needs testing that I don't know how to do.

src/importers/lastpassCsvImporter.ts
```diff
@@ -166,11 +166,23 @@ export class LastPassCsvImporter extends BaseImporter implements Importer {
...
-                    const mappedData = this.parseSecureNoteMapping<CardView>(extraParts, {
+                    let mappedData = this.parseSecureNoteMapping<CardView>(extraParts, {
                         'Number': 'number',
                         'Name on Card': 'cardholderName',
                         'Security Code': 'code',
+                        'Expiration Date': 'expMonth',
                     });
+
+                    // Split string 'January,2020' into two
+                    const tempExpDate = mappedData[0].expMonth.split(',');
```
kspearrin: Can `mappedData[0].expMonth` be null/undefined? Seems you need some null checking here, right?

felipecassiors: That's right. Done.

kspearrin:
https://github.com/bitwarden/jslib/pull/65 was merged, which builds from this.

PR #62, opened 2020-02-01
fix `fullAddressPart2` state if empty should not show in address line

casz:
we should not always append state as some countries do not have states in the addresses.

### DIFF
--- a/src/models/view/identityView.ts
+++ b/src/models/view/identityView.ts
@@ -110,11 +110,11 @@ export class IdentityView implements View {
         const city = this.city || '-';
         const state = this.state;
         const postalCode = this.postalCode || '-';
-        let addressPart2 = city
+        let addressPart2 = city;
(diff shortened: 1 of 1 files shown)


src/models/view/identityView.ts
```diff
@@ -108,8 +108,13 @@ export class IdentityView implements View {
             return null;
         }
         const city = this.city || '-';
-        const state = this.state || '-';
+        const state = this.state;
         const postalCode = this.postalCode || '-';
-        return city + ', ' + state + ', ' + postalCode;
+        let addressPart2 = city
```
kspearrin: Please use line-ending semicolons.

casz: Good habits die hard 😅

casz:
Feel free to squash 🙇

casz:
Many thanks for the quick merge!

PR #65, opened 2020-02-05
fix lastpass import credit card expiration

figadore:
throwing my hat in the ring to fix https://github.com/bitwarden/web/issues/443
builds off #57 

first time using typescript. suggestions welcome

### DIFF
--- a/src/importers/lastpassCsvImporter.ts
+++ b/src/importers/lastpassCsvImporter.ts
@@ -166,22 +166,31 @@ export class LastPassCsvImporter extends BaseImporter implements Importer {
             if (typeParts.length > 1 && typeParts[0] === 'NoteType' &&
                 (typeParts[1] === 'Credit Card' || typeParts[1] === 'Address')) {
                 if (typeParts[1] === 'Credit Card') {
-                    let mappedData = this.parseSecureNoteMapping<CardView>(extraParts, {
+                    const mappedData = this.parseSecureNoteMapping<CardView>(extraParts, {
                         'Number': 'number',
                         'Name on Card': 'cardholderName',
                         'Security Code': 'code',
+                        // LP provides date in a format like 'June,2020'
+                        // Store in expMonth, then parse and modify
                         'Expiration Date': 'expMonth',
                     });
 
-                    // Split string 'January,2020' into two
-                    const tempExpDate = mappedData[0].expMonth.split(',');
-                    // Parse month name into number
-                    mappedData[0].expMonth = (
-                        new Date(
-                            Date.parse(tempExpDate[0] + '1, 2012')
-                        ).getMonth() + 1
-                    ).toString();
-                    mappedData[0].expYear = tempExpDate[1];
+                    if (mappedData[0].expMonth === ',') {
+                        // No expiration data
+                        delete mappedData[0].expMonth;
+                    } else {
+                        const [monthString, year] = mappedData[0].expMonth.split(',');
+                        // Parse month name into number
+                        if (monthString) {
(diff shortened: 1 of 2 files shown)


src/importers/lastpassCsvImporter.ts
```diff
@@ -170,7 +170,28 @@ export class LastPassCsvImporter extends BaseImporter implements Importer {
...
+                    } else {
+                        const [monthString, year] = mappedData[0].expMonth.split(',');
+                        // Parse month name into number
+                        if (monthString) {
+                            const month = new Date(Date.parse(monthString + '1, 2012')).getMonth() + 1;
+                            mappedData[0].expMonth = month.toString();
+                        } else {
+                            delete mappedData[0].expMonth;
+                        }
+                        if (year) {
```
kspearrin: same

src/importers/lastpassCsvImporter.ts
```diff
@@ -170,7 +170,28 @@ export class LastPassCsvImporter extends BaseImporter implements Importer {
                         'Number': 'number',
                         'Name on Card': 'cardholderName',
                         'Security Code': 'code',
+                        // LP provides date in a format like 'June,2020'
+                        // Store in expMonth, then parse and modify
+                        'Expiration Date': 'expMonth',
                     });
+
+                    if (mappedData[0].expMonth === ',') {
```
kspearrin: Is there ever a case where `mappedData[0].expMonth` is just an empty string, or null after `parseSecureNoteMapping`?

figadore: not in my tests. in  the `should parse blank card note` test i added, the data is from a note i exported from lastpass that literally had no fields filled out, and the result had `Expiration Date:,` as a line in the `extra` column

src/importers/lastpassCsvImporter.ts
figadore: updated

src/importers/lastpassCsvImporter.ts
```diff
@@ -170,7 +170,28 @@ export class LastPassCsvImporter extends BaseImporter implements Importer {
...
+                        // No expiration data
+                        delete mappedData[0].expMonth;
+                    } else {
+                        const [monthString, year] = mappedData[0].expMonth.split(',');
+                        // Parse month name into number
+                        if (monthString) {
```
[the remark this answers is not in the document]
figadore: updated

src/importers/lastpassCsvImporter.ts
kspearrin: Ok, but we still should guard against null reference exceptions here. Lastpass could change something in the future

src/importers/lastpassCsvImporter.ts
```diff
@@ -170,7 +170,29 @@ export class LastPassCsvImporter extends BaseImporter implements Importer {
                         'Number': 'number',
                         'Name on Card': 'cardholderName',
                         'Security Code': 'code',
+                        // LP provides date in a format like 'June,2020'
+                        // Store in expMonth, then parse and modify
+                        'Expiration Date': 'expMonth',
                     });
+
+                    const exp = mappedData[0].expMonth;
```
figadore: not too familiar with the codebase. can we rely on `mappedData[0]` to exist in this context?

src/importers/lastpassCsvImporter.ts
figadore: sure thing. updated

src/importers/lastpassCsvImporter.ts
figadore: looks like yes

PR #66, opened 2020-02-06
Lp import improvements

figadore:
Requesting feedback before digging into this task. My import experience was a little less than ideal once I found that it imported most of the fields as plain text in the 'notes' field.
I'm wondering about modifying the lastpass csv importer to read the remaining extra lines into custom fields or key value pairs, similar to https://github.com/bitwarden/jslib/blob/1859357ddbde2f55ae0ff033cdc9f4cf6cd0b4f6/src/importers/onepassword1PifImporter.ts#L245
maybe they'd all be text fields unless the keys contain certain keywords like `key` or `password`, in which case they'd be hidden.
I have some sample export data which I'd use as reference along the way and in the tests (see line 119). not planning on committing the csv itself
1. would this have any negative implications?
2. would this be worth merging if accomplished?
3. any other contribution tips, like will the commits be squashed it merge, or whether or not to force push, etc?

### DIFF
--- a/lastpass-export-examples.csv
+++ b/lastpass-export-examples.csv
@@ -1,0 +1,241 @@
+url,username,password,extra,name,grouping,fav
+http://password.com,frank,fr@nk,"my notes
+",test-import-password,Social,0
+http://sn,,,"some notes
+on multiple lines",test-import-note,Social,0
+http://sn,,,"NoteType:Address
+Language:en-US
+Title:mr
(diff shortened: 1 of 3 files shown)


kspearrin:
This should do it for you: https://github.com/bitwarden/jslib/commit/3c6f6dbe2f9c51400b7a8977c8b6492f1892ca9d

PR #68, opened 2020-02-08
improve lastpass imports

figadore:
* parse key-value pairs as custom fields for secure notes (previously only worked for all other types)
* infer field type from key name

tests currently do not pass because it does not handle multiline values. not sure if this is fixable, the only solution i've come up with so far is checking for `:`, which would work for lastpass' ssh key pair note type, _may_ be the only note type with a multiline field other than `notes`, but if this is not the case, then that logic would fail in other scenarios where `:` is a valid character. 

also not sure why the `public key` didn't get converted to a custom field, and shows up in `notes`

### DIFF
--- a/spec/common/importers/lastpassCsvImporter.spec.ts
+++ b/spec/common/importers/lastpassCsvImporter.spec.ts
@@ -33,7 +33,7 @@ Notes:some text
             folderId: null,
             name: 'Credit-card',
             notes: 'some text\n',
-            type: 3,
+            type: 3, //card
             card: {
                 cardholderName: 'John Doe',
                 number: '1234567812345678',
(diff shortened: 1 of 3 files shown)


eliykat:
Hi @figadore, thanks for your submission! If you're still interested in progressing this PR, please:

As this PR is more than 12 months old, I'll close it for now. However, this is just housekeeping. We encourage our community members to get involved and make contributions, so you can comment here even after it's closed and I'll reopen it if required.

figadore:
@eliykat I've signed the CLA, but I haven't checked whether the underlying code has changed since I submitted this

eliykat:
Thanks! I'll have a quick look at it as-is, just to confirm that it's acceptable in principle, before I ask you to fix any merge errors.

Hinton:
Hi,

PR #71, opened 2020-02-13
Add in ability to auto-logout after time-out #1143

lazear:
This pull request is paired with my pull request for the main browser plug in: https://github.com/bitwarden/browser/pull/1143

### DIFF
--- a/src/abstractions/lock.service.ts
+++ b/src/abstractions/lock.service.ts
@@ -7,5 +7,6 @@ export abstract class LockService {
     lock: (allowSoftLock?: boolean) => Promise<void>;
     setLockOption: (lockOption: number) => Promise<void>;
     isPinLockSet: () => Promise<[boolean, boolean]>;
+    isLogoutInstead: () => Promise<boolean>;
     clear: () => Promise<any>;
 }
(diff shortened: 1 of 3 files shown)


kspearrin:
Thanks for the PR. I believe this is going to cause inconsistencies with standard locking. For example, if you are using any of the lock options, the vault will lock whenever you restart the application/browser. Users will expect the same to happen here with logging out. The solution you have here will not log the user out whenever they restart the application.

lazear:
I did notice that on my end as well during testing. Should be feasible to implement that though, correct? If you're open to merging this (or similar functionality) once it's implemented properly I'd be willing to take a stab at it.

kspearrin:
Solving that problem will be a bit more involved. I think that it will come down to storing the tokens in memory rather than to disk.

I am also a bit torn as to whether we should do it this way, with a simple checkbox that plays off of the lock options, or if we should implement a separate logout timer so that you can use locking and logout together.

lazear:
I personally would be more in favor of having a separate timer for logout, so that you can set different durations for lock vs logout - the checkbox was just a way to get a quick prototype out. I can try and get a cleaner version with a separate logout service/timer up and running this weekend. Just wanted to get some initial feedback and see if the concept was acceptable.

vincentsalucci:
@lazear - I just wanted to let you know that we've taken your initial feedback here and are in the process of morphing it into a solution that is scalable for the product as a whole. Thanks for your contribution!

lazear:
@vincentsalucci That is great news, I'm glad you guys are gonna implement the feature - Thanks!

eliykat:
Closing this as the feature has now been implemented and the primary issue has also been closed.

PR #74, opened 2020-02-23
Add minimizeOnCopyToClipboardKey constant

elias-pap:


elias-pap:
This is needed for [PR #390](https://github.com/bitwarden/desktop/pull/390).

kspearrin:
Since these are desktop specific, they should probably go in src/electron/electronConstants

elias-pap:
> Since these are desktop specific, they should probably go in src/electron/electronConstants

Implemented with dbadaab.

src/angular/components/view.component.ts
```diff
@@ -184,6 +190,14 @@ export class ViewComponent implements OnDestroy, OnInit {
         event.dataTransfer.setData('text', data);
     }
 
+    public async minimizeIfNeeded(): Promise<void> {
```
kspearrin: As-is, this is going into other Bitwarden apps like the browser extension and web vault. This should all be in the desktop repo since it is desktop specific. You can overload the copy function.

elias-pap: Do you mean override the copy function ?

elias-pap:
I moved the constant to jslib of the desktop repo. We may close this one.

PR #75, opened 2020-02-25
Enforce Password Generator Policy

vincentsalucci:
## Code Changes
- **abstractions/passwordGeneration.service.ts**: Adjusted abstraction to mirror function calls in `passwordGeneration.service.ts`
- **password-generator.component.ts**: Utilized new `getOptions` function that injects intermediate step for password options retrieval. It applies the rules set by the governing organization to the user's password generation options. UI adjusted by using information stored in the `PasswordGeneratorPolicyOptions` object. Because field's floors were being altered, the `normalizeOptions` function was used to further adjust potential changes.
- **passwordGeneratorPolicyOptions.ts**: Created domain object for storage/retrieval of compiled password generator organization policies
- **passwordGeneration.service.ts**: Adjust `getOptions` function to utilize new `getPasswordGeneratorPolicyOptions` helper method and manipulate the user's options based on the organizations' enabled policy.

### DIFF
--- a/src/angular/components/password-generator.component.ts
+++ b/src/angular/components/password-generator.component.ts
@@ -8,6 +8,9 @@ import {
 import { I18nService } from '../../abstractions/i18n.service';
 import { PasswordGenerationService } from '../../abstractions/passwordGeneration.service';
 import { PlatformUtilsService } from '../../abstractions/platformUtils.service';
+import { PolicyService } from '../../abstractions/policy.service';
+import { PolicyType } from '../../enums/policyType';
+import { Policy } from '../../models/domain/policy';
 
 export class PasswordGeneratorComponent implements OnInit {
     @Input() showSelect: boolean = false;
@@ -17,15 +20,23 @@ export class PasswordGeneratorComponent implements OnInit {
     password: string = '-';
     showOptions = false;
     avoidAmbiguous = false;
+    enforcedMinLength = 0;
+    hasEnforcedUppercase = false;
+    hasEnforcedLowercase = false;
+    hasEnforcedNumbers = false;
+    enforcedNumberCount = 0;
+    hasEnforcedSpecial = false;
+    enforcedSpecialCount = 0;
 
     constructor(protected passwordGenerationService: PasswordGenerationService,
         protected platformUtilsService: PlatformUtilsService, protected i18nService: I18nService,
-        private win: Window) { }
+        protected policyService: PolicyService, private win: Window) { }
 
     async ngOnInit() {
         this.options = await this.passwordGenerationService.getOptions();
         this.avoidAmbiguous = !this.options.ambiguous;
         this.options.type = this.options.type === 'passphrase' ? 'passphrase' : 'password';
+        this.enforcePolicies(await this.policyService.getAll(PolicyType.PasswordGenerator));
         this.password = await this.passwordGenerationService.generatePassword(this.options);
         this.platformUtilsService.eventTrack('Generated Password');
         await this.passwordGenerationService.addHistory(this.password);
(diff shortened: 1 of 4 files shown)


src/angular/components/password-generator.component.ts
```diff
@@ -8,9 +8,7 @@ import {
 import { I18nService } from '../../abstractions/i18n.service';
 import { PasswordGenerationService } from '../../abstractions/passwordGeneration.service';
 import { PlatformUtilsService } from '../../abstractions/platformUtils.service';
-import { PolicyService } from '../../abstractions/policy.service';
-import { PolicyType } from '../../enums/policyType';
-import { Policy } from '../../models/domain/policy';
+import { PasswordGeneratorPolicyOptions } from '../../models/domain/passwordGeneratorPolicyOptions';
```
kspearrin:
```suggestion

import { PasswordGeneratorPolicyOptions } from '../../models/domain/passwordGeneratorPolicyOptions';
```

src/services/passwordGeneration.service.ts
```diff
@@ -217,7 +223,82 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
             }
         }
 
-        return this.optionsCache;
+        const enforcedPolicyOptions = await this.getPasswordGeneratorPolicyOptions();
+        if (this.optionsCache.length < enforcedPolicyOptions.minLength) {
```
kspearrin: Is there a way you can just return `null` here if there are no policies to enforce? Then you could just skip all of this condition checking.

src/services/passwordGeneration.service.ts
```diff
@@ -217,7 +223,82 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
...
+
+        if (policies == null || policies.length === 0) {
+            return enforcedOptions;
+        }
+
+        policies.forEach((currentPolicy) => {
+            if (!currentPolicy.enabled || currentPolicy.data == null) {
+                return;
+            }
+            const currentPolicyData = currentPolicy.data;
```
kspearrin: Not really necessary. Saving 1 character.

src/services/passwordGeneration.service.ts
```diff
@@ -217,7 +223,82 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
...
+
+            if (currentPolicyData.minLength != null
+                && currentPolicyData.minLength > enforcedOptions.minLength) {
+                enforcedOptions.minLength = currentPolicyData.minLength;
+            }
+
+            if (currentPolicyData.useUpper && !enforcedOptions.useUppercase) {
```
kspearrin: On these boolean ones, you can skip the `&& !enforcedOptions.useUppercase` again, like we did in the `optionsCache` in the method above. Just set it to true even if it is already true.

src/models/domain/passwordGeneratorPolicyOptions.ts
```diff
@@ -0,0 +1,50 @@
...
+export class PasswordGeneratorPolicyOptions extends Domain {
+    minLength: number = 0;
+    useUppercase: boolean = false;
+    useLowercase: boolean = false;
+    useNumbers: boolean = false;
+    numberCount: number = 0;
+    useSpecial: boolean = false;
+    specialCount: number = 0;
+
+    constructor(
```
kspearrin: Let's just ditch this ctor. Extra code that isn't being used. It's just more to maintain.

src/angular/components/password-generator.component.ts
kspearrin: Separate this from the section above with a new line. Everything else in this "Section" are services.

src/services/passwordGeneration.service.ts
vincentsalucci: Fixed in next update.

src/services/passwordGeneration.service.ts
vincentsalucci: Fixed in next update

src/models/domain/passwordGeneratorPolicyOptions.ts
vincentsalucci: Removed in next update

src/services/passwordGeneration.service.ts
vincentsalucci: Yes - added a few lines to both this function and the `getPasswordGeneratorPolicyOptions` to handle that and keep the UI layer happy at the same time.

PR #79, opened 2020-03-01
Enforce master password policy

vincentsalucci:
## Code Changes
- **abstractions/policy.service.ts**: Added the two new functions `getMasterPasswordPolicyOptions`and `evaluateMasterPassword`
- **domain/masterPasswordPolicyOptions**: Created object for storing the strictest master password policy options between organizations
- **services/policy.service.ts**: Added function to retrieve and compile the strictest master password policy options.  Added function that evaluates the supplied master password against the password strength in conjunction with the enforced policy options.

### DIFF
--- a/src/abstractions/policy.service.ts
+++ b/src/abstractions/policy.service.ts
@@ -1,6 +1,7 @@
 import { PolicyData } from '../models/data/policyData';
 
-import { Policy } from '../models/domain/policy';
+import { MasterPasswordPolicyOptions } from '../models/domain/masterPasswordPolicyOptions'
+import { Policy } from '../models/domain/policy'
 
 import { PolicyType } from '../enums/policyType';
 
@@ -11,4 +12,7 @@ export abstract class PolicyService {
     getAll: (type?: PolicyType) => Promise<Policy[]>;
     replace: (policies: { [id: string]: PolicyData; }) => Promise<any>;
     clear: (userId: string) => Promise<any>;
+    getMasterPasswordPolicyOptions: () => Promise<MasterPasswordPolicyOptions>;
+    evaluateMasterPassword: (passwordStrength: number, newPassword: string,
+        enforcedPolicyOptions?: MasterPasswordPolicyOptions) => Promise<boolean>;
 }
--- a/src/abstractions/policy.service.ts
+++ b/src/abstractions/policy.service.ts
@@ -14,5 +14,5 @@ export abstract class PolicyService {
     clear: (userId: string) => Promise<any>;
     getMasterPasswordPolicyOptions: () => Promise<MasterPasswordPolicyOptions>;
     evaluateMasterPassword: (passwordStrength: number, newPassword: string,
-        enforcedPolicyOptions?: MasterPasswordPolicyOptions) => Promise<boolean>;
+        enforcedPolicyOptions?: MasterPasswordPolicyOptions) => boolean;
 }
--- a/src/services/policy.service.ts
+++ b/src/services/policy.service.ts
@@ -5,6 +5,7 @@ import { UserService } from '../abstractions/user.service';
 import { PolicyData } from '../models/data/policyData';
 
 import { Policy } from '../models/domain/policy';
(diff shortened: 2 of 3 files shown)


src/services/policy.service.ts
```diff
@@ -52,4 +53,84 @@ export class PolicyService implements PolicyServiceAbstraction {
...
+
+            if (currentPolicy.data.requireSpecial) {
+                enforcedOptions.requireSpecial = true;
+            }
+        });
+
+        return enforcedOptions;
+    }
+
+    async evaluateMasterPassword(passwordStrength: number, newPassword: string,
```
kspearrin: I don't see any need for async here?

src/services/policy.service.ts
```diff
@@ -52,4 +53,84 @@ export class PolicyService implements PolicyServiceAbstraction {
...
+        enforcedPolicyOptions: MasterPasswordPolicyOptions): Promise<boolean> {
+        if (enforcedPolicyOptions == null) {
+            return true;
+        }
+
+        if (enforcedPolicyOptions.minComplexity > 0 && enforcedPolicyOptions.minComplexity > passwordStrength) {
```
kspearrin: Shouldn't this be `enforcedPolicyOptions.minComplexity >= passwordStrength`?

src/services/policy.service.ts
```diff
@@ -52,4 +53,84 @@ export class PolicyService implements PolicyServiceAbstraction {
...
+            return false;
+        }
+
+        if (enforcedPolicyOptions.minLength > 0 && enforcedPolicyOptions.minLength > newPassword.length) {
+            return false;
+        }
+
+        if (enforcedPolicyOptions.requireUpper && newPassword.toLowerCase() === newPassword) {
```
kspearrin: I don't think this works. We want to know if the password has any lowercase characters in it.

src/services/policy.service.ts
```diff
@@ -52,4 +53,84 @@ export class PolicyService implements PolicyServiceAbstraction {
...
+            return false;
+        }
+
+        if (enforcedPolicyOptions.requireLower && newPassword.toUpperCase() === newPassword) {
```
kspearrin: I don't think this works. We want to know if the password has any uppercase characters in it.

src/abstractions/policy.service.ts
```diff
@@ -1,6 +1,7 @@
 import { PolicyData } from '../models/data/policyData';
 
-import { Policy } from '../models/domain/policy';
+import { MasterPasswordPolicyOptions } from '../models/domain/masterPasswordPolicyOptions'
+import { Policy } from '../models/domain/policy'
```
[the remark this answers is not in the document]
vincentsalucci: For consistency, this is imported from `../models/domain/..` which is the same path as `Policy` and results in alphabetical ordering of the imports within the same grouping.  

Do you still want it separated by a new line?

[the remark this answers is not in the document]
kspearrin: I mis-read. You have it correct.

src/services/policy.service.ts
vincentsalucci: I believe you're correct. Fixing on updated PR.

src/services/policy.service.ts
vincentsalucci: No sir. I'm explicitly looking for the failure case only.
MinComplexity = 3
passwordStrength = 3

This is acceptable because we're setting a **minimum** complexity score which implies inclusivity. I only want it to fail if the minimum score is **greater than** the password strength.

src/services/policy.service.ts
vincentsalucci: We discussed this - the solution does work but we're going to use the locale flow to properly compare the letters in the correct location context.

src/services/policy.service.ts
vincentsalucci: We discussed this - the solution does work but we're going to use the locale flow to properly compare the letters in the correct location context.

src/services/policy.service.ts
kspearrin: Yup, I read that backwards.

PR #85, opened 2020-03-08
Password Generator Passphrase Policy

vincentsalucci:
## Code Changes
- **password-generator.component.ts**: Updated logic for defining a default type (user preference/password/passphrase). Updated `normalizeOptions` function to include raising the floor for the minimum number of words during passphrase generation
- **passwordGeneratorPolicyOptions.ts**: Added params for tracking passphrase policy options and default type selection and also added a helper method to retrieve whether or not the policy is in effect.
- **passwordGeneration.service.ts**: Updated `getPasswordGeneratorPolicyOptions()` function to instantiate/retrieve/set policy data options. Multi-org collision: password default type always wins (to mimic default options).   Updated `enforcePasswordGeneratorPoliciesOnOptions()` function to apply the enforced options.  Overrides default type selection if possible.

### DIFF
--- a/src/angular/components/password-generator.component.ts
+++ b/src/angular/components/password-generator.component.ts
@@ -31,13 +31,18 @@ export class PasswordGeneratorComponent implements OnInit {
         this.options = optionsResponse[0];
         this.enforcedPolicyOptions = optionsResponse[1];
         this.policyInEffect = this.enforcedPolicyOptions != null && (
+            this.enforcedPolicyOptions.allowPassword ||
             this.enforcedPolicyOptions.minLength > 0 ||
             this.enforcedPolicyOptions.numberCount > 0 ||
             this.enforcedPolicyOptions.specialCount > 0 ||
             this.enforcedPolicyOptions.useUppercase ||
             this.enforcedPolicyOptions.useLowercase ||
             this.enforcedPolicyOptions.useNumbers ||
-            this.enforcedPolicyOptions.useSpecial);
+            this.enforcedPolicyOptions.useSpecial ||
+            this.enforcedPolicyOptions.allowPassphrase ||
+            this.enforcedPolicyOptions.minNumberWords > 0 ||
+            this.enforcedPolicyOptions.capitalize ||
+            this.enforcedPolicyOptions.includeNumber);
         this.avoidAmbiguous = !this.options.ambiguous;
         this.options.type = this.options.type === 'passphrase' ? 'passphrase' : 'password';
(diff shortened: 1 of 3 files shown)


src/services/passwordGeneration.service.ts
```diff
@@ -314,9 +303,11 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
                 enforcedOptions = new PasswordGeneratorPolicyOptions();
             }
 
-            // Could be null if the policy options weren't adjusted with new data
-            if (currentPolicy.data.allowPassword != null && !currentPolicy.data.allowPassword) {
-                enforcedOptions.allowPassword = false;
+            // Password wins in multi-org collisions
+            if (currentPolicy.data.defaultType &&
+                currentPolicy.data.defaultType !== 'user' &&
```
kspearrin: user?

vincentsalucci: During policy edit I've added a dropdown selector for a default type. To maintain backwards compatibility and current use cases (user defines/selects the password TYPE within the generator) I've add a `User Preference` option so the org will not force a default type.

Screen Shot 2020-03-10 at 9 32 35 AM

kspearrin: Can we just handle that as `null`, meaning no default type is selected for this policy?

vincentsalucci: That can be done - just went the explicit way for easy understanding of the code.  `null` or `' '` should fail the initial condition (was looking for a truthy response - did I do that correctly?)

kspearrin: I would prefer to stay away from truthy evaluations if we can. Do explicit null check like `currentPolicy.data.defaultType != null`

vincentsalucci: Alrighty - will update.

src/angular/components/password-generator.component.ts
```diff
@@ -31,13 +31,17 @@ export class PasswordGeneratorComponent implements OnInit {
         this.options = optionsResponse[0];
         this.enforcedPolicyOptions = optionsResponse[1];
         this.policyInEffect = this.enforcedPolicyOptions != null && (
+            this.enforcedPolicyOptions.defaultType !== '' ||
```
kspearrin: Maybe we add an `inEffect(): boolean` function within this object that we just call that does all of this? Seems this is getting a bit out of hand.

vincentsalucci: Added a helper method to the enforced options object.

PR #86, opened 2020-03-11
Add noImplicitAny to tsc compiler options

mtgto:
To gain more security, we should use `--noImplicitAny` option of tsc.
https://www.typescriptlang.org/docs/handbook/compiler-options.html

Some spec files needs to be fixed.

desktop already enables `noImplicitAny`.
https://github.com/bitwarden/desktop/blob/master/tsconfig.json
cli enables too.
https://github.com/bitwarden/cli/blob/master/tsconfig.json


mtgto:
Thanks!

PR #87, opened 2020-03-11
Add the lint to CI script

mtgto:
Not to forget like #82, #70.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "clean": "rimraf dist/**/*",
     "build": "npm run clean && tsc",
     "build:watch": "npm run clean && tsc -watch",
-    "lint": "tslint src/**/*.ts spec/**/*.ts || true",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "clean": "rimraf dist/**/*",
     "build": "npm run clean && tsc",
     "build:watch": "npm run clean && tsc -watch",
-    "lint": "tslint src/**/*.ts spec/**/*.ts || true",
```
mtgto: I find the commit when `|| true` is added.
https://github.com/bitwarden/jslib/commit/1cebe27dbee59a789891ac7831ba458a79b9be02#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R18

I don't find the reason why it needs.

mtgto:
Thanks.

PR #89, opened 2020-03-18
Password Generator Sanitize Length

vincentsalucci:
## Code Changes
- **abstractions/passwordGeneration.service.ts**: Added `normalizeOptions` function
- **password-generator.component.ts**: Removed all non-application dependent logic from the `normalizeOptions` function.  Added call to the `passwordGenerationService.normalizeOptions`
- **passwordGeneration.service.ts**: Added 'normalizeOptions` function. Fully replicates existing logic except has added call to new `sanitizePasswordLength` function. Created `sanitizePasswordLength` function that can be called with UI changes or for password generation only.  Maintains existing workflows but will now properly update the UI if a minimum length condition is met.

### DIFF
--- a/src/services/passwordGeneration.service.ts
+++ b/src/services/passwordGeneration.service.ts
@@ -419,6 +419,87 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
         return result;
     }
 
+    normalizeOptions(options: any, enforcedPolicyOptions: PasswordGeneratorPolicyOptions) {
+        options.minLowercase = 0;
+        options.minUppercase = 0;
+
+        if (!options.length || options.length < 5) {
+            options.length = 5;
+        } else if (options.length > 128) {
+            options.length = 128;
+        }
+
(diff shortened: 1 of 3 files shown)


src/services/passwordGeneration.service.ts
```diff
@@ -473,4 +506,54 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
             [array[i], array[j]] = [array[j], array[i]];
         }
     }
+
+    private sanitizePasswordLength(options: any, forGeneration: boolean) {
+        let minUppercaseCalc: number = 0;
+        let minLowercaseCalc: number = 0;
```
kspearrin: You dont need to specify type here since you are direct assigning it as a number already.

```suggestion
        let minUppercaseCalc = 0;
        let minLowercaseCalc = 0;
```

vincentsalucci: Updated.

PR #91, opened 2020-03-26
[Auto-Logout] Refactor LockService and Update Dependencies

vincentsalucci:
## Code Changes
- **abstractions/index.ts**: Removed `LockService` and added `VaultTimeoutService`
- **abstractions/platformUtils.service.ts**: Added depricated decorator - All platforms return `null` for this function in their specific implementations. The only accessor to it is in `VaultTimeoutService` and the call is immediately followed by a call to storage to retrieve the value. @kspearrin Thoughts on best action moving forward for this function?
- **abstractions/vaultTimeout.service.ts**: Renamed `LockService` to `VaultTimeoutService`. Added/renamed methods necessary for carrying out auto-logout.
- **lock.component.ts**: Updated import dependency. Updated associated calls to the service.
- **auth-guard.service.ts**: Updated import dependency. Updated constructor. Updated associated call to the service.
- **constants.service.ts**: Updated `lockOptionKey` variable to `vaultTimeoutKey` and added `vaultTimeoutActionKey`.
- **crypto.service.ts**: Updated `ConstantsService` calls to reference new key.
- **services/index.ts**: Removed `LockService` and added `VaultTimeoutService`
- **notifications.service.ts**: Updated import dependency. Updated constructor. Updated associated call to the service.
- **system.service.ts**: Updated import dependency. Updated constructor. Updated associated call to the service.
- **services/vaultTimeout.service.ts**: Changed class name to `VaultTimeoutService`. Updated import dependency. Updated calls to new abstraction methods. Left TODOs for logic changes.


src/services/constants.service.ts
```diff
@@ -7,7 +7,8 @@ export class ConstantsService {
     static readonly disableFaviconKey: string = 'disableFavicon';
     static readonly disableAutoTotpCopyKey: string = 'disableAutoTotpCopy';
     static readonly enableAutoFillOnPageLoadKey: string = 'enableAutoFillOnPageLoad';
-    static readonly lockOptionKey: string = 'lockOption';
+    static readonly vaultTimeoutKey: string = 'vaultTimeout';
```
kspearrin: This will break people's existing saved settings. We have to be backwards compat here. We can change the variable name, but not the string value.

vincentsalucci: I've currently accounted for that within the sub-classed storageService class (specific to each platform).

The `init` call will look for the legacy `lockOption` key and use that value if a value for the new `vaultTimeout` key is not found (aka, they haven't saved it yet).

```
        // Handle legacy conversion to new storage key
        const legacyLockOption = await this.get<number>('lockOption');
        const vaultTimeout = await this.get<number>(ConstantsService.vaultTimeoutKey);
        if (vaultTimeout == null && !this.platformUtilsService.isDev()) {
            await this.save(ConstantsService.vaultTimeoutKey, legacyLockOption == null ? 15 : 
                  legacyLockOption);
        }
```

Can obviously do either way - just wanted to share what my thoughts were before confirming this change.

kspearrin: You will have to implement that on each platform, which is going to be a little messy. Let's just leave the old `lockOption` string for now. We can always clean up key strings later if we decide to.

vincentsalucci: Sounds good sir. Thanks!

src/abstractions/platformUtils.service.ts
```diff
@@ -14,6 +14,9 @@ export abstract class PlatformUtilsService {
     isMacAppStore: () => boolean;
     analyticsId: () => string;
     isViewOpen: () => Promise<boolean>;
+    /**
+     * @deprecated This only ever returns null. Pull from your platform's storage using ConstantsService.vaultTimeoutKey
+     */
```
kspearrin: is this true now, or will be based on your changes?

vincentsalucci: This seems to be true today. Investigated all implementations of the `PlatformUtils` interface and no usage returned any value besides `null`. 

The second-half of that comment to use the new `ConstantsService` key will be from here moving forward.

Will keep an eye on this as I dive into each platform to make sure that statement remains true.

PR #92, opened 2020-03-27
[Auto-Logout] Implement logout functionality in VaultTimeoutService

vincentsalucci:
## Code Changes
- **vaultTimeout.service.ts**
  - Updated constructor to include new `loggedOutCallback` for bubbling up when the platform in use should logout due to interval checks
  - `checkVaultTimeout` now has updated variable names and includes logic to pivot based on the saved action. 
  - `logout` uses the newly created callback to alert the application
  -  `setVaultTimeoutOptions` now saves the action as well

### DIFF
--- a/src/services/vaultTimeout.service.ts
+++ b/src/services/vaultTimeout.service.ts
@@ -107,7 +107,7 @@ export class VaultTimeoutService implements VaultTimeoutServiceAbstraction {
         }
     }
 
-    async logout(): Promise<void> {
+    async logOut(): Promise<void> {
         if (this.loggedOutCallback != null) {
             await this.loggedOutCallback();
         }
--- a/src/services/vaultTimeout.service.ts
+++ b/src/services/vaultTimeout.service.ts
@@ -80,7 +80,7 @@ export class VaultTimeoutService implements VaultTimeoutServiceAbstraction {
         if (diffSeconds >= vaultTimeoutSeconds) {
             // Pivot based on the saved vault timeout action
             await this.storageService.get<string>(ConstantsService.vaultTimeoutActionKey) === 'lock' ?
-                await this.lock(true) : await this.logout();
+                await this.lock(true) : await this.logOut();
         }
     }
 
--- a/src/services/vaultTimeout.service.ts
+++ b/src/services/vaultTimeout.service.ts
@@ -79,8 +79,8 @@ export class VaultTimeoutService implements VaultTimeoutServiceAbstraction {
         const diffSeconds = ((new Date()).getTime() - lastActive) / 1000;
         if (diffSeconds >= vaultTimeoutSeconds) {
             // Pivot based on the saved vault timeout action
-            await this.storageService.get<string>(ConstantsService.vaultTimeoutActionKey) === 'lock' ?
-                await this.lock(true) : await this.logOut();
+            const timeoutAction = await this.storageService.get<string>(ConstantsService.vaultTimeoutActionKey);
+            timeoutAction === 'lock' ? await this.lock(true) : await this.logOut();
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/services/vaultTimeout.service.ts
```diff
@@ -58,11 +60,13 @@ export class VaultTimeoutService implements VaultTimeoutServiceAbstraction {
             return;
         }
 
-        let lockOption = this.platformUtilsService.lockTimeout();
-        if (lockOption == null) {
-            lockOption = await this.storageService.get<number>(ConstantsService.vaultTimeoutKey);
+        // This has the potential to be removed. Evaluate after all platforms complete with auto-logout
+        let vaultTimeout = this.platformUtilsService.lockTimeout();
```
cscharf: Was lockTimeout() in platformUtilsService not updated with the new vaultTimeout naming?

vincentsalucci: It was not due to the fact that - as of right now - I've deprecated its usage because it is unused by all platform local implementations (only returns `null`). This usage here is the only use case (but as you can see its immediately followed by a null check and proper retrieval using the storage service.

I've left it here just in case I missed something. After I visit all platforms, I can come back and clean up this piece of code (and the `PlatformUtilsService` interface) as necessary. 

For a similar explanation see the comments [here](https://github.com/bitwarden/jslib/pull/91)

src/services/vaultTimeout.service.ts
```diff
@@ -71,12 +75,12 @@ export class VaultTimeoutService implements VaultTimeoutServiceAbstraction {
...
-        const lockOptionSeconds = lockOption * 60;
+        const vaultTimeoutSeconds = vaultTimeout * 60;
         const diffSeconds = ((new Date()).getTime() - lastActive) / 1000;
-        if (diffSeconds >= lockOptionSeconds) {
-            // need to lock now
-            await this.lock(true);
+        if (diffSeconds >= vaultTimeoutSeconds) {
+            // Pivot based on the saved vault timeout action
+            await this.storageService.get<string>(ConstantsService.vaultTimeoutActionKey) === 'lock' ?
+                await this.lock(true) : await this.logOut();
```
kspearrin:
```suggestion
            const timeoutAction = await this.storageService.get<string>(ConstantsService.vaultTimeoutActionKey);
            timeoutAction === 'lock' ? await this.lock(true) : await this.logOut();
```

src/services/vaultTimeout.service.ts
```diff
@@ -103,13 +107,15 @@ export class VaultTimeoutService implements VaultTimeoutServiceAbstraction {
...
+            await this.loggedOutCallback();
+        }
     }
 
     async setVaultTimeoutOptions(vaultTimeout: number, vaultTimeoutAction: string): Promise<void> {
         await this.storageService.save(ConstantsService.vaultTimeoutKey, vaultTimeout);
-        // TODO Add logic for vaultTimeoutAction
+        await this.storageService.save(ConstantsService.vaultTimeoutActionKey, vaultTimeoutAction);
         await this.cryptoService.toggleKey();
     }
```
kspearrin: Param names seem a little too verbose. I'd suggest just `timeout` and `action`. The method name already tells us this is for settings vault timeout.

vincentsalucci: Updated.

src/services/vaultTimeout.service.ts
vincentsalucci: Updated.  

What's the reason for this request (just curious)?

kspearrin: It was just a suggestion that kept the conditional check line a bit cleaner IMO

vincentsalucci: Fair and accurate. Thanks!

PR #94, opened 2020-04-03
[Auto-Logout] Update Token Service 

vincentsalucci:
## Code Changes 
- **abstractions/token.service.ts**: Added `toggleTokens` to interface
- **token.service.ts**: Updated `setTokens` to handle async. Updated `setToken` and `setRefreshToken` functions to not save to disk if a timeout it set and the action is log out. Added `toggleTokens` function for adjusting token storage based on change in timeout/action. Added helper function `skipTokenStorage`.
- **vaultTimeout.service.ts**: Added `tokenStorage` dependency. Added call to `toggleTokens` on saving of timeout/action.

### DIFF
--- a/src/services/token.service.ts
+++ b/src/services/token.service.ts
@@ -19,7 +19,7 @@ export class TokenService implements TokenServiceAbstraction {
     constructor(private storageService: StorageService) {
     }
 
-    async setTokens(accessToken: string, refreshToken: string): Promise<any> {
+    setTokens(accessToken: string, refreshToken: string): Promise<any> {
         return Promise.all([
             this.setToken(accessToken),
             this.setRefreshToken(refreshToken),
--- a/src/services/token.service.ts
+++ b/src/services/token.service.ts
@@ -21,8 +21,8 @@ export class TokenService implements TokenServiceAbstraction {
 
     async setTokens(accessToken: string, refreshToken: string): Promise<any> {
         return Promise.all([
-            await this.setToken(accessToken),
-            await this.setRefreshToken(refreshToken),
+            this.setToken(accessToken),
+            this.setRefreshToken(refreshToken),
         ]);
     }
 
--- a/src/services/token.service.ts
+++ b/src/services/token.service.ts
@@ -19,16 +19,22 @@ export class TokenService implements TokenServiceAbstraction {
     constructor(private storageService: StorageService) {
     }
 
-    setTokens(accessToken: string, refreshToken: string): Promise<any> {
+    async setTokens(accessToken: string, refreshToken: string): Promise<any> {
         return Promise.all([
-            this.setToken(accessToken),
-            this.setRefreshToken(refreshToken),
+            await this.setToken(accessToken),
+            await this.setRefreshToken(refreshToken),
         ]);
     }
 
-    setToken(token: string): Promise<any> {
+    async setToken(token: string): Promise<any> {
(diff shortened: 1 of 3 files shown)


src/services/token.service.ts
```diff
@@ -55,6 +67,23 @@ export class TokenService implements TokenServiceAbstraction {
         return this.refreshToken;
     }
 
+    async toggleTokens(): Promise<any> {
+        const token = await this.getToken();
+        const refreshToken = await this.getRefreshToken();
+        const timeout = await this.storageService.get(ConstantsService.vaultTimeoutKey);
+        const action = await this.storageService.get(ConstantsService.vaultTimeoutActionKey);
+        if ((timeout != null || timeout === 0) && action === 'logOut') {
```
vincentsalucci, line 75: @kspearrin I was trying to figure out why the check here also includes an explicit comparison for 0 while the check in `skipTokenStorage` is only for `!= null` : this is following the same pattern as `cryptoStorage`.

src/services/token.service.ts
```diff
@@ -19,16 +19,22 @@ export class TokenService implements TokenServiceAbstraction {
     constructor(private storageService: StorageService) {
     }
 
-    setTokens(accessToken: string, refreshToken: string): Promise<any> {
+    async setTokens(accessToken: string, refreshToken: string): Promise<any> {
         return Promise.all([
-            this.setToken(accessToken),
-            this.setRefreshToken(refreshToken),
+            await this.setToken(accessToken),
+            await this.setRefreshToken(refreshToken),
```
kspearrin, line 25: You don't need to await these. The purpose of `Promise.all` is to run things concurrently. These methods already return a promise.

src/services/token.service.ts
kspearrin: Yea, I am not sure... I guess it doesn't hurt to leave it here...

src/services/token.service.ts
```diff
@@ -183,4 +212,10 @@ export class TokenService implements TokenServiceAbstraction {
 
         return decoded.iss as string;
     }
+
+    private async skipTokenStorage(): Promise<boolean> {
+        const timeout = await this.storageService.get<number>(ConstantsService.vaultTimeoutKey);
+        const action = await this.storageService.get<string>(ConstantsService.vaultTimeoutActionKey);
+        return timeout != null && action === 'logOut';
```
kspearrin, line 219: Is timeout = null the "Never" case?

vincentsalucci: ^ That is correct!

vincentsalucci: Here is a breakdown of the values from each platform:

**Browser**
- `null` -> Never
- `0` -> Immediately

**Desktop**
- `null` -> Never

**Web Vault**
- `null` -> Never

src/services/token.service.ts
vincentsalucci: Updated.

src/services/token.service.ts
```diff
@@ -19,16 +19,22 @@ export class TokenService implements TokenServiceAbstraction {
     constructor(private storageService: StorageService) {
     }
 
-    setTokens(accessToken: string, refreshToken: string): Promise<any> {
+    async setTokens(accessToken: string, refreshToken: string): Promise<any> {
```
kspearrin, line 22: dont need async if no awaits

PR #95, opened 2020-04-07
Soft delete component and API updates in jslib

cscharf:
## Objective
Integrate the new API updates for soft delete (PUT /delete) and bulk operations as well as restore (PUT /restore) into the jslib common library for use by JavaScript based clients. This would integrate the fundamental tenants of:

- Items deleted now go to Trash (soft delete)
- Items in Trash when deleted are deleted permanently
- Items in Trash can be restored
- Searching items excludes soft deleted items in Trash unless you are viewing the Trash
- Delete/Restore operations follow the same pattern(s) as existing Delete operations (bulk/admin/org)
- New event type enum changes are propagated to the client

### DIFF
--- a/src/services/search.service.ts
+++ b/src/services/search.service.ts
@@ -71,7 +71,8 @@ export class SearchService implements SearchServiceAbstraction {
         console.timeEnd('search indexing');
     }
 
-    async searchCiphers(query: string, filter: (cipher: CipherView) => boolean = null, ciphers: CipherView[] = null):
+    async searchCiphers(query: string, filter: (cipher: CipherView) => boolean = null, ciphers: CipherView[] = null,
+        deleted: boolean = false):
         Promise<CipherView[]> {
         const results: CipherView[] = [];
         if (query != null) {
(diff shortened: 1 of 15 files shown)


src/services/search.service.ts
```diff
@@ -84,9 +85,16 @@ export class SearchService implements SearchServiceAbstraction {
         if (ciphers == null) {
             ciphers = await this.cipherService.getAllDecrypted();
         }
-        if (filter != null) {
-            ciphers = ciphers.filter(filter);
-        }
+
+        ciphers = ciphers.filter((c) => {
+            if (deleted !== c.isDeleted) {
```
kspearrin, line 90: Since this is a boolean comparison, you can just do `!=`, right?

src/services/search.service.ts
```diff
@@ -138,9 +146,12 @@ export class SearchService implements SearchServiceAbstraction {
         return results;
     }
 
-    searchCiphersBasic(ciphers: CipherView[], query: string) {
+    searchCiphersBasic(ciphers: CipherView[], query: string, deleted: boolean = false) {
         query = query.trim().toLowerCase();
         return ciphers.filter((c) => {
+            if (deleted !== c.isDeleted) {
```
kspearrin, line 152: same

PR #96, opened 2020-04-08
Soft delete - update cipher search and added restore functionality

cscharf:
Added the following changes with latest commit:
* **search.service.ts** - filter can be 1 or array of functions (or null or any combination, array with nulls, etc.)
  * removed deleted flag parameter for search function, ensure a deleted filter is included in call
* **components** - ensure restore functionality works, also ensure all deletes detect whether they should be soft or permanent deletes

> ## Objective
> Integrate the new API updates for soft delete (PUT /delete) and bulk operations as well as restore (PUT /restore) into the jslib common library for use by JavaScript based clients. This would integrate the fundamental tenants of:
> 
> * Items deleted now go to Trash (soft delete)
> * Items in Trash when deleted are deleted permanently
> * Items in Trash can be restored
> * Searching items excludes soft deleted items in Trash unless you are viewing the Trash
> * Delete/Restore operations follow the same pattern(s) as existing Delete operations (bulk/admin/org)
> * New event type enum changes are propagated to the client

### DIFF
--- a/src/angular/components/add-edit.component.ts
+++ b/src/angular/components/add-edit.component.ts
@@ -50,6 +50,7 @@ export class AddEditComponent implements OnInit {
     @Input() organizationId: string = null;
     @Output() onSavedCipher = new EventEmitter<CipherView>();
     @Output() onDeletedCipher = new EventEmitter<CipherView>();
+    @Output() onRestoredCipher = new EventEmitter<CipherView>();
     @Output() onCancelled = new EventEmitter<CipherView>();
     @Output() onEditAttachments = new EventEmitter<CipherView>();
     @Output() onShareCipher = new EventEmitter<CipherView>();
@@ -63,6 +64,7 @@ export class AddEditComponent implements OnInit {
     title: string;
     formPromise: Promise<any>;
     deletePromise: Promise<any>;
+    restorePromise: Promise<any>;
     checkPasswordPromise: Promise<number>;
     showPassword: boolean = false;
     showCardCode: boolean = false;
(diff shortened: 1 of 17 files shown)


src/angular/components/add-edit.component.ts
```diff
@@ -221,6 +223,10 @@ export class AddEditComponent implements OnInit {
     }
 
     async submit(): Promise<boolean> {
+        if (this.cipher.isDeleted) {
+            return this.restore();
+        }
```
kspearrin, line 228: I guess this is just a safety check? This shouldn't ever happen right?

src/angular/components/ciphers.component.ts
```diff
@@ -80,14 +80,15 @@ export class CiphersComponent {
         if (this.searchTimeout != null) {
             clearTimeout(this.searchTimeout);
         }
+        const deletedFilter: (cipher: CipherView) => boolean = (c) => c.isDeleted === this.deleted;
```
kspearrin, line 83: This is a little verbose to look at. We can let the type be inferred?

cscharf: I did try that initially, perhaps my inexperience with TypeScript (I did google it a bit) but the compiler would not leave it alone, it could infer the `[]` type, but not the singular type version :-(

src/angular/components/add-edit.component.ts
cscharf: No, this was when the user clicks on a cipher from the list in their vault, it opens the add-edit component modal, which only has a single button - Submit (labelled "Save") which I relabel "Restore" as the only viable option, and there's only one handler for onSubmit -> submit() so I'm hijacking that if the cipher is deleted to simply redirect to the restore action. I'm open to making 2 buttons on the html template with `*ngIf` or some other direction and different event handler if that appears like a cleaner implementation; Open to either approach.

cscharf:
I have tested this with the various downstream dependencies of jslib to ensure it's reverse compatible and haven't ran into any issues with the API changes (JS API/TS changes), at least at build-time and general run. Please pull this version and double check before merging into /master (sanity check).

src/angular/components/ciphers.component.ts
kspearrin: Weird.

PR #100, opened 2020-05-03
Remove tld checking

KazuAlex:
Fix https://github.com/bitwarden/web/issues/507

Uri doesn't systematically have a tld. Even more from the list defined in the regex "tldEndingRegex" (/.*\.(com|net|org|edu|uk|gov|ca|de|jp|fr|au|ru|ch|io|es|us|co|xyz|info|ly|mil)$/) in src/misc/utils.ts

Considere the uri as a "http" resource if the user doesn't provide a scheme.

### DIFF
--- a/src/models/view/loginUriView.ts
+++ b/src/models/view/loginUriView.ts
@@ -74,7 +74,7 @@ export class LoginUriView implements View {
 
     get isWebsite(): boolean {
         const uri = this.launchUri;
-        return uri != null &&
+        return uri !== null && uri !== '' &&
             (uri.indexOf('http://') === 0 || uri.indexOf('https://') === 0);
     }
 
@@ -82,7 +82,7 @@ export class LoginUriView implements View {
         if (this._canLaunch != null) {
             return this._canLaunch;
         }
-        if (this.uri != null && this.match !== UriMatchType.RegularExpression) {
+        if (this.uri !== null && this.uri !== '' && this.match !== UriMatchType.RegularExpression) {
             const uri = this.launchUri;
             for (let i = 0; i < CanLaunchWhitelist.length; i++) {
                 if (uri.indexOf(CanLaunchWhitelist[i]) === 0) {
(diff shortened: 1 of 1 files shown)


src/models/view/loginUriView.ts
```diff
@@ -95,6 +95,6 @@ export class LoginUriView implements View {
     }
 
     get launchUri(): string {
-        return this.uri.indexOf('://') < 0 && Utils.tldEndingRegex.test(this.uri) ? ('http://' + this.uri) : this.uri;
+        return this.uri.indexOf('://') < 0 ? ('http://' + this.uri) : this.uri;
```
Nokel81, line 98: This change should also be done to the `isWebsite` code above.

Nokel81:
However, I think that it probably would be a good idea to check the `hostname` if there is one. But I don't know the format of `this.uri`...

src/models/view/loginUriView.ts
KazuAlex: Absolutely !

KazuAlex:
> However, I think that it probably would be a good idea to check the `hostname` if there is one. But I don't know the format of `this.uri`...

There are a RFC about the uri format. But i don't know exactly the acceptations.
But i'm sure about that : checking the TLD is really a bad idea. For example if you use IPs or http://localhost as hostname.

Nokel81:
Fair I guess

clayadams5226:
We use GitHub issues as a place to track bugs and other development related issues. The [Bitwarden Community Forums](https://community.bitwarden.com) has a section for submitting, voting for, and discussing product [feature requests](https://community.bitwarden.com/c/feature-requests) like this one.

Please [sign up on our forums](https://community.bitwarden.com/signup) and search to see if this request already exists. If so, you can vote for it and contribute to any discussions about it. If not, you can re-create the request there so that it can be properly tracked.

This issue will now be closed. Thanks!

PR #101, opened 2020-05-07
Enable alternative ways for settings passwords

psiniemi:
* the environment variable BW_PASSWORD
* prefix the command line argument with "file:" and the password will read from the first line of that file
* prefix the command line argument with "env:" and the password will be read from that environment variable

I would like to avoid giving passwords on command lines. From the command line it would then be written into command histories and visible to other users and processes on the same machine.

### DIFF
--- a/src/cli/commands/login.command.ts
+++ b/src/cli/commands/login.command.ts
@@ -41,10 +41,10 @@ export class LoginCommand {
         }
 
         if (password == null || password === '') {
-            if (cmd.passwordFile) {
-                password = await NodeUtils.readFirstLine(cmd.passwordFile);
-            } else if (cmd.passwordEnv && process.env[cmd.passwordEnv]) {
-                password = process.env[cmd.passwordEnv];
+            if (cmd.passwordfile) {
+                password = await NodeUtils.readFirstLine(cmd.passwordfile);
+            } else if (cmd.passwordenv && process.env[cmd.passwordenv]) {
+                password = process.env[cmd.passwordenv];
             } else if (canInteract) {
                 const answer: inquirer.Answers = await inquirer.createPromptModule({ output: process.stderr })({
                     type: 'password',
--- a/src/cli/commands/login.command.ts
+++ b/src/cli/commands/login.command.ts
@@ -41,8 +41,10 @@ export class LoginCommand {
         }
 
         if (password == null || password === '') {
-            if (process.env.BW_PASSWORD) {
-                password = process.env.BW_PASSWORD;
+            if (cmd.passwordFile) {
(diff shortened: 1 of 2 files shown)


src/cli/commands/login.command.ts
```diff
@@ -38,18 +40,28 @@ export class LoginCommand {
...
             return Response.badRequest('Master password is required.');
         }
-
+        if (password.startsWith('file:')) {
+            const filename = password.substring(5);
+            password = await NodeUtils.readFirstLine(filename);
+        }
+        if (password.startsWith('env:') && process.env[password.substring(4)]) {
+            password = process.env[password.substring(4)];
+        }
```
kspearrin, line 64: What if someone's real password starts with "file:" or "env:"?

psiniemi: I would think that is rare, also as a workaround you can set the real value into a file or environment variable.

kspearrin: Could this be better handled with additional option flags?

psiniemi: This is similar to how openssl handles passwords, so it's not totally unfamiliar to people that work with security tools. Maybe a clearest option would be to have a group of optional option flags like --password --password-file and --password-env, but it would change the current command usage and I was trying to make this backwards compatible.

kspearrin: Adding new flags wouldn't be an issue with backwards compatibility. I like that option if it still solves the problem.

psiniemi: Ok, I'll try and come up with a solution like that.

psiniemi:
So I've made a solution with options and have also the corresponding code ready in directory-connector if this got approved.

src/cli/commands/login.command.ts
```diff
@@ -38,14 +40,21 @@ export class LoginCommand {
...
-            const answer: inquirer.Answers = await inquirer.createPromptModule({ output: process.stderr })({
-                type: 'password',
-                name: 'password',
-                message: 'Master password:',
-            });
-            password = answer.password;
+        if (password == null || password === '') {
+            if (cmd.passwordFile) {
+                password = await NodeUtils.readFirstLine(cmd.passwordFile);
+            } else if (cmd.passwordEnv && process.env[cmd.passwordEnv]) {
```
kspearrin, line 46: Can you change these to `passwordenv` and `passwordfile`. Looking at our other multi-word CLI options, everything is lowercase. Otherewise, looks good to go.

psiniemi: Done. Will mean that these are --paswordfile and --passwordenv instead of --password-file and --password-env in the command then.

PR #104, opened 2020-05-21
Add support for collections with hide passwords

Hinton:
This PR adds support for the hid passwords functionality added in https://github.com/bitwarden/server/pull/743.

### DIFF
--- a/src/models/data/cipherData.ts
+++ b/src/models/data/cipherData.ts
@@ -16,6 +16,7 @@ export class CipherData {
     folderId: string;
     userId: string;
     edit: boolean;
+    viewPassword: boolean;
     organizationUseTotp: boolean;
     favorite: boolean;
     revisionDate: string;
@@ -43,6 +44,7 @@ export class CipherData {
         this.folderId = response.folderId;
         this.userId = userId;
         this.edit = response.edit;
+        this.viewPassword = response.viewPassword;
         this.organizationUseTotp = response.organizationUseTotp;
         this.favorite = response.favorite;
         this.revisionDate = response.revisionDate;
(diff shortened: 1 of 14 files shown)


src/models/response/cipherResponse.ts
```diff
@@ -38,7 +39,12 @@ export class CipherResponse extends BaseResponse {
         this.name = this.getResponseProperty('Name');
         this.notes = this.getResponseProperty('Notes');
         this.favorite = this.getResponseProperty('Favorite') || false;
-        this.edit = this.getResponseProperty('Edit') || true;
+        this.edit = !!this.getResponseProperty('Edit');
+        if (this.getResponseProperty('ViewPassword') == null) {
```
cscharf, line 43: Given this will be a non null field in the DB, required through the server-side pipeline, would there ever be a case where this would be `null`?

Hinton: I added the check to ensure compatibility with older releases of the server. This will be null should the client connect to an old server which would cause `viewPassword` to be `false`, which is undesired.

src/models/data/cipherData.ts
```diff
@@ -43,6 +44,7 @@ export class CipherData {
         this.folderId = response.folderId;
         this.userId = userId;
         this.edit = response.edit;
+        this.viewPassword = response.viewPassword;
```
cscharf, line 47: Assuming this will need to change once changes are made to https://github.com/bitwarden/server/pull/743 per https://github.com/bitwarden/server/pull/743#discussion_r428747641

src/models/view/fieldView.ts
```diff
@@ -8,6 +8,7 @@ export class FieldView implements View {
     name: string = null;
     value: string = null;
     type: FieldType = null;
+    newField: boolean = false; // Marks if the filed is new and haven't been saved
```
cscharf, line 11: small spelling item in the comments: Marks if the **field** is new and **hasn't** been saved

PR #109, opened 2020-06-03
Reworked passphrase generation

Root-Core:
Features:
- Passphrase generation is now possible with a maximum length
- Passphrase generation is still possible with word count
- Passphrase generation can now force short words (Requested [here](https://community.bitwarden.com/t/make-the-passphrase-generator-use-shorter-words/12314))
- If the separator is null, don't force space, this allows CamelCase (Requested [here](https://community.bitwarden.com/t/allow-generating-passphrases-without-delimiter/8857) and [here](https://community.bitwarden.com/t/combine-passphrase-words-when-a-separator-isnt-used/10257))

Technical:
- Wordlist sorted by length and alphabet
- Enabling to access specific word lengths by offset


kspearrin:
Hey, thanks for the PRs. I haven't had a chance to review this in detail yet, but did want to point out one thing that that I saw. We cannot change the long word list. There are crypto operations with account fingerprint phrases that rely on the list is static and never changing. If it changes, people's account fingerprint phrases would change.

Root-Core:
You are right, totally missed that. Possible solutions are lookup tables, sorting algorithm or a separate list, which I would like to avoid.

The lookup table might be viable.

Root-Core:
@kspearrin I implemented a lookup function for the account fingerprint phrase. It creates the same passphrases at minimal overhead. Please let me know if something else is holding the PRs back.

JonathanGawrych:
This also resolves bitwarden/browser#1097, which I was about to open a PR for, but was happy to find one already exists 👍

djhanggi:
Is there a recent status on this? If there's a fundamental reason to not allowing limit-by-length, I wonder if displaying the password length in the generator would help in a subset of the cases (which would allow easily generating a number of passwords until a desired size is found).

cscharf (commented):
@kspearrin , after a careful review this looks good and the workaround is feasible to allow thumbprints in-tact. I'll leave it to you for one last review/pass-through.

src/services/passwordGeneration.service.ts
```diff
@@ -157,33 +158,70 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
...
-                wordList[i] = this.capitalize(EEFLongWordList[wordIndex]);
-            } else {
-                wordList[i] = EEFLongWordList[wordIndex];
+        // Words with up to 5 chars are considered short
+        const listLength = o.shortWords ? WordList.EEFLongWordListOffsets.Last_5_chars :
+                                          WordList.EEFLongWordListOffsets.Last;
+
+        const wordList = new Array<string>();
+
+        if (o.type === 'passphrase_limited') { // by char limiting
```
kspearrin, line 191: Could this just be an option of type passphrase? This doesn't seem like it should be a whole new type.

kspearrin (commented):
I am assuming we'll need to apply updates across all clients for these new options, as well as update mobile with the same logic changes. Also might need to update the password generator policy for new passphrase options.

Hinton:
Hi,

PR #111, opened 2020-06-10
add uriOptions array & input toggle func

mkanavakatini:
- Add array to hold uri options
- Add function to toggle uri input mode

Have also raised a PR in the browser repo
https://github.com/bitwarden/browser/pull/1290


kspearrin:
This seems browser extension specific, so it should just go there. No need for uriOptions to be shared through jslib.

mkanavakatini:
Cool, thanks for the direction @kspearrin, I'll close this now and make changes in the browser repo.

PR #114, opened 2020-06-12
Ensure viewPassword has a default value

Hinton:
Resolves bitwarden/mobile#973 for remaining client libs.

Tested on:

- [x] Web
- [x] Desktop
- [x] Browser

Had to use a slightly different approach since default values in TypeScript are not set when casting objects. Hence why it's instead checked if the value is null (undefined) in `Cipher` and not `CipherData`.


Hinton:
Seems to work fine in all environments. Except on Browser when I did an update on a running extension, however this should only happen for developers, since a regular extension upgrade would restart it.

PR #119, opened 2020-06-28
Desktop biometrics support

Hinton:
Initial draft with support for Windows Hello.

Implemented mostly the same as on mobile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,14 @@
 # Bitwarden JavaScript Library
 
 Common code referenced across Bitwarden JavaScript projects.
+
+## Requirements
+
+* Git
+* node-gyp
+
+### Windows
+
+* *Microsoft Build Tools 2015* in Visual Studio Installer
+* [Windows 10 SDK 17134](https://developer.microsoft.com/en-us/windows/downloads/sdk-archive/)
+either by downloading it seperately or through the Visual Studio Installer.
--- a/src/angular/components/lock.component.ts
+++ b/src/angular/components/lock.component.ts
@@ -25,14 +25,13 @@ export class LockComponent implements OnInit {
     email: string;
     pinLock: boolean = false;
     webVaultHostname: string = '';
-    supportsBiometric: boolean;
+    biometricLock: boolean;
 
     protected successRoute: string = 'vault';
     protected onSuccessfulSubmit: () => void;
 
     private invalidPinAttempts = 0;
     private pinSet: [boolean, boolean];
-    private biometricLock: boolean;
 
     constructor(protected router: Router, protected i18nService: I18nService,
         protected platformUtilsService: PlatformUtilsService, protected messagingService: MessagingService,
@@ -50,7 +49,6 @@ export class LockComponent implements OnInit {
             vaultUrl = 'https://bitwarden.com';
         }
         this.webVaultHostname = Utils.getHostname(vaultUrl);
-        this.supportsBiometric = await this.platformUtilsService.supportsBiometric();
     }
 
     async submit() {
--- a/src/angular/components/lock.component.ts
+++ b/src/angular/components/lock.component.ts
@@ -60,7 +60,7 @@ export class LockComponent implements OnInit {
         }
         this.webVaultHostname = Utils.getHostname(vaultUrl);
         this.router.routerState.root.queryParams.pipe(first()).subscribe(params => {
-            if (params['promptBiometric']) {
+            if (this.supportsBiometric && params['promptBiometric']) {
                 this.unlockBiometric();
             }
         });
--- a/src/angular/components/lock.component.ts
+++ b/src/angular/components/lock.component.ts
@@ -25,6 +25,7 @@ export class LockComponent implements OnInit {
     email: string;
     pinLock: boolean = false;
     webVaultHostname: string = '';
+    supportsBiometric: boolean;
     biometricLock: boolean;
     biometricText: string;
 
@@ -43,6 +44,7 @@ export class LockComponent implements OnInit {
     async ngOnInit() {
         this.pinSet = await this.vaultTimeoutService.isPinLockSet();
         this.pinLock = (this.pinSet[0] && this.vaultTimeoutService.pinProtectedKey != null) || this.pinSet[1];
+        this.supportsBiometric = await this.platformUtilsService.supportsBiometric();
         this.biometricLock = await this.vaultTimeoutService.isBiometricLockSet();
         this.biometricText = await this.storageService.get(ConstantsService.biometricText);
         this.email = await this.userService.getEmail();
--- a/src/angular/components/lock.component.ts
+++ b/src/angular/components/lock.component.ts
@@ -1,6 +1,7 @@
 import { OnInit } from '@angular/core';
 import { Router } from '@angular/router';
 
+import { ApiService } from '../../abstractions/api.service';
 import { CryptoService } from '../../abstractions/crypto.service';
 import { EnvironmentService } from '../../abstractions/environment.service';
 import { I18nService } from '../../abstractions/i18n.service';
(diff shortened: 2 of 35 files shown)


src/globals.d.ts
```diff
@@ -1,3 +1,192 @@
 declare function escape(s: string): string;
 declare function unescape(s: string): string;
 declare module 'duo_web_sdk';
+
+declare module "@nodert-win10-rs4/windows.security.credentials" {
```
kspearrin, line 5: Did you generate these typing somehow?

src/services/constants.service.ts
```diff
@@ -23,6 +23,7 @@ export class ConstantsService {
     static readonly protectedPin: string = 'protectedPin';
     static readonly clearClipboardKey: string = 'clearClipboardKey';
     static readonly eventCollectionKey: string = 'eventCollection';
+    static readonly biometricUnlockKey = 'fingerprintUnlock';
```
kspearrin, line 26: We can call the string biometric here. It was left as fingerprint in mobile for backwards compat

src/electron/biometric.main.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const requestCreateAsync = util.promisify(
+    KeyCredentialManager.requestCreateAsync,
+);
+
+export class BiometricMain {
+    constructor(private storageService: StorageService) {}
+
+    async init() {
+        this.storageService.save(ElectronConstants.enableBiometric, await this.supportsBiometric());
```
kspearrin, line 19: Why not just ipc like the request one? Not suggesting it, just wondering why?

src/globals.d.ts
Hinton: The typings within the package uses the name `windows.security.credentials` which were not picked up. I did some quick searching but seems like there is no way to create an "alias". Not sure why the package provides typing without the prefix.

src/globals.d.ts
```diff
@@ -1,3 +1,192 @@
...
+      retrievePublicKey(blobType: Number): Object;
+  
+    }
+  
+    export class KeyCredentialManager {
+      constructor();
+  
+      static isSupportedAsync(callback: (error: Error, result: Boolean) => void): void ;
+  
+
```
cscharf, line 81: will need just some minor cleanup of whitespace, etc.

src/electron/biometric.main.ts
```diff
@@ -1,16 +1,20 @@
...
+    UserConsentVerifier,
+    UserConsentVerifierAvailability,
+    UserConsentVerificationResult,
+} from '@nodert-win10-rs4/windows.security.credentials.ui';
 
-const requestCreateAsync = util.promisify(
-    KeyCredentialManager.requestCreateAsync,
-);
+const requestVerification: any = util.promisify(UserConsentVerifier.requestVerificationAsync);
+const checkAvailability: any= util.promisify(UserConsentVerifier.checkAvailabilityAsync);
```
cscharf, line 12: add space between `any=`

src/electron/biometric.main.ts
```diff
@@ -23,27 +27,15 @@ export class BiometricMain {
...
-                (error, resp) => {
-                    if (resp.status === KeyCredentialStatus.success) {
-                        resolve(true);
-                    } else {
-                        resolve(false);
-                    }
-                },
-            );
-        });
+        const verification = await requestVerification("Please authenticate yourself to unlock the Bitwarden Vault.");
```
cscharf, line 37: string needs to be localized?

src/globals.d.ts
```diff
@@ -1,192 +1,103 @@
...
-      success,
-      unknownError,
-      notFound,
-      userCanceled,
-      userPrefersPassword,
-      credentialAlreadyExists,
-      securityDeviceLocked,
+declare module "@nodert-win10-rs4/windows.security.credentials.ui" {
+    export enum AuthenticationProtocol {
+        basic,
```
cscharf, line 7: remove added whitespace

src/electron/biometric.main.ts
Hinton: It seemed like a better solution to store it rather than constantly calling the method.

README.md
```diff
@@ -3,3 +3,14 @@
...
 Common code referenced across Bitwarden JavaScript projects.
+
+## Requirements
+
+* Git
+* node-gyp
+
+### Windows
+
+* *Microsoft Build Tools 2015* in Visual Studio Installer
```
kspearrin, line 14: Do we know if 2017 build tools are also required?

src/angular/components/lock.component.ts
```diff
@@ -46,12 +50,20 @@ export class LockComponent implements OnInit {
...
         let vaultUrl = this.environmentService.getWebVaultUrl();
         if (vaultUrl == null) {
             vaultUrl = 'https://bitwarden.com';
         }
         this.webVaultHostname = Utils.getHostname(vaultUrl);
+        this.router.routerState.root.queryParams.pipe(first()).subscribe(params => {
+            if (this.supportsBiometric && params['promptBiometric']) {
+                this.unlockBiometric();
+            }
+        });
```
kspearrin, line 66: I don't think this is how we check query params in our other components. Search `this.route.queryParams.subscribe` in code. Not sure I follow why this scenario is different.

src/angular/components/lock.component.ts
```diff
@@ -46,12 +50,20 @@ export class LockComponent implements OnInit {
...
+        this.biometricLock = await this.vaultTimeoutService.isBiometricLockSet();
+        this.biometricText = await this.storageService.get(ConstantsService.biometricText);
         this.email = await this.userService.getEmail();
         let vaultUrl = this.environmentService.getWebVaultUrl();
         if (vaultUrl == null) {
             vaultUrl = 'https://bitwarden.com';
         }
         this.webVaultHostname = Utils.getHostname(vaultUrl);
+        this.router.routerState.root.queryParams.pipe(first()).subscribe(params => {
+            if (this.supportsBiometric && params['promptBiometric']) {
```
kspearrin, line 63: linter should complain that you can use `params.promptBiometric`

src/electron/services/electronPlatformUtils.service.ts
```diff
@@ -16,7 +17,9 @@ import { I18nService } from '../../abstractions/i18n.service';
 import { MessagingService } from '../../abstractions/messaging.service';
 import { PlatformUtilsService } from '../../abstractions/platformUtils.service';
 
+import { StorageService } from '../../abstractions';
```
kspearrin, line 20: be explicit with abstraction imports like the above

src/electron/biometric.windows.main.ts
```diff
@@ -0,0 +1,46 @@
...
+} from '@nodert-win10-rs4/windows.security.credentials.ui';
+import { I18nService, StorageService } from '../abstractions';
+
+import { ipcMain } from 'electron';
+import { BiometricMain } from '../abstractions/biometric.main';
+import { ConstantsService } from '../services';
+import { ElectronConstants } from './electronConstants';
+
+const requestVerification: any = util.promisify(UserConsentVerifier.requestVerificationAsync);
+const checkAvailability: any = util.promisify(UserConsentVerifier.checkAvailabilityAsync);
```
kspearrin, line 16: TIL

src/electron/biometric.windows.main.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const allowedAvailabilities = [
```
kspearrin, line 18: const for the class like this we name with PascalCase. `AllowedAvailabilities`

README.md
Hinton: I have VS 2019 installed in a VM with only 2015 and not 2017.

src/angular/components/lock.component.ts
Hinton: We use it in vault.component, however that scenario is a bit different since the route actually changes and interacts with the component. In this case I was only interested in reading the queryParms once to decide if it should prompt a biometric unlock hence the use if `first()` which automatically closes the subscription after the first message.

PR #124, opened 2020-07-10
Bug.568.organization search ignores custom fields

addisonbeck:
## Objective
Add custom fields to consideration when returning a filtered list of ciphers

## Code change
Added some custom fields fun to the basic search method

## Relevant Issues
Fixes bitwarden/web#568

### DIFF
--- a/src/services/search.service.ts
+++ b/src/services/search.service.ts
@@ -164,6 +164,20 @@ export class SearchService implements SearchServiceAbstraction {
             if (c.login && c.login.uri != null && c.login.uri.toLowerCase().indexOf(query) > -1) {
                 return true;
             }
+
+
+            if(c.hasFields){
+                for (let field of c.fields) {
+                    if (field.value.toLowerCase().indexOf(query) > -1) {
+                        return true;
+                    }
+
+                    if(field.name.toLowerCase().indexOf(query) > -1) {
+                        return true;
+                    }
+                }
+            }
+
             return false;
         });
     }
--- a/src/services/search.service.ts
+++ b/src/services/search.service.ts
@@ -166,7 +166,7 @@ export class SearchService implements SearchServiceAbstraction {
             }
 
             if (c.hasFields) {
-                for (let field of c.fields) {
+                for (const field of c.fields) {
                     if (field.value.toLowerCase().indexOf(query) > -1) {
                         return true;
                     }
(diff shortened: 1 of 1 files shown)


src/services/search.service.ts
```diff
@@ -164,6 +164,19 @@ export class SearchService implements SearchServiceAbstraction {
             if (c.login && c.login.uri != null && c.login.uri.toLowerCase().indexOf(query) > -1) {
                 return true;
             }
+
+            if(c.hasFields) {
```
cscharf, line 168: Please check your VS Code + VS macOS formatting settings to ensure for JS/TS there's a space between control keywords and the open parens on these blocks.

addisonbeck: Sorry about that. I'm a Vim guy, still working out some discrepancies between my formatting setup & the VS Code defaults. Should be resolved now.

kspearrin:
historically this was purposely not in there because i thought it might make this method of search even more inefficient than it already is. have we evaluated if there is any perceivable performance cost from doing this? also, we would need to update other codebases like CLI so that basic search is consistent.

addisonbeck:
@kspearrin do you think a better solution would be moving Organization search off `searchCiphersBasic()` and using `searchCiphers()` instead?

kspearrin:
Yes, but I was not able to come up with a good solution for that since the organization vault is not maintained in local storage. You'd have to maintain the search index in some other way.

addisonbeck:
Is this something that might be better left alone, in that case? The issue is marked as low priority, and I would imagine low priority issues aren't worth the risk involved in designing a new solution for vault storage. 

One possible solution that still exists is simply toggling custom field filtering on basic search behind a parameter, allowing us to use it where we want and not take the potential performance hit where the functionality isn't needed. As for the performance hit itself, it would of course scale with how many custom fields exist in a given vault, but I think it would take a significant amount before any slowness would be noticeable. If we decide to go this route I'll try and push it to see for sure.

Also, and this one is just more for curiosity, why are org vaults not stored in local storage? I'm assuming it has to do with sync concerns, but am interested to hear about it.

kspearrin:
I think just added custom fields will still leave more to be desired and is just kicking the can down the road. The real solution that users are looking for is full-text search like we have implemented the client vault.

@cscharf  Let's maybe schedule a call to discuss possible solutions.

PR #125, opened 2020-07-13
Modified response to include port if exists

MattTheDev:
## Objective

It was requested in https://github.com/bitwarden/desktop/issues/449 that we don't parse out port in the Website list. Small modification / request, added port if it exists to the host that returns from getHostname to support this request.

## Code Changes

**src/misc/utils.ts** Modified response - added if port isn't null, append to end of hostname.

### DIFF
--- a/src/misc/utils.ts
+++ b/src/misc/utils.ts
@@ -159,7 +159,7 @@ export class Utils {
         try {
             let hostname = url != null && url.hostname !== '' ? url.hostname : null;
 
-            if(hostname != null && url.port !== '') {
+            if (hostname != null && url.port !== '') {
                 hostname += ":" + url.port;
             }
 
--- a/src/misc/utils.ts
+++ b/src/misc/utils.ts
@@ -157,7 +157,13 @@ export class Utils {
     static getHostname(uriString: string): string {
         const url = Utils.getUrl(uriString);
         try {
-            return url != null && url.hostname !== '' ? url.hostname : null;
+            let hostname = url != null && url.hostname !== '' ? url.hostname : null;
+
+            if(hostname != null && url.port !== '') {
+                hostname += ":" + url.port;
+            }
+
+            return hostname;
         } catch {
             return null;
         }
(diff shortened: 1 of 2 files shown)


src/misc/utils.ts
```diff
@@ -157,7 +157,13 @@ export class Utils {
...
         const url = Utils.getUrl(uriString);
         try {
-            return url != null && url.hostname !== '' ? url.hostname : null;
+            let hostname = url != null && url.hostname !== '' ? url.hostname : null;
+
+            if (hostname != null && url.port !== '') {
+                hostname += ":" + url.port;
+            }
+
+            return hostname;
```
kspearrin, line 166: This is what `getHost()` below is for. It should include the port already.

cscharf: @kspearrin , any objection changing the various clients' `view` UI to use `getHost()` instead of `getHostname()` for display purposes?

kspearrin: Yes, that would be the proper fix for the above-linked issue.

MattTheDev: Reverted change. Made call to getHost.

src/models/view/loginUriView.ts
```diff
@@ -62,7 +62,7 @@ export class LoginUriView implements View {
             return null;
         }
         if (this._hostname == null && this.uri != null) {
-            this._hostname = Utils.getHostname(this.uri);
```
kspearrin, line 65: Can you please verify all locations where this `get hostname()` property is used? The property is explicitly called `hostname`, which is different than a `host`, which we have now changed it to. I think a more proper fix here would be to introduce a new `host` property and then change the clients to reference it on the "view" pages, rather than hostname.

MattTheDev: That's the route I was going to take - was afraid that I was going to touch something and inadvertently break something somewhere else. 

Will go this route.

MattTheDev: What I did -

1. Added _host, get host, and get hostOrUri .. mirroring the hostname functionality.
2. When this gets approve - I will modify the following projects:

image

that reference hostnameOrUri on the UI to hostOrUri .. this will make it consistent across the board. 

3. Lastly, will modify mobile project to do the same .. it has a similar function, though not shared in jslib.

PR #127, opened 2020-07-15
sso support

kspearrin:
support for sso implementation on server and web

- update auth service to support sso flow with `logInSso`
- lock component to support sso cases in which there is no existing master password hash. call to api for verification
- pull kdf info and new reset master password properties from token response
- add sso state and code verified keys to constants

### DIFF
--- a/src/models/request/tokenRequest.ts
+++ b/src/models/request/tokenRequest.ts
@@ -44,6 +44,8 @@ export class TokenRequest {
             obj.code = this.code;
             obj.code_verifier = this.codeVerifier;
             obj.redirect_uri = this.redirectUri;
+        } else {
+            throw new Error('must provide credentials or codes');
         }
 
         if (this.device) {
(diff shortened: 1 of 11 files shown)


src/models/request/tokenRequest.ts
```diff
@@ -22,13 +31,21 @@ export class TokenRequest {
...
+        if (this.masterPasswordHash != null && this.email != null) {
+            obj.grant_type = 'password';
+            obj.username = this.email;
+            obj.password = this.masterPasswordHash;
+        } else if (this.code != null && this.codeVerifier != null && this.redirectUri != null) {
+            obj.grant_type = 'authorization_code';
+            obj.code = this.code;
+            obj.code_verifier = this.codeVerifier;
+            obj.redirect_uri = this.redirectUri;
+        }
```
cscharf, line 47: Is there a handler or possibility that this could ever have one final fall-out where `grant_type` is not set or validation around that? Say if `username` or `masterPasswordHash` are `null` and/or one of the `sso` parameters?

PR #130, opened 2020-07-16
Set up API calls for bulk delete from within an org vault

addisonbeck:
## Objective
Create necessary API calls for bulk deleting ciphers from an org vault

## Code Changes
Created relevant requests

## Related Issues
Relates to [bitwarden/web #534](https://github.com/bitwarden/web/issues/534)
[bitwarden/web PR #577](https://github.com/bitwarden/web/pull/577) depends on this change


src/angular/components/ciphers.component.ts
```diff
@@ -126,4 +128,31 @@ export class CiphersComponent {
...
+
+    getSelected(): CipherView[] {
+        if (this.ciphers == null) {
+            return [];
+        }
+        return this.ciphers.filter((c) => !!(c as any).checked);
+    }
+
+    getSelectedIds(): string[] {
+        return this.getSelected().map((c) => c.id);
```
kspearrin, line 156: I dont think this needs to be in jslib since it still only applies to the web vault. the org ciphers componenet already inherits from the regular ciphers component where this lives in the web repo so it should be reusable already.

addisonbeck: Done

PR #131, opened 2020-07-17
Reference id to data conversion

cscharf:
## Objective
Companion to the server change for reference id to be a data field (json blob) with a structured model for API submission through jslib. Need to capture 3 data points to send back through for flow and layout in addition to id.

### DIFF
--- a/src/angular/components/register.component.ts
+++ b/src/angular/components/register.component.ts
@@ -12,7 +12,7 @@ import { PlatformUtilsService } from '../../abstractions/platformUtils.service';
 import { StateService } from '../../abstractions/state.service';
 
 import { KdfType } from '../../enums/kdfType';
-import { ReferenceEventData } from '../../models/domain/referenceEventData';
+import { ReferenceEventRequest } from '../../models/request/referenceEventRequest';
 
 export class RegisterComponent {
     name: string = '';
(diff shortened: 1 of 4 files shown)


src/models/domain/referenceEventData.ts
```diff
@@ -0,0 +1,5 @@
+export class ReferenceEventData {
```
kspearrin, line 1: Per server PR comments, this should probably be moved and renamed as a API request model.

src/models/request/registerRequest.ts
```diff
@@ -1,6 +1,7 @@
 import { KeysRequest } from './keysRequest';
 
 import { KdfType } from '../../enums/kdfType';
+import { ReferenceEventData } from '../domain/referenceEventData';
```
kspearrin, line 4: can move this up with the other request model

cscharf: Done

src/models/domain/referenceEventData.ts
cscharf: Done, hopefully 🤞

src/angular/components/register.component.ts
```diff
@@ -12,6 +12,7 @@ import { PlatformUtilsService } from '../../abstractions/platformUtils.service';
 import { StateService } from '../../abstractions/state.service';
 
 import { KdfType } from '../../enums/kdfType';
+import { ReferenceEventRequest } from '../../models/request/referenceEventRequest';
```
kspearrin, line 15: you'll notice i try to keep all of these imports in a logical grouping. all request models are together, all services together, enums together, etc. move this up with the other request models.

cscharf: Done, sorry, I missed this one.

PR #138, opened 2020-07-31
try catch nodeRT module via require

cscharf:
This is an attempt to fix https://github.com/bitwarden/desktop/issues/495 by lazy-loading the offending module inside a `try {} catch {}` instead of using `import`. This should prevent an app crash as well as mark an `isError` flag on the biometric module so the UI can display a message as to why it's not available (i.e. not supported on machine or need dependency, etc.)

@kspearrin , I have been unable to test this, I can't get it build in my Windows VM and since I'm using macOS natively my build can't compile native Windows applications so I'm not able to test.

Can you please pull this draft PR and attempt a build/check. If it actually loads and Windows Hello still works, can you also please post it to 495 linked above for the user-community to test with as well (I also believe Luc can reproduce the issue so he'd also be able to test).


cscharf:
Fixes https://github.com/bitwarden/desktop/issues/495 (confirmed)

PR #140, opened 2020-08-03
SSO login for generic clients and CLI

kspearrin:
- Support SSO option when invoking the login command
- Support for other clients to call into SSO component ans pass parameters

### DIFF
--- a/src/cli/commands/login.command.ts
+++ b/src/cli/commands/login.command.ts
@@ -243,8 +243,8 @@ export class LoginCommand {
                             '&redirectUri=' + encodeURIComponent(this.ssoRedirectUri) +
                             '&state=' + state + '&codeChallenge=' + codeChallenge);
                     });
+                    break;
                 } catch { }
-                break;
             }
         });
     }
--- a/src/cli/commands/login.command.ts
+++ b/src/cli/commands/login.command.ts
@@ -234,6 +234,7 @@ export class LoginCommand {
                     callbackServer.close(() => reject());
                 }
             });
+            let foundPort = false;;
             const webUrl = this.environmentService.webVaultUrl == null ? 'https://vault.bitwarden.com' :
                 this.environmentService.webVaultUrl;
             for (let port = 8065; port <= 8070; port++) {
@@ -244,8 +245,12 @@ export class LoginCommand {
                             '&redirectUri=' + encodeURIComponent(this.ssoRedirectUri) +
                             '&state=' + state + '&codeChallenge=' + codeChallenge);
                     });
+                    foundPort = true;
                     break;
                 } catch { }
+            }
+            if (!foundPort) {
+                reject();
             }
         });
     }
--- a/src/cli/commands/login.command.ts
+++ b/src/cli/commands/login.command.ts
@@ -1,4 +1,5 @@
 import * as program from 'commander';
+import * as http from 'http';
 import * as inquirer from 'inquirer';
 
 import { TwoFactorProviderType } from '../../enums/twoFactorProviderType';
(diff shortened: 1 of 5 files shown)


src/cli/commands/login.command.ts
```diff
@@ -162,4 +208,44 @@ export class LoginCommand {
...
+        return new Promise((resolve, reject) => {
+            const callbackServer = http.createServer((req, res) => {
+                const urlString = 'http://localhost' + req.url;
+                const url = new URL(urlString);
+                const code = url.searchParams.get('code');
+                const receivedState = url.searchParams.get('state');
+                res.setHeader('Content-Type', 'text/html');
+                if (code != null && receivedState != null && receivedState === state) {
+                    res.writeHead(200);
+                    res.end('<html><head><title>Success | Bitwarden CLI</title></head><body>' +
```
cscharf, line 222: Will want to consider localization from language resources or a localized html template/file for this.

src/cli/commands/login.command.ts
```diff
@@ -162,4 +208,44 @@ export class LoginCommand {
...
+                        '<h1>Successfully authenticated with the Bitwarden CLI</h1>' +
+                        '<p>You may now close this tab and return to the terminal.</p>' +
+                        '</body></html>');
+                    callbackServer.close(() => resolve(code));
+                } else {
+                    res.writeHead(400);
+                    res.end('<html><head><title>Failed | Bitwarden CLI</title></head><body>' +
```
cscharf, line 229: same

src/cli/commands/login.command.ts
```diff
@@ -162,4 +208,44 @@ export class LoginCommand {
...
+            for (let port = 8065; port <= 8070; port++) {
+                try {
+                    this.ssoRedirectUri = 'http://localhost:' + port;
+                    callbackServer.listen(port, async () => {
+                        await open(webUrl + '/#/sso?clientId=cli' +
+                            '&redirectUri=' + encodeURIComponent(this.ssoRedirectUri) +
+                            '&state=' + state + '&codeChallenge=' + codeChallenge);
+                    });
+                } catch { }
+                break;
```
cscharf, line 247: Shouldn't `break;` of the loop be inside of the `try { }` portion of this, otherwise it'll try exactly once (assuming the idea here is to try against several ports until one successfully is open/available?)

src/cli/commands/login.command.ts
kspearrin: Probably, however, the CLI does not localize anything today.

src/cli/commands/login.command.ts
kspearrin: yea, this looks wrong. will fix.

kspearrin: fixed

cscharf: Also, wouldn't you also need to call the `reject` if an open port could not be established or code execution reaches the end of this method? Perhaps `return;` would appropriate instead of `break;` where with outside of the loop finally rejecting the promise to resolve/return control to the main flow.

src/services/passwordGeneration.service.ts
```diff
@@ -261,7 +261,8 @@ export class PasswordGenerationService implements PasswordGenerationServiceAbstr
     }
 
     async getPasswordGeneratorPolicyOptions(): Promise<PasswordGeneratorPolicyOptions> {
-        const policies: Policy[] = await this.policyService.getAll(PolicyType.PasswordGenerator);
+        const policies: Policy[] = this.policyService == null ? null :
+            await this.policyService.getAll(PolicyType.PasswordGenerator);
```
kspearrin, line 265: whoops. this is for my next PR for bwdc, but let's just keep it here.

src/cli/commands/login.command.ts
```diff
@@ -162,4 +209,49 @@ export class LoginCommand {
...
+                } else {
+                    res.writeHead(400);
+                    res.end('<html><head><title>Failed | Bitwarden CLI</title></head><body>' +
+                        '<h1>Something went wrong logging into the Bitwarden CLI</h1>' +
+                        '<p>You may now close this tab and return to the terminal.</p>' +
+                        '</body></html>');
+                    callbackServer.close(() => reject());
+                }
+            });
+            let foundPort = false;;
```
cscharf, line 237: ERROR: src/cli/commands/login.command.ts:237:35 - Unnecessary semicolon
ERROR: src/cli/commands/login.command.ts:237:35 - missing whitespace

PR #144, opened 2020-08-07
[SSO] Login - added launchSsoBrowser method

vincentsalucci:
## Code Changes
- **login.component.ts**: Added `launchSsoBrowser` method. Cleaned up imports and adjusted constructor.

### DIFF
--- a/src/angular/components/login.component.ts
+++ b/src/angular/components/login.component.ts
@@ -130,6 +130,10 @@ export class LoginComponent implements OnInit {
         const codeVerifierHash = await this.cryptoFunctionService.hash(ssoCodeVerifier, 'sha256');
         const codeChallenge = Utils.fromBufferToUrlB64(codeVerifierHash);
 
+        // Save sso params
+        await this.storageService.save(ConstantsService.ssoStateKey, state);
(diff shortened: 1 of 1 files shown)


src/angular/components/login.component.ts
```diff
@@ -37,7 +40,9 @@ export class LoginComponent implements OnInit {
 
     constructor(protected authService: AuthService, protected router: Router,
         protected platformUtilsService: PlatformUtilsService, protected i18nService: I18nService,
-        private storageService: StorageService, protected stateService: StorageService) { }
+        protected stateService: StorageService, protected environmentService: EnvironmentService,
```
addisonbeck, line 43: @vincentsalucci is this intended to be have two instances of a StorageService? Seems like it should be `stateService: StateService` ?

addisonbeck: I'm also wondering if the CryptoFunctionService needs to get added to the index.ts in the abstractions folder?

addisonbeck: ( went ahead and threw together a PR for this stuff in case it is needed )

PR #146, opened 2020-08-11
Misc jslib cleanups

addisonbeck:
### Code Changes

`login.component`
changed variable `stateService` to be of type `StateService()`
combined the abstractions imports into one group

`abstractions/index.ts`
added an export for `CryptoFunctionService`

### DIFF
--- a/src/abstractions/index.ts
+++ b/src/abstractions/index.ts
@@ -4,6 +4,7 @@ export { AuditService } from './audit.service';
 export { AuthService } from './auth.service';
 export { CipherService } from './cipher.service';
 export { CollectionService } from './collection.service';
+export { CryptoFunctionService } from './cryptoFunction.service';
 export { CryptoService } from './crypto.service';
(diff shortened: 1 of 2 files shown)


src/abstractions/index.ts
```diff
@@ -4,6 +4,7 @@ export { AuditService } from './audit.service';
 export { AuthService } from './auth.service';
 export { CipherService } from './cipher.service';
 export { CollectionService } from './collection.service';
+export { CryptoFunctionService } from './cryptoFunction.service';
 export { CryptoService } from './crypto.service';
 export { EnvironmentService } from './environment.service';
 export { FolderService } from './folder.service';
```
kspearrin, line 10: I have not been using these index barrels for a while now and would actually like to do away with them completely. They create issues with importing dependencies on certain platforms.

addisonbeck: Nice, I think that's for the best. I added a deprecation warning to index.ts instead.

PR #147, opened 2020-08-12
Modifications made to support Browser based SSO.

MattTheDev:
## Objective

Add the ability to SSO via Browser Extension.

## Code Changes

sso.component - 

The path for browser SSO is slightly different.. and we're routing the data back to the extension to complete the loginSso portion. I wrote this in a way that works .. but being in jslib, maybe it can be written better? Very open to feedback here.

Lines 64 - 72:

This data is required to track back in the message to the extension.

Lines 116 - 119:

This flow is required to post the message, so the extension can complete its part of the authentication.

constants.service - 

added 3 new constants for easy use / removal of random strings in code that may be reused elsewhere.

### DIFF
--- a/src/abstractions/platformUtils.service.ts
+++ b/src/abstractions/platformUtils.service.ts
@@ -19,6 +19,7 @@ export abstract class PlatformUtilsService {
      */
     lockTimeout: () => number;
     launchUri: (uri: string, options?: any) => void;
+    closeCurrentTab: () => void;
     saveFile: (win: Window, blobData: any, blobOptions: any, fileName: string) => void;
     getApplicationVersion: () => string;
     supportsU2f: (win: Window) => boolean;
--- a/src/angular/components/sso.component.ts
+++ b/src/angular/components/sso.component.ts
@@ -51,7 +51,7 @@ export class SsoComponent {
                 await this.storageService.remove(ConstantsService.ssoCodeVerifierKey);
                 await this.storageService.remove(ConstantsService.ssoStateKey);
                 if (qParams.code != null && codeVerifier != null && state != null && state === qParams.state) {
-                    await this.logIn(qParams.code, codeVerifier);
+                    await this.logIn(qParams.code, codeVerifier, state);
                 }
             } else if (qParams.clientId != null && qParams.redirectUri != null && qParams.state != null &&
                 qParams.codeChallenge != null) {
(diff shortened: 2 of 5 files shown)


src/angular/components/sso.component.ts
```diff
@@ -61,7 +69,7 @@ export class SsoComponent {
                 this.clientId = qParams.clientId;
             }
 
-            if (qParams.clientId === ConstantsService.browserClientId) {
+            if (qParams.clientId == ConstantsService.browserClientId) {
```
cscharf, line 72: Why not use the explicit equality test? Was there a reason perhaps `===` was failing the test while the equalish test was passing?

src/abstractions/platformUtils.service.ts
```diff
@@ -19,6 +19,7 @@ export abstract class PlatformUtilsService {
      */
     lockTimeout: () => number;
     launchUri: (uri: string, options?: any) => void;
+    closeCurrentTab: () => void;
```
kspearrin, line 22: Does this need to be a global util? Seems like you will just have a bunch of no-ops. Could you just add this to the browser extension's popup-utils?

src/angular/components/sso.component.ts
```diff
@@ -53,13 +53,32 @@ export class SsoComponent {
...
                 this.clientId = qParams.clientId;
             }
+
+            if (qParams.clientId == ConstantsService.browserClientId) {
+                this.redirectUri = qParams.redirectUri;
+                this.state = qParams.state;
+                this.codeChallenge = qParams.codeChallenge;
+                this.clientId = qParams.clientId;
+                await this.storageService.save(ConstantsService.ssoCodeVerifierKey, qParams.codeVerifier);
+                await this.storageService.save(ConstantsService.ssoStateKey, qParams.state);
```
kspearrin, line 78: I cannot think of a use case where these need to be persisted in the sso component like this. This should be done on the client itself, the browser extension whenever they click the button on the login/home page.

src/angular/components/sso.component.ts
```diff
@@ -53,13 +53,32 @@ export class SsoComponent {
...
+                }
             } else if (qParams.clientId != null && qParams.redirectUri != null && qParams.state != null &&
                 qParams.codeChallenge != null) {
                 this.redirectUri = qParams.redirectUri;
                 this.state = qParams.state;
                 this.codeChallenge = qParams.codeChallenge;
                 this.clientId = qParams.clientId;
             }
+
+            if (qParams.clientId == ConstantsService.browserClientId) {
```
kspearrin, line 72: I think we could just piggyback on the qParams.redirectUri to get the desired functionality without having to introduce a new param. Let's have a call.

MattTheDev:
Per convo with Kyle - these changes can be put into the Web/browser projects to prevent impact to jslib .. which is always preferred :)

PR #148, opened 2020-08-12
Upgrade TypeScript

Hinton:
TypeScript 3.6+ is required to access newer versions of the navigator. apis.

- TypeScript updated to 3.6.5 which is the minimum minor version required.
- TSLint updated to latest, however TSlint is now deprecated so we probably want to switch to ESlint and Prettier sometime in the future.
- @types/node updated to 10.17.28

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "rimraf": "^2.6.2",
     "tslint": "^5.12.1",
     "typemoq": "^2.1.0",
-    "typescript": "3.2.4"
+    "typescript": "3.6.5"
   },
   "dependencies": {
     "@angular/animations": "7.2.1",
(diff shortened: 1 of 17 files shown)


kspearrin:
We'll need to have a follow up PR in all repos that match this upgrade.

package.json
```diff
@@ -30,7 +30,7 @@
     "@types/jasmine": "^3.3.12",
     "@types/lowdb": "^1.0.5",
     "@types/lunr": "^2.3.3",
-    "@types/node": "^10.9.4",
+    "@types/node": "^12.12.54",
```
kspearrin, line 33: We still use node 10 for building our apps. Is this necessary?

Hinton: We get compile errors otherwise, I can try using a newer version of 10.

Hinton: 10.17.28 seems to work

PR #150, opened 2020-08-14
extend functionality for set password flow

kspearrin:
Extend set password models to accept new api properties

update change password base component to work for setting password and other crypto keys for new users via sso


kspearrin:
@vincentsalucci yes, in normal authed states, you can fetch the information from userService. SSO auth doesn't return that when going through the new user flow (since you had no password set yet).

PR #151, opened 2020-08-14
Upgrade Angular to 9

Hinton:
Updated using `ng update`.


src/angular/components/view.component.ts
```diff
@@ -6,6 +6,7 @@ import {
     OnDestroy,
     OnInit,
     Output,
+    Directive,
```
kspearrin, line 9: linter should complain not alphabetial

Hinton: We seem to be missing the `ordered-imports` rule in tslint configs. I added the rule and ran `lint:fix`.

Hinton: After digging a bit, it seems that tslint:recommended has changed a bit between versions. Since it was updated during the `TypeScript` upgrade we got the new recommended rules.

Most if not all of the code formatting rules have been removed, https://github.com/palantir/tslint/pull/4871. TSlint is now also deprecated with the recommendation to migrate over to ESlint and use Prettier for code formatting rules.

We can either re-add the removed rules or migrate to ESlint and Prettier.

package.json
```diff
@@ -59,19 +59,18 @@
     "rimraf": "^2.6.2",
     "tslint": "^6.1.3",
     "typemoq": "^2.1.0",
-    "typescript": "3.6.5"
+    "typescript": "3.8.3"
   },
   "dependencies": {
-    "@angular/animations": "7.2.1",
+    "@angular/animations": "9.1.12",
     "@angular/cdk": "7.2.1",
```
kspearrin, line 66: Does this CDK need updating too?

PR #156, opened 2020-08-22
[SSO] Bug - Fixed set password route

vincentsalucci:
## Code Changes
- **two-factor.component.ts**: Fixed route to the `set-password` component (across all platforms), by checking the returned `AuthResult` and adjusting the success route (if necessary)

### DIFF
--- a/src/angular/components/two-factor.component.ts
+++ b/src/angular/components/two-factor.component.ts
@@ -2,7 +2,11 @@ import {
     OnDestroy,
     OnInit,
 } from '@angular/core';
-import { Router } from '@angular/router';
+
+import {
+    ActivatedRoute,
+    Router,
+} from '@angular/router';
 
 import { DeviceType } from '../../enums/deviceType';
 import { TwoFactorProviderType } from '../../enums/twoFactorProviderType';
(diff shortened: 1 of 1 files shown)


src/angular/components/two-factor.component.ts
```diff
@@ -59,8 +63,18 @@ export class TwoFactorComponent implements OnInit, OnDestroy {
...
         }
 
+        const queryParamsSub = this.route.queryParams.subscribe((qParams) => {
+            if (qParams.resetMasterPassword != null) {
+                this.resetMasterPassword = qParams.resetMasterPassword;
+            }
+
+            if (queryParamsSub != null) {
+                queryParamsSub.unsubscribe();
+            }
```
kspearrin, line 73: I am confused as to what this is used for. Why are we controlling this via query string?

vincentsalucci: Takes the `resetMasterPassword` information from the original response in the `sso.component` and passes it along to the `two-factor.component` so the route can be altered on a successful 2fa in the event of 2fa + reset master password flow.

Is there a better way to do this?

kspearrin: The 2fa token response should have a property that says you need to resetPassword.

vincentsalucci: Understood - reverted and adjusted conditional.

PR #158, opened 2020-08-25
Link existing user to sso

addisonbeck:
Objective

Create a jslib environment that will allow for linking/unlinking existing users to organization SSO

Code changes

Created API passthroughs for the nearly created DeleteSsoUser API method
Added Identifier, UseSso, and SsoBound properties to some response models for use in bitwarden/web
Added a returnUri property to the sso component for better control over where we put users post SSO authentication
Alphabetized the 'S' section of the Api methods in api.service

**NOTE** this PR was resubmitted because the old one was showing unchanged files as changed, not sure why, but a fresh PR removed the bad diffs.

### DIFF
--- a/src/angular/components/sso.component.ts
+++ b/src/angular/components/sso.component.ts
@@ -105,7 +105,7 @@ export class SsoComponent {
 
         if (this.linkLoggedInUser) {
             const userIdentifier = await this.apiService.getSsoUserIdentifier();
-            authorizeUrl += `&user_identifier=${userIdentifier}`;
+            authorizeUrl += `&user_identifier=${encodeURIComponent(userIdentifier)}`;
         }
 
         return authorizeUrl;
--- a/src/angular/components/sso.component.ts
+++ b/src/angular/components/sso.component.ts
@@ -36,6 +36,7 @@ export class SsoComponent {
     protected state: string;
     protected codeChallenge: string;
     protected returnUri: string;
+    protected userIdentifier: string;
 
     constructor(protected authService: AuthService, protected router: Router,
         protected i18nService: I18nService, protected route: ActivatedRoute,
@@ -93,12 +94,15 @@ export class SsoComponent {
             await this.storageService.save(ConstantsService.ssoStateKey, state);
         }
 
-        const authorizeUrl = this.apiService.identityBaseUrl + '/connect/authorize?' +
+        let authorizeUrl = this.apiService.identityBaseUrl + '/connect/authorize?' +
             'client_id=' + this.clientId + '&redirect_uri=' + encodeURIComponent(this.redirectUri) + '&' +
             'response_type=code&scope=api offline_access&' +
             'state=' + state + '&code_challenge=' + codeChallenge + '&' +
             'code_challenge_method=S256&response_mode=query&' +
             'domain_hint=' + encodeURIComponent(this.identifier);
+        if (this.userIdentifier) {
+            authorizeUrl += '&user_identifier=' + this.userIdentifier;
+        }
         this.platformUtilsService.launchUri(authorizeUrl, { sameWindow: true });
     }
 
--- a/src/angular/components/sso.component.ts
+++ b/src/angular/components/sso.component.ts
@@ -36,7 +36,16 @@ export class SsoComponent {
     protected state: string;
     protected codeChallenge: string;
     protected returnUri: string;
-    protected userIdentifier: string;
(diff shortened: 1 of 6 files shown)


src/models/data/organizationData.ts
```diff
@@ -17,11 +17,14 @@ export class OrganizationData {
     use2fa: boolean;
     useApi: boolean;
     useBusinessPortal: boolean;
+    useSso: boolean;
     selfHost: boolean;
     usersGetPremium: boolean;
     seats: number;
     maxCollections: number;
     maxStorageGb?: number;
+    ssoBound: boolean;
```
addisonbeck, line 26: @cscharf addressing your concerns on the old PR about ssoBound:

I'd agree that it may not be appropriate to have ssoBound on this model as it is currently named, but the inheritance hierarchy implied by these constructors starts with a parent model: ProfileOrganizationResponse. The version of this model on the server already had ssoBound, and it pulls from a table that includes org and orgUser information. The organizationData model is also what is put in storage, which is what is referenced on the organizations page that is relevant to ssoBound. 

I'd argue that instead of removing the property from this model, maybe the model should be renamed to more appropriately contextualize the data it provides. I imagine most of the time an Organization appears in a UI it is going to be more relevant to an OrgUser link than to just an Organization.

Let me know your thoughts on this.

cscharf: Ah, makes sense, thanks for the clarification @addisonbeck , this should be good as-is for now.

src/angular/components/sso.component.ts
```diff
@@ -35,6 +35,17 @@ export class SsoComponent {
     protected redirectUri: string;
     protected state: string;
     protected codeChallenge: string;
+    protected returnUri: string;
+    protected linkLoggedInUser: boolean;
```
kspearrin, line 39: Can this just be made a parameter of `buildAuthorizeUrl` instead? I am also thinking that `buildAuthorizeUrl` should probably be protected rather than public since it doesnt need to be called from the UI.

src/angular/components/sso.component.ts
```diff
@@ -66,34 +77,42 @@ export class SsoComponent {
...
 
-        const authorizeUrl = this.apiService.identityBaseUrl + '/connect/authorize?' +
+        let authorizeUrl = this.apiService.identityBaseUrl + '/connect/authorize?' +
             'client_id=' + this.clientId + '&redirect_uri=' + encodeURIComponent(this.redirectUri) + '&' +
             'response_type=code&scope=api offline_access&' +
             'state=' + state + '&code_challenge=' + codeChallenge + '&' +
             'code_challenge_method=S256&response_mode=query&' +
             'domain_hint=' + encodeURIComponent(this.identifier);
+
+        if (this.linkLoggedInUser) {
```
kspearrin, line 106: Maybe change this to a function param of `includeUserIdentifier` or something. Default it to false.

src/angular/components/sso.component.ts
```diff
@@ -66,34 +77,42 @@ export class SsoComponent {
...
+            const codeVerifier = await this.passwordGenerationService.generatePassword(this.passwordOptions);
             const codeVerifierHash = await this.cryptoFunctionService.hash(codeVerifier, 'sha256');
             codeChallenge = Utils.fromBufferToUrlB64(codeVerifierHash);
             await this.storageService.save(ConstantsService.ssoCodeVerifierKey, codeVerifier);
         }
+
         if (state == null) {
-            state = await this.passwordGenerationService.generatePassword(passwordOptions);
+            state = await this.passwordGenerationService.generatePassword(this.passwordOptions);
+            if (this.returnUri) {
```
kspearrin, line 92: Also thinking that this should just be a function parameter too.

src/angular/components/sso.component.ts
addisonbeck: Done, but for context this also involved adding some optional params to submit()

src/angular/components/sso.component.ts
addisonbeck: Done. Instead of defaulting to false I just made the parameter optional, but the effect is the same.

src/angular/components/sso.component.ts
addisonbeck: This one didn't have to even be a function parameter. It's constant and scoped just to builAuthorizeUri, so I just made it a variable there.

kspearrin: I was referring to redirectUri

PR #159, opened 2020-08-26
[SSO] Set Password flow sync

vincentsalucci:
## Code Changes
- **set-password.component.ts**: Added private `syncService`. Added tracking boolean for UI management.
- **login.command.ts**: Added error response for `resetMasterPassword` flow.

### DIFF
--- a/src/angular/components/set-password.component.ts
+++ b/src/angular/components/set-password.component.ts
@@ -7,6 +7,7 @@ import { MessagingService } from '../../abstractions/messaging.service';
 import { PasswordGenerationService } from '../../abstractions/passwordGeneration.service';
 import { PlatformUtilsService } from '../../abstractions/platformUtils.service';
 import { PolicyService } from '../../abstractions/policy.service';
+import { SyncService } from '../../abstractions/sync.service';
 import { UserService } from '../../abstractions/user.service';
 
 import { CipherString } from '../../models/domain/cipherString';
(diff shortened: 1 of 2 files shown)


src/angular/components/set-password.component.ts
```diff
@@ -29,11 +30,17 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
...
         platformUtilsService: PlatformUtilsService, policyService: PolicyService, private router: Router,
-        private apiService: ApiService) {
+        private apiService: ApiService, private syncService: SyncService) {
         super(i18nService, cryptoService, messagingService, userService, passwordGenerationService,
             platformUtilsService, policyService);
     }
 
+    async ngOnInit() {
+        this.formPromise = this.syncService.fullSync(true);
+        await this.formPromise;
```
kspearrin, line 40: This isn't related to the UI form so I would not be assigning it to the form promise. We might need a `loading` state with a spinner for this now, though, since the UI will not be in a valid state until the sync completes.

src/cli/commands/login.command.ts
```diff
@@ -200,6 +200,11 @@ export class LoginCommand {
                 return Response.error('Login failed.');
             }
 
+            if (response.resetMasterPassword) {
+                return Response.error('In order to login with SSO from the CLI, you must first initiate the same' +
+                    ' process through the web vault to set a master password.');
```
kspearrin, line 205:
```suggestion
                return Response.error('In order to log in with SSO from the CLI, you must first log in' +
                    ' through the web vault to set your master password.');
```

src/angular/components/set-password.component.ts
vincentsalucci: Updated with tracking boolean.

src/cli/commands/login.command.ts
vincentsalucci: Updated to suggestion.

src/angular/components/set-password.component.ts
```diff
@@ -20,6 +21,7 @@ import { ChangePasswordComponent as BaseChangePasswordComponent } from './change
 import { KdfType } from '../../enums/kdfType';
 
 export class SetPasswordComponent extends BaseChangePasswordComponent {
+    loadingSync: boolean = true;
```
kspearrin, line 24: Shouldnt you be setting this to false after sync?

vincentsalucci: Yep - it didn't make it into the commit for some reason. Updated.

PR #160, opened 2020-08-28
Wrap sso login with pre-validation check

cscharf:
## Overview
Added a pre-validation check for SSO calls to pass the org identifier (domain_hint) through the Identity PreValidate endpoint to ensure the SSO pipeline _should_ work end-to-end.

## Changes
* **api.service** - function calls the Identity /Account/PreValidate pre-validate endpoint which runs a full-spectrum deep validation of SSO setup before actually redirecting the user to attempt the SSO flow itself. This will show any error messages to the user in a friendly manner, and otherwise kick off the flow as usual/expected.
* **sso.component** - Added `preValidate()` method to wrap and handle the call to validate the domain_hint and pipe the error to the user if necessary.

## Screen Shot
image

### DIFF
--- a/src/angular/components/sso.component.ts
+++ b/src/angular/components/sso.component.ts
@@ -21,8 +21,10 @@ import { AuthResult } from '../../models/domain/authResult';
 export class SsoComponent {
     identifier: string;
     loggingIn = false;
+    initiatingSso = false;
 
     formPromise: Promise<AuthResult>;
+    initiateSsoFormPromise: Promise<boolean>;
     onSuccessfulLogin: () => Promise<any>;
     onSuccessfulLoginNavigate: () => Promise<any>;
     onSuccessfulLoginTwoFactorNavigate: () => Promise<any>;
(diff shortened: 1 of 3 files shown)


src/angular/components/sso.component.ts
```diff
@@ -67,8 +67,23 @@ export class SsoComponent {
     }
 
     async submit(returnUri?: string, includeUserIdentifier?: boolean) {
-        const authorizeUrl = await this.buildAuthorizeUrl(returnUri, includeUserIdentifier);
-        this.platformUtilsService.launchUri(authorizeUrl, { sameWindow: true });
+        if (await this.preValidate()) {
```
kspearrin, line 70: Generally, with anything network related like this API call, we associate this form with a promise and use `[appApiAction]="formPromise"` on the form itself so we can show a spinner on submission. You should be able to find multiple examples.

cscharf: I've updated the pattern

kspearrin (approved):

PR #164, opened 2020-09-04
updated sso password string to be 32 characters

addisonbeck:
Split the length of the sso.component generated password in an attempt to bring URL size down

### DIFF
--- a/src/angular/components/sso.component.ts
+++ b/src/angular/components/sso.component.ts
@@ -77,7 +77,7 @@ export class SsoComponent {
 
         const passwordOptions: any = {
             type: 'password',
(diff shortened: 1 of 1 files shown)


cscharf (approved):

kspearrin:
@mportune-bw should update mobile right? There are other places too where this is done. CLI for example.

cscharf:
> should update mobile right? There are other places too where this is done. CLI for example.

let's ensure this even works first or makes things better... then we can update the other clients... CLI, Desktop, etc. should all pull this in from jslib automatically once we update them with latest jslib, mobile will need to manage it separately.

PR #169, opened 2020-09-17
Conditional load of biometrics API on Win10+

cscharf:
Related to and _should_ resolve https://github.com/bitwarden/desktop/issues/507.

## Overview
Apparently just a `try { ... } catch { ... }` alone isn't enough to prevent an app crash for some Windows 7 machines when attempting to load the WinRT node module used for Biometrics support (Windows Hello). Instead we need to simply circumvent that attempted load and just detect the major Windows version ahead of time and not attempt the module load on anything other than Windows 10+ (Windows Hello is not supported in prior versions of Windows, and Windows build 1511 is the earliest build supported by NodeRT module anyway).

Our current version of nodeJS we're using for build + dev doesn't yet support the `os.version()` (returning the build number) method, so we're stuck with `os.release()` (returns the major labeled version, i.e. "7.x", "8.x", 10.x", etc.).

## Screen Shots
Using the following code:
```JavaScript
// C:\AWS\os.js
const os = require('os');
const version = os.release();
console.info('Release: ' + version);
console.info(Number.parseInt(version.split(',')[0]));
```
image

### DIFF
--- a/src/electron/biometric.windows.main.ts
+++ b/src/electron/biometric.windows.main.ts
@@ -113,13 +113,13 @@ export default class BiometricWindowsMain implements BiometricMain {
 
     getWindowsMajorVersion(): number {
         if (process.platform !== 'win32') {
-            return null;
+            return -1;
         }
         try {
             const version = require('os').release();
             return Number.parseInt(version.split('.')[0], 10);
         }
         catch { }
-        return null;
+        return -1;
     }
 }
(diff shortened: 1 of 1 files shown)


src/electron/biometric.windows.main.ts
```diff
@@ -110,4 +110,16 @@ export default class BiometricWindowsMain implements BiometricMain {
         } catch { /*Ignore error*/ }
         return [];
     }
+
+    getWindowsMajorVersion(): number {
+        if (process.platform !== 'win32') {
+            return null;
```
kspearrin, line 116: it seems weird to me to return null on something that expects a number. maybe that's coming from my c# senses. could this return -1 or something instead?

cscharf: Yes, I can do that.

kspearrin (approved):

PR #173, opened 2020-09-22
[SSO] New user provision flow

vincentsalucci:
## Code Changes
- **set-password.component.ts**: Added new string `identifier` for parsing the query params and keep track of the entered org identifier. This is passed through to the API.
- **sso.component.ts**: Added org identifier to the state and saved (regardless of the `state` object's state). Passed along the org identifier to two of the potential routes (2fa, set password). Created function to parse org identifier from the saved `state` object.
- **two-factor.component.ts**: Parsed query params for saved org identifier. Passed along the org identifier to the `set-password` flow.
- **setPasswordRequest.ts**: Added `orgIdentifier` string parameter

### DIFF
--- a/src/angular/components/sso.component.ts
+++ b/src/angular/components/sso.component.ts
@@ -52,7 +52,7 @@ export class SsoComponent {
                 await this.storageService.remove(ConstantsService.ssoCodeVerifierKey);
                 await this.storageService.remove(ConstantsService.ssoStateKey);
                 if (qParams.code != null && codeVerifier != null && state != null && state === qParams.state) {
-                    await this.logIn(qParams.code, codeVerifier);
+                    await this.logIn(qParams.code, codeVerifier, this.getOrgIdentiferFromState(state));
                 }
             } else if (qParams.clientId != null && qParams.redirectUri != null && qParams.state != null &&
                 qParams.codeChallenge != null) {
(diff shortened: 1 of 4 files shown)


src/angular/components/sso.component.ts
```diff
@@ -109,10 +109,14 @@ export class SsoComponent {
...
             }
-
-            await this.storageService.save(ConstantsService.ssoStateKey, state);
         }
 
+        // Add Organization Identifier to state
+        state += `_identifier=${this.identifier}`;
+
+        // Save state (regardless of new or existing)
+        await this.storageService.save(ConstantsService.ssoStateKey, state);
```
kspearrin, line 118: State can be passed in via query string, which I think is why it existed behind the null check. Is there still a need to save it in that case?

vincentsalucci: That is correct - but this is the only place where we're adding on the `orgIdentifier`, so an update to whats in storage is necessary.

vincentsalucci:
Merged `master` because this branch was over 2 weeks stale.

src/angular/components/sso.component.ts
```diff
@@ -167,4 +175,13 @@ export class SsoComponent {
...
         this.loggingIn = false;
     }
+
+    private getOrgIdentiferFromState(state: string): string {
+        if (!state) {
+            return null;
+        }
+
+        const stateSplit = state.split('_identifier=');
+        return stateSplit[1] ? stateSplit[1] : null;
```
cscharf, line 185: Should instead use: `stateSplit.length > 1 ? stateSplit[1] : null;`

cscharf (commented):
Just one small change and then should be good to go

cscharf (approved):

PR #179, opened 2020-09-24
Chmod data.json to 600

ngdangtu-vn:
Nothing scary or serious, this pull request is to complete the https://github.com/bitwarden/cli/issues/160.

I just changed mode to 600 with `fs.chmodSync`. I think `chmodSync` is pretty safe since the `lowdb` lib also use sync functions for both [writing and reading](https://github.com/typicode/lowdb/blob/b70669c12ac0d40e297da38620e210a805c4d8d9/src/adapters/FileSync.js#L5). I'm also aware that the read/write function in `lowdb` is imported from `graceful-fs` lib. However, `graceful-fs` is just improvements to help the `fs` "access more resilient to errors". This means no drastical change should happen and we are safe to use `fs`.

(I did this one because I could. I actually not so sure how value it is. So I hope it won't annoy anyone in here)

### DIFF
--- a/src/services/lowdbStorage.service.ts
+++ b/src/services/lowdbStorage.service.ts
@@ -25,13 +25,11 @@ export class LowdbStorageService implements StorageService {
             adapter = new FileSync(this.dataFilePath);
         }
         try {
-            this.db = lowdb(adapter);
-            fs.chmodSync(this.dataFilePath, "600");
+            this.installDB(adapter);
         } catch (e) {
             if (e instanceof SyntaxError) {
                 adapter.write({});
-                this.db = lowdb(adapter);
-                fs.chmodSync(this.dataFilePath, "600");
+                this.installDB(adapter);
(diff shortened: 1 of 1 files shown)


src/services/lowdbStorage.service.ts
```diff
@@ -26,10 +26,12 @@ export class LowdbStorageService implements StorageService {
         }
         try {
             this.db = lowdb(adapter);
+            fs.chmodSync(this.dataFilePath, "600");
```
kspearrin, line 29: I don't think we want to be setting the permissions on every app init like this. What if the person wanted to manually change it to something else? Wouldn't this reset their changes every time? I would imagine we want to set this only once on file creation.

ngdangtu-vn: Hello @kspearrin 
I've updated new commit as your suggestion.

PR #184, opened 2020-10-08
New method to update the last used index

xusoo:
Instead of updating it every time you call getNext(), it will be updated in a separate call, to avoid updating the index when the cipher did not auto-fill correctly (e.g wrong frame)
Fixes https://github.com/bitwarden/browser/issues/1392


kspearrin (approved):

AriZoNaiCe:
Probably my favorite feature of Bitwarden right now, among many great features.  Thanks, all!

PR #185, opened 2020-10-11
Browser <-> desktop communication

Hinton:
* Changed so that passwords are only stored in desktop (secure storage).
* Resolved a bug where `biometricLocked` was not set to true after manual lock (desktop), which provided access to some things from the lock screen (after first unlock).
* Added constants required for browser biometric unlock.

Required by https://github.com/bitwarden/desktop/pull/566 and https://github.com/bitwarden/browser/pull/1426

### DIFF
--- a/src/angular/components/lock.component.ts
+++ b/src/angular/components/lock.component.ts
@@ -50,7 +50,7 @@ export class LockComponent implements OnInit {
         this.pinSet = await this.vaultTimeoutService.isPinLockSet();
         this.pinLock = (this.pinSet[0] && this.vaultTimeoutService.pinProtectedKey != null) || this.pinSet[1];
         this.supportsBiometric = await this.platformUtilsService.supportsBiometric();
-        this.biometricLock = await this.vaultTimeoutService.isBiometricLockSet() && (await this.cryptoService.hasKey() || this.platformUtilsService.identityClientId !== 'desktop');
+        this.biometricLock = await this.vaultTimeoutService.isBiometricLockSet() && (await this.cryptoService.hasKey() || !this.platformUtilsService.supportsSecureStorage());
         this.biometricText = await this.storageService.get(ConstantsService.biometricText);
         this.email = await this.userService.getEmail();
         let vaultUrl = this.environmentService.getWebVaultUrl();
--- a/src/angular/components/lock.component.ts
+++ b/src/angular/components/lock.component.ts
@@ -158,7 +158,6 @@ export class LockComponent implements OnInit {
         }
         const success = await this.platformUtilsService.authenticateBiometric();
 
-        this.vaultTimeoutService.biometricLocked = !success;
         if (success) {
             await this.doContinue();
         }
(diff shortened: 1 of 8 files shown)


src/services/vaultTimeout.service.ts
```diff
@@ -97,6 +97,7 @@ export class VaultTimeoutService implements VaultTimeoutServiceAbstraction {
             return;
         }
 
+        this.biometricLocked = true;
```
kspearrin, line 100: This a bug fix or something?

cscharf (dismissed):

src/angular/components/lock.component.ts
```diff
@@ -50,7 +50,7 @@ export class LockComponent implements OnInit {
         this.pinSet = await this.vaultTimeoutService.isPinLockSet();
         this.pinLock = (this.pinSet[0] && this.vaultTimeoutService.pinProtectedKey != null) || this.pinSet[1];
         this.supportsBiometric = await this.platformUtilsService.supportsBiometric();
-        this.biometricLock = await this.vaultTimeoutService.isBiometricLockSet() && await this.cryptoService.hasKey();
+        this.biometricLock = await this.vaultTimeoutService.isBiometricLockSet() && (await this.cryptoService.hasKey() || this.platformUtilsService.identityClientId !== 'desktop');
```
kspearrin, line 53: What is the specific reason for the != desktop condition here? This seems like a bit of a hack to use this property. Would it be more appropriate to create some new property that is defined around this condition?

src/services/crypto.service.ts
```diff
@@ -36,14 +37,14 @@ export class CryptoService implements CryptoServiceAbstraction {
...
-        private cryptoFunctionService: CryptoFunctionService) { }
+        private cryptoFunctionService: CryptoFunctionService, private platformUtilService: PlatformUtilsService) { }
 
     async setKey(key: SymmetricCryptoKey): Promise<any> {
         this.key = key;
 
         const option = await this.storageService.get<number>(ConstantsService.vaultTimeoutKey);
         const biometric = await this.storageService.get<boolean>(ConstantsService.biometricUnlockKey);
-        if (option != null && !biometric) {
+        if (option != null && !(biometric && this.platformUtilService.identityClientId === 'desktop')) {
```
kspearrin, line 47: same

src/angular/components/lock.component.ts
Hinton: We want to only store the keys on desktop since the other `jslib` clients don't provide a secure storage. But yea, I agree that abstracting it away would improve readability and provide a context.

How about `this.platformUtilsService.supportsSecureStorage` or `this.platformUtilsService.canSaveKeys`?

cscharf: I like that, `this.platformUtilsService.supportsSecureStorage`

Hinton: Renamed.

cscharf (approved):

PR #191, opened 2020-10-29
Add capability to track window hide

cscharf:
-tbd, draft for example only


src/angular/components/add-edit.component.ts
```diff
@@ -160,6 +163,7 @@ export class AddEditComponent implements OnInit {
             }
         });
         this.writeableCollections = await this.loadCollections();
+        this.broadcasterService.subscribe('hideMain', this.onCurrentWindowHide);
```
kspearrin, line 166: Since this stuff only applies to desktop, I would recommend that this be in the desktop app components, not in the shared jslib ones.

Also, look at how some other componeents handle `subscribe` and `unsubscribe`. The subscriber name is usually the component name and then we parse that for specific event commands.

PR #192, opened 2020-11-02
Some groundwork for Send

kspearrin:
This PR adds the groundwork to jslib for Send so that it can be implemented in other clients (starting with web, PR to come).

- Send APIs and tweaks to error parsing to support password required responses on unauthenticated send endpoints (access)
- Necessary models to parse API requests/responses
- Helpers to create parse send id and keys as URL-base-64 encoded strings
- Make send encryption keys by expanding with HKDF

### DIFF
--- a/src/models/view/sendAccessView.ts
+++ b/src/models/view/sendAccessView.ts
@@ -1,0 +1,24 @@
+import { SendType } from '../../enums/sendType';
+
+import { SendAccess } from '../domain/sendAccess';
+
+import { SendFileView } from './sendFileView';
+import { SendTextView } from './sendTextView';
+import { View } from './view';
+
+export class SendAccessView implements View {
+    id: string = null;
+    name: string = null;
+    type: SendType = null;
+    text = new SendTextView();
+    file = new SendFileView();
+
+    constructor(s?: SendAccess) {
+        if (!s) {
+            return;
+        }
+
+        this.id = s.id;
+        this.type = s.type;
+    }
+}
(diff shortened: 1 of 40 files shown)


src/models/view/sendAccessView.ts
```diff
@@ -0,0 +1,24 @@
...
+    text = new SendTextView();
+    file = new SendFileView();
+
+    constructor(s?: SendAccess) {
+        if (!s) {
+            return;
+        }
+
+        this.id = s.id;
+        this.type = s.type;
```
cscharf, line 22: Wouldn't you also want to set `this.name = s.name`?

src/models/view/sendView.ts
```diff
@@ -0,0 +1,49 @@
...
+    expirationDate: Date = null;
+    password: string = null;
+    disabled: boolean = false;
+
+    constructor(s?: Send) {
+        if (!s) {
+            return;
+        }
+
+        this.id = s.id;
```
cscharf, line 34: Would you also want to set `this.name` and `this.notes` here as well?

src/models/view/sendAccessView.ts
kspearrin: No, the pattern I followed has all non-encrypted things being set in the ctor like this while the encrypted values get decrypted and assigned in another function.

src/models/view/sendView.ts
kspearrin: same

cscharf (approved):

PR #194, opened 2020-11-03
Add verification for password hint - the hint should not equal the pa…

cwille97:
Add verification for password hint - the hint should not equal the password for security reasons. See related pull request on the web repo: https://github.com/bitwarden/web/pull/685

### DIFF
--- a/src/angular/components/register.component.ts
+++ b/src/angular/components/register.component.ts
@@ -112,6 +112,11 @@ export class RegisterComponent {
             }
         }
 
+        if (this.hint === this.masterPassword) {
+            this.platformUtilsService.showToast('error', this.i18nService.t('errorOccurred'), this.i18nService.t('hintEqualsPassword'));
+            return;
+        }
+
         this.name = this.name === '' ? null : this.name;
         this.email = this.email.trim().toLowerCase();
         const kdf = KdfType.PBKDF2_SHA256;


src/angular/components/register.component.ts
```diff
@@ -167,4 +167,11 @@ export class RegisterComponent {
         }
         return userInput;
     }
+
+    verifyPasswordHint() {
+        if (this.hint === this.masterPassword) {
+            this.platformUtilsService.showToast('error', 'Invalid password hint', 'Your password hint cannot be the same as your password.');
+            return;
+        }
```
kspearrin, line 175: This should be moved up with all the other validation on submit.

kspearrin (changes requested):
.

src/angular/components/register.component.ts
cwille97: Updated the MR

src/angular/components/register.component.ts
```diff
@@ -137,6 +137,11 @@ export class RegisterComponent {
             this.platformUtilsService.showToast('success', null, this.i18nService.t('newAccountCreated'));
             this.router.navigate([this.successRoute], { queryParams: { email: this.email } });
         } catch { }
+
+        if (this.hint === this.masterPassword) {
+            this.platformUtilsService.showToast('error', 'Invalid password hint', 'Your password hint cannot be the same as your password.');
+            return;
+        }
```
kspearrin, line 144: this needs to be up higher, where all the other validation checks are.

kspearrin (changes requested):

src/angular/components/register.component.ts
cwille97: Updated

kspearrin: OK, looks good. Last thing now is these strings need to be localized. They can't be hard coded like this. See the others above for an example. We then need to add the localized keys to each project (browser, web, desktop) under the English file so that they can be translated.

cwille97: Updated, adding links for the localized strings

- https://github.com/bitwarden/web/pull/701

kspearrin: I don't see you referencing those strings here?

cwille97: Whoops, I must have pushed the wrong version with git. I'll fix it after work

cwille97: Should be correct now. I'm not sure what went wrong with git but the changes I meant to push were just sitting locally so maybe I just forgot.

cscharf (approved):
@kspearrin , just needs your final approval.

kspearrin (approved):

PR #199, opened 2020-11-09
"Auto-fill on page load" options

eliykat:
These are the changes to jslib required for this feature.

Refer to the main pull request here: https://github.com/bitwarden/browser/pull/1455


eliykat:
This is ready for final review for merging.

cscharf (dismissed):

eliykat:
Just had to cause and then resolve some conflicts

src/models/data/loginData.ts
```diff
@@ -8,6 +8,7 @@ export class LoginData {
     password: string;
     passwordRevisionDate: string;
     totp: string;
+    autofillOnPageLoad: boolean;
```
cscharf, line 11: Does it matter that your server model is `nullable` here?

src/models/domain/login.ts
```diff
@@ -14,6 +14,7 @@ export class Login extends Domain {
     password: EncString;
     passwordRevisionDate?: Date;
     totp: EncString;
+    autofillOnPageLoad: boolean;
```
cscharf, line 17: Same, nullabe?

src/models/data/loginData.ts
eliykat: It's intentional - `true` is autofill, `false` is don't autofill, `null` is treated as "not configured (therefore use default settings)". I originally had an enum for this, but I thought that was overkill when a `bool` would do the job.

src/models/domain/login.ts
eliykat: See explanation above

src/models/data/loginData.ts
cscharf: Yeah, for whatever reason I was thinking you had to define `boolean | null` but you don't, so yeah, it's my dumb 🤦🏻

cscharf (approved):

PR #204, opened 2020-11-18
Add support for Emergency Access

Hinton:
Adds support for the new Emergency Access feature. In short it provides a way to designate an emergency contact whom after a configurable timeout may either takeover the account or be granted view access.

https://app.asana.com/0/1195018406457675/1153292148278672

## Code Changes

* Added code required for Emergency Access
* Changed `cryptoService.remakeEncKey` to accept an `encKey` which is required to do the crypto for account takeover. Also changed `cryptoService.rsaDecrypt` to public.
* Changed `decrypt` methods to accept an optional `encKey` which is used for decrypting the ciphers when viewing another users account.

Relevant PRs:
https://github.com/bitwarden/web/pull/707
https://github.com/bitwarden/server/pull/1000


src/models/domain/attachment.ts
```diff
@@ -50,7 +50,7 @@ export class Attachment extends Domain {
 
             try {
                 const orgKey = await cryptoService.getOrgKey(orgId);
-                const decValue = await cryptoService.decryptToBytes(this.key, orgKey);
+                const decValue = await cryptoService.decryptToBytes(this.key, orgKey ? orgKey : encKey);
```
cscharf, line 53: `orgKey ? orgKey : encKey` should be simplified to `orgKey ?? encKey`

src/models/domain/cipherString.ts
```diff
@@ -105,6 +105,7 @@ export class CipherString {
         }
 
         try {
+            const orgKey = await cryptoService.getOrgKey(orgId);
```
cscharf, line 108: Why is this const declared and method called? Should it throw an exception OR throw away the result?

src/models/domain/cipherString.ts
```diff
@@ -105,6 +105,7 @@ export class CipherString {
...
             if (key == null) {
                 key = await cryptoService.getOrgKey(orgId);
```
cscharf, line 110: Why not here use the const, `orgKey` created earlier vs. calling this again?

src/models/domain/cipherString.ts
Hinton: I suspect this might have been the result of a merge conflict, I know another PR added support for overwriting the key to some models.

cscharf (approved):

PR #205, opened 2020-11-18
send service and syncing send data

kspearrin:
This PR adds support to sync and store sends locally with other vault data.

- Adds send service to centralize send business logic
- Extends sync response model to include list of sends from server
- Store sends locally on sync operation

### DIFF
--- a/src/services/send.service.ts
+++ b/src/services/send.service.ts
@@ -1,0 +1,237 @@
+
+import { SendData } from '../models/data/sendData';
+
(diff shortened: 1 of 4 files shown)


src/services/send.service.ts
```diff
@@ -0,0 +1,237 @@
+
```
cscharf, line 1: Remove leading blank line?

cscharf (approved):

PR #207, opened 2020-11-20
sync: move try-catch out of needsSyncing and handle errors in fullSync

fredrikekre:
The motivation for this is https://github.com/bitwarden/cli/issues/129
where failed sync's are swallowed by try-catch. By moving the try-catch
to the outside it is possible to reuse the already existing
allowThrowOnError argument which callers can use to signal whether
fullSync should throw or ignore errors silently. This patch is
companioned with a patch to the SyncCommand CLI command to pass
allowThrowOnError.

Edit: see https://github.com/bitwarden/cli/pull/190, although the patches are independet of each other.

### DIFF
--- a/src/services/sync.service.ts
+++ b/src/services/sync.service.ts
@@ -76,12 +76,13 @@ export class SyncService implements SyncServiceAbstraction {
         }
 
         const now = new Date();
-        const needsSyncResult = await this.needsSyncing(forceSync);
-        const needsSync = needsSyncResult[0];
-        const skipped = needsSyncResult[1];
-
-        if (skipped) {
-            return this.syncCompleted(false);
+        var needsSync = false;
+        try {
+            needsSync = await this.needsSyncing(forceSync);
+        } catch (e) {
+            if (allowThrowOnError) {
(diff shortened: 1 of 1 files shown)


src/services/sync.service.ts
```diff
@@ -76,12 +76,11 @@ export class SyncService implements SyncServiceAbstraction {
...
-        const needsSync = needsSyncResult[0];
-        const skipped = needsSyncResult[1];
-
-        if (skipped) {
-            return this.syncCompleted(false);
+        var needsSync = false;
+        try {
+            needsSync = await this.needsSyncing(forceSync);
+        } catch (e) {
+            if (allowThrowOnError) { throw e; }
```
cscharf, line 83: Formatting nit, please re-format throw on it's own line:
```
if (...) {
    throw e;
}
```

cscharf (changes requested):
Thanks for your submission! Looks good, simplified; please see 1 small formatting nit.

cscharf (approved):
Looks good, thanks!

PR #214, opened 2020-11-30
Use logService for console messages

MGibson1:
# Objective

Establish a centralized means of handling console logging. Partially resolves bitwarden/cli/issues/188, which calls out the need to shunt all stdout to stderr when the --response flag is used. This is the first step toward that.

# Files changed

* **log.service.ts**: node's console.time is used to measure time taken handling Ciphers and so must be implemented on logService.
* **cli/baseProgram.ts**: `--response` flag is meant to **guarantee** json output to stdout.
* **consoleLogService.ts**: Implement console.time similar to node's library. Output is in ms rather than s+ms.
* **electronLogService.ts**: Same as **consoleLogService.ts**
* **baseImporter.ts**: provide a basic consoleLogService in non-dev mode for importer. We may want to look at determining dev mode here. Use the logService for error reporting as we were previously using console.log/warn.
* **auth.service.ts/crypto.service.ts/notification.service.ts/search.service.ts**: provide optional logService constructor parameter and default to  consoleLogService in non-dev mode. We may want to look into determining dev mode. Use LogService for error reporting as we were previously using console.log/warn. The reason to use an optional parameter is because only `CLI` and `Desktop` use logService at the moment, by defaulting to a consoleLogService, we maintain behavior for `web` and `browser` without requiring code changes.

This commit will require commensurate merges in https://github.com/bitwarden/cli and https://github.com/bitwarden/desktop

### DIFF
--- a/src/services/crypto.service.ts
+++ b/src/services/crypto.service.ts
@@ -5,19 +5,21 @@ import { KdfType } from '../enums/kdfType';
 
 import { CipherString } from '../models/domain/cipherString';
 import { EncryptedObject } from '../models/domain/encryptedObject';
-import { SymmetricCryptoKey } from '../models/domain/symmetricCryptoKey';
 import { ProfileOrganizationResponse } from '../models/response/profileOrganizationResponse';
+import { SymmetricCryptoKey } from '../models/domain/symmetricCryptoKey';
 
 import { CryptoService as CryptoServiceAbstraction } from '../abstractions/crypto.service';
 import { CryptoFunctionService } from '../abstractions/cryptoFunction.service';
+import { ConsoleLogService } from '../cli/services/consoleLog.service';
+import { LogService } from '..//abstractions/log.service';
 import { PlatformUtilsService } from '../abstractions/platformUtils.service';
 import { StorageService } from '../abstractions/storage.service';
 
 import { ConstantsService } from './constants.service';
 
+import { EEFLongWordList } from '../misc/wordlist';
 import { sequentialize } from '../misc/sequentialize';
 import { Utils } from '../misc/utils';
-import { EEFLongWordList } from '../misc/wordlist';
 
 const Keys = {
     key: 'key', // Master Key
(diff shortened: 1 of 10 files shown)


src/services/notifications.service.ts
```diff
@@ -3,13 +3,15 @@ import * as signalRMsgPack from '@microsoft/signalr-protocol-msgpack';
...
 
+import { LogService } from '../abstractions'
 import { ApiService } from '../abstractions/api.service';
 import { AppIdService } from '../abstractions/appId.service';
 import { EnvironmentService } from '../abstractions/environment.service';
 import { NotificationsService as NotificationsServiceAbstraction } from '../abstractions/notifications.service';
 import { SyncService } from '../abstractions/sync.service';
 import { UserService } from '../abstractions/user.service';
 import { VaultTimeoutService } from '../abstractions/vaultTimeout.service';
+import { ConsoleLogService } from '../cli/services/consoleLog.service';
```
kspearrin, line 14: We should not be mixing in CLI services here. Dependencies need to be passed in.

src/services/notifications.service.ts
```diff
@@ -3,13 +3,15 @@ import * as signalRMsgPack from '@microsoft/signalr-protocol-msgpack';
 
 import { NotificationType } from '../enums/notificationType';
 
+import { LogService } from '../abstractions'
```
kspearrin, line 6: Please use fully path like others

kspearrin (changes requested):

src/services/notifications.service.ts
MGibson1: How do you feel about a base ConsoleLogService, which the CLI one will inherit from?

kspearrin: That would be fine

cscharf (dismissed):
Looks good... @kspearrin , just needs your 2nd approval since you last requested changes.

src/services/crypto.service.ts
```diff
@@ -8,10 +8,12 @@ import { EncryptedObject } from '../models/domain/encryptedObject';
 import { SymmetricCryptoKey } from '../models/domain/symmetricCryptoKey';
 import { ProfileOrganizationResponse } from '../models/response/profileOrganizationResponse';
 
+import { LogService } from '..//abstractions/log.service';
```
kspearrin, line 11: why the double slash?

MGibson1: This was a typo from deleting `cli` from the path, should be fixed now.

kspearrin (approved):

PR #216, opened 2020-12-03
support for encrypted json export

kspearrin:
This PR adds support to the export service and components for "encrypted_json" export format.

- Extend all export models to support view and domain models
- Update export service to process encrypted vs decrypted formats

### DIFF
--- a/src/services/export.service.ts
+++ b/src/services/export.service.ts
@@ -29,7 +29,7 @@ export class ExportService implements ExportServiceAbstraction {
 
     async getExport(format: 'csv' | 'json' | 'encrypted_json' = 'csv'): Promise<string> {
         if (format === 'encrypted_json') {
-            return this.getEncryptedExport(format);
+            return this.getEncryptedExport();
         } else {
             return this.getDecryptedExport(format);
         }
@@ -38,7 +38,7 @@ export class ExportService implements ExportServiceAbstraction {
     async getOrganizationExport(organizationId: string,
         format: 'csv' | 'json' | 'encrypted_json' = 'csv'): Promise<string> {
         if (format === 'encrypted_json') {
-            return this.getOrganizationEncryptedExport(organizationId, format);
+            return this.getOrganizationEncryptedExport(organizationId);
         } else {
             return this.getOrganizationDecryptedExport(organizationId, format);
         }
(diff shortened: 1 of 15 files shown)


src/services/export.service.ts
```diff
@@ -175,14 +238,56 @@ export class ExportService implements ExportServiceAbstraction {
...
+                    const cipher = new Cipher(new CipherData(r));
+                    ciphers.push(cipher);
+                });
+            }
+            return Promise.all(cipherPromises);
+        }));
+
+        await Promise.all(promises);
+
+        if (format === 'encrypted_json') {
```
cscharf, line 270: won't the format already be `encrypted_json` at this point?

src/services/export.service.ts
```diff
@@ -97,7 +125,42 @@ export class ExportService implements ExportServiceAbstraction {
         }
     }
 
-    async getOrganizationExport(organizationId: string, format: 'csv' | 'json' = 'csv'): Promise<string> {
+    private async getEncryptedExport(format: 'encrypted_json'): Promise<string> {
+        const folders = await this.folderService.getAll();
+        const ciphers = await this.cipherService.getAll();
+
+        if (format === 'encrypted_json') {
```
cscharf, line 132: Wouldn't the format already be `encrypted_json`, that seems to be the only accepted value/parameter in the method.

src/services/export.service.ts
```diff
@@ -26,7 +27,34 @@ export class ExportService implements ExportServiceAbstraction {
...
+        if (format === 'csv' || format === 'json') {
+            return this.getDecryptedExport(format);
+        } else {
+            return this.getEncryptedExport(format);
+        }
+    }
+
+    async getOrganizationExport(organizationId: string,
+        format: 'csv' | 'json' | 'encrypted_json' = 'csv'): Promise<string> {
+        if (format === 'csv' || format === 'json') {
```
cscharf, line 40: Maybe consider reversing the `if...else` order to manage by exception / shorter list of conditions `if (format === 'encrypted_json')`

src/services/export.service.ts
```diff
@@ -26,7 +27,34 @@ export class ExportService implements ExportServiceAbstraction {
     constructor(private folderService: FolderService, private cipherService: CipherService,
         private apiService: ApiService) { }
 
-    async getExport(format: 'csv' | 'json' = 'csv'): Promise<string> {
+    async getExport(format: 'csv' | 'json' | 'encrypted_json' = 'csv'): Promise<string> {
+        if (format === 'csv' || format === 'json') {
```
cscharf, line 31: Maybe consider reversing the `if...else` order to manage by exception / shorter list of conditions `if (format === 'encrypted_json')`

src/services/export.service.ts
MGibson1, line 132: type check takes care of this exception path

MGibson1: nm, chad is on it :)

kspearrin: yea, i suppose so. I just don't trust the typing sometimes since you can always pass something as type `any` with another value

kspearrin: was also trying to leave room for other format type, like the decrypted options.

cscharf (approved):

PR #221, opened 2020-12-04
Use log service for console messages

MGibson1:
This is a repeat of #214 with a shim layer for browser timing and added (any) tests.

# Objective

Establish a centralized means of handling console logging. Partially resolves bitwarden/cli/issues/188, which calls out the need to shunt all stdout to stderr when the --response flag is used. This is the first step toward that.

# Files changed

* **log.service.ts**: node's console.time is used to measure time taken handling Ciphers and so must be implemented on logService.
* **cli/baseProgram.ts**: `--response` flag is meant to **guarantee** json output to stdout.
* **consoleLogService.ts**: Implement console.time similar to node's library. Output is in ms rather than s+ms.
* **electronLogService.ts**: Same as **consoleLogService.ts**
* **baseImporter.ts**: provide a basic consoleLogService in non-dev mode for importer. We may want to look at determining dev mode here. Use the logService for error reporting as we were previously using console.log/warn.
* **auth.service.ts/crypto.service.ts/notification.service.ts/search.service.ts**: provide optional logService constructor parameter and default to  consoleLogService in non-dev mode. We may want to look into determining dev mode. Use LogService for error reporting as we were previously using console.log/warn. The reason to use an optional parameter is because only `CLI` and `Desktop` use logService at the moment, by defaulting to a consoleLogService, we maintain behavior for `web` and `browser` without requiring code changes.

# Changes from previous commit
Use browser-process-hrtime as a shim between node and browsers. This is well supported on [canIUse](https://caniuse.com/?search=performance.now).

I've also added tests of compatibility and behavior for base class, cli's log service, and electron log service. The base class is working in Karma and other tests are limited to Node.

## Files
* **package-lock.json/package.json**: add browser-process-hrtime and ts-node. ts-node is for local test running.
* ** consoleLog.service.spec.ts**: test for base console log class. Tests browser shim and logging behavior.
* **electronLog.service.spec.ts**: tests unique electron initializations.
* **cli/consoleLog.service.spec.ts**: Tests --response override of log level and behavior of timer.
* **jasmine.json**: include new test locations

This commit will require commensurate merges in /bitwarden/cli#197 and /bitwarden/desktop#602


src/services/consoleLog.service.ts
```diff
@@ -0,0 +1,70 @@
...
+            this.timersMap.set(label, hrtime());
+        }
+    }
+
+    timeEnd(label: string = 'default'): [number, number] {
+        const elapsed = hrtime(this.timersMap.get(label));
+        this.timersMap.delete(label);
+        this.write(LogLevelType.Info, `${label}: ${elapsed[0] * 1000 + elapsed[1] / 10e6}ms`);
+        return elapsed;
+    }
```
kspearrin, line 69: Can you just use `console.time` and `console.timeEnd`? https://developer.mozilla.org/en-US/docs/Web/API/Console/time

Not sure I understand what you're trying to do here.

MGibson1: The reason I don't want to use those methods is they always output to console.log. The idea is full control over the console output so that CLI can guarantee not stdout output if `--response` is set.

package.json
```diff
@@ -75,6 +76,7 @@
     "@microsoft/signalr-protocol-msgpack": "3.1.0",
     "@nodert-win10-rs4/windows.security.credentials.ui": "^0.4.4",
     "big-integer": "1.6.36",
+    "browser-process-hrtime": "1.0.0",
```
kspearrin, line 79: You are going to have to update every browser-based project with this new dependency.

kspearrin (dismissed):

package.json
MGibson1: Thanks for the heads up. I have PRs in for all 4.

src/services/search.service.ts
```diff
@@ -5,15 +5,20 @@ import { CipherView } from '../models/view/cipherView';
 import { CipherService } from '../abstractions/cipher.service';
 import { SearchService as SearchServiceAbstraction } from '../abstractions/search.service';
 
+import { LogService } from '../abstractions/log.service';
```
cscharf, line 8: Should move the abstraction import to the block with the other abstractions.

src/services/search.service.ts
```diff
@@ -5,15 +5,20 @@ import { CipherView } from '../models/view/cipherView';
...
 import { CipherType } from '../enums/cipherType';
 import { FieldType } from '../enums/fieldType';
 import { UriMatchType } from '../enums/uriMatchType';
+import { ConsoleLogService } from '../services/consoleLog.service';
```
cscharf, line 12: Should move the actual service import into its own group.....
Curious, after seeing so many of these; wouldn't the Angular app inject the appropriate log service, as long as each client has provided the init code to new up one and provide it? We don't follow this pattern with any other services where the constructor checks if null and creates otherwise; instead leaving the default instantiation to the overall app (main.ts, etc., class `Main`) and DI.

src/services/notifications.service.ts
```diff
@@ -6,10 +6,12 @@ import { NotificationType } from '../enums/notificationType';
 import { ApiService } from '../abstractions/api.service';
 import { AppIdService } from '../abstractions/appId.service';
 import { EnvironmentService } from '../abstractions/environment.service';
+import { LogService } from '../abstractions/log.service'
 import { NotificationsService as NotificationsServiceAbstraction } from '../abstractions/notifications.service';
 import { SyncService } from '../abstractions/sync.service';
 import { UserService } from '../abstractions/user.service';
 import { VaultTimeoutService } from '../abstractions/vaultTimeout.service';
+import { ConsoleLogService } from '../services/consoleLog.service';
```
cscharf, line 14: Move services import to its own block, below abstractions

src/services/crypto.service.ts
```diff
@@ -10,8 +10,10 @@ import { ProfileOrganizationResponse } from '../models/response/profileOrganizat
 
 import { CryptoService as CryptoServiceAbstraction } from '../abstractions/crypto.service';
 import { CryptoFunctionService } from '../abstractions/cryptoFunction.service';
+import { LogService } from '../abstractions/log.service';
 import { PlatformUtilsService } from '../abstractions/platformUtils.service';
 import { StorageService } from '../abstractions/storage.service';
+import { ConsoleLogService } from '../services/consoleLog.service';
```
cscharf, line 16: Move services import to its own block under abstractions

src/services/auth.service.ts
```diff
@@ -17,11 +17,13 @@ import { AppIdService } from '../abstractions/appId.service';
 import { AuthService as AuthServiceAbstraction } from '../abstractions/auth.service';
 import { CryptoService } from '../abstractions/crypto.service';
 import { I18nService } from '../abstractions/i18n.service';
+import { LogService } from '../abstractions/log.service';
 import { MessagingService } from '../abstractions/messaging.service';
 import { PlatformUtilsService } from '../abstractions/platformUtils.service';
 import { TokenService } from '../abstractions/token.service';
 import { UserService } from '../abstractions/user.service';
 import { VaultTimeoutService } from '../abstractions/vaultTimeout.service';
+import { ConsoleLogService } from '../services/consoleLog.service';
```
cscharf, line 26: Move services import to its own block

src/importers/baseImporter.ts
```diff
@@ -13,13 +13,17 @@ import { FolderView } from '../models/view/folderView';
 import { LoginView } from '../models/view/loginView';
 import { SecureNoteView } from '../models/view/secureNoteView';
 
+import { LogService } from '../abstractions/log.service';
```
cscharf, line 16: Should move the abstraction import to it's own block

src/importers/baseImporter.ts
```diff
@@ -13,13 +13,17 @@ import { FolderView } from '../models/view/folderView';
...
 import { CipherType } from '../enums/cipherType';
 import { FieldType } from '../enums/fieldType';
 import { SecureNoteType } from '../enums/secureNoteType';
+import { ConsoleLogService } from '../services/consoleLog.service';
```
cscharf, line 20: Should move the service import to its own block

src/services/search.service.ts
MGibson1: Web and Browser never implement a ConsoleLogService This was a cheeky thing to try and avoid changing those repos. Since I need to add `browser-process-hrtime` to those repos, that obviously hasn't worked. 

I also think that this service is a bit of a special case. If I don't care about `console`, I'd rather not think about it than have to provide one explicitly. I can get angular to supply these if we want to go that route for congruity.

cscharf: Yes, let's have angular supply these through the normal course of DI and for congruity (nice word). Thanks Matt!

cscharf (approved):

PR #222, opened 2020-12-08
Fix 1password importer

MGibson1:
# Objective

Fix the 1password windows importer and add support for exports from the mac desktop client.

# Methodology

Most properties can be shared between mac and windows. Only the divergence needs to be specified in either. This can be done either through overloading or removing the parser that is failing, or by adding another that will handle the exceptional case.

# Files changed
* **onepassword\*.spec.ts**: spec files testing correct import of identity and credit card ciphers from 1password exports. Tests windows and mac exports for Identity, Credit Card, and multiple type exports
* **CipherImportContext**: POCO object containing context data needed for parsing a csv entry column to cipher property.
* **onepassword1pifImporter**: move to onepasswordImporters folder.
* **onepasswordCsvImporter**: base onepassword csv importer. All row column parsers are methods that can be overloaded to change behavior. These lists of default parsers can also be added to by child classes.
* **onepasswordMacCsvImporter**: child importer class customized for importing csv generated from 1password desktop on mac.
* **onepasswordWinCsvImporter**: child importer class customized for importing csvs generated from 1password desktop on windows.


src/importers/onepasswordImporters/onepasswordCsvImporter.ts
```diff
@@ -0,0 +1,293 @@
...
+import { CipherImportContext } from './cipherImportContext';
+import { FieldType } from '../../enums';
+
+export const IgnoredProperties = ['ainfo', 'autosubmit', 'notesplain', 'ps', 'scope', 'tags', 'title', 'uuid', 'notes'];
+
+export abstract class OnePasswordCsvImporter extends BaseImporter implements Importer {
+    abstract setCipherType(value: any, cipher: CipherView): void;
+    protected loginPropertyParsers = [this.setLoginUsername, this.setLoginPassword, this.setLoginUris];
+    protected creditCardPropertyParsers = [this.setCreditCardNumber, this.setCreditCardVerification, this.setCreditCardCardholderName, this.setCreditCardExpiry];
+    protected identityPropertyParsers = [this.setIdentityFirstName, this.setIdentityInitial, this.setIdentityLastName, this.setIdentityUserName, this.setIdentityEmail, this.setIdentityPhone, this.setIdentityCompany]
```
cscharf, line 16: Missing semi-colon.

src/importers/onepasswordImporters/onepasswordCsvImporter.ts
```diff
@@ -0,0 +1,293 @@
...
+            result.success = false;
+            return Promise.resolve(result);
+        }
+
+        results.forEach((value) => {
+            if (this.isNullOrWhitespace(this.getProp(value, 'title'))) {
+                return;
+            }
+
+            const cipher = this.initLoginCipher();
```
cscharf, line 35: I may be mistaken, but I believe this should only be called (where the `LoginView` is set) when the cipher type is a Login. If you look at other implementations, only the View of the given type is set generally (LoginView/CardView/etc.). I didn't see anywhere this was unset for alternate scenarios (but I could be wrong).

src/importers/onepasswordImporters/onepasswordMacCsvImporter.ts
```diff
@@ -0,0 +1,29 @@
+import { IgnoredProperties, OnePasswordCsvImporter } from './onepasswordCsvImporter'
+import { Importer } from '../importer';
+
+
```
cscharf, line 4: remove extra blank line.

src/importers/onepasswordImporters/onepasswordCsvImporter.ts
```diff
@@ -0,0 +1,293 @@
...
+import { FieldType } from '../../enums';
+
+export const IgnoredProperties = ['ainfo', 'autosubmit', 'notesplain', 'ps', 'scope', 'tags', 'title', 'uuid', 'notes'];
+
+export abstract class OnePasswordCsvImporter extends BaseImporter implements Importer {
+    abstract setCipherType(value: any, cipher: CipherView): void;
+    protected loginPropertyParsers = [this.setLoginUsername, this.setLoginPassword, this.setLoginUris];
+    protected creditCardPropertyParsers = [this.setCreditCardNumber, this.setCreditCardVerification, this.setCreditCardCardholderName, this.setCreditCardExpiry];
+    protected identityPropertyParsers = [this.setIdentityFirstName, this.setIdentityInitial, this.setIdentityLastName, this.setIdentityUserName, this.setIdentityEmail, this.setIdentityPhone, this.setIdentityCompany]
+
```
cscharf, line 17: Remove extra blank line.

cscharf (commented):
couple of small things and one question(ish)

src/importers/onepasswordImporters/onepasswordCsvImporter.ts
MGibson1: I don't think so, the method creates a new `LoginView` itself.

I also pirated this directly from the [`onepasswordWinCsvImporter`](https://github.com/bitwarden/jslib/blob/c9df039fa9f392462130b733b4a724a024672866/src/importers/onepasswordWinCsvImporter.ts#L25-L33) before I started messing with it. It's possible this one is an odd duck and we should be following your suggestions.

Items imported without removing the other views still render only the relevant view types as evidenced by the following screenshot, which was imported using this code (I haven't checked all clients, just web and browser, but I think this rendering is in jslib):
image

I like this method since it fully initializes everything required to make the cipher viewable. I could create an `initCipher` helper that wouldn't set the type, but then we're relying on downstream methods to make the cipher viable

cscharf (commented):
Looks good to me, will review once more after out of draft

cscharf (approved):

PR #235, opened 2020-12-19
Add getManyByCipherId method to collection service

mauamy:
This PR is related to https://github.com/bitwarden/web/pull/647
Like recommended I created a service method for getting all collections by a given cipher id to the collection service.

### DIFF
--- a/src/services/collection.service.ts
+++ b/src/services/collection.service.ts
@@ -5,6 +5,7 @@ import { TreeNode } from '../models/domain/treeNode';
 
 import { CollectionView } from '../models/view/collectionView';
 
+import { CipherService } from '../abstractions';
 import { CollectionService as CollectionServiceAbstraction } from '../abstractions/collection.service';
 import { CryptoService } from '../abstractions/crypto.service';
 import { I18nService } from '../abstractions/i18n.service';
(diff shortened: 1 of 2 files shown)


src/services/collection.service.ts
```diff
@@ -98,6 +99,19 @@ export class CollectionService implements CollectionServiceAbstraction {
         return this.decryptedCollectionCache;
     }
 
+    async getManyByCipherId(cipherId: string): Promise<Collection[]> {
+        const cipher = await this.cipherService.get(cipherId)
```
cscharf, line 103: missing semicolon, `;`

cscharf (commented):
Just one small formatting item, otherwise looks good, thank you!

**NOTE** on this, since this updates/changes the abstraction contract in jslib, this could cause breaking changes in any override implementations and constructor calls of that abstraction and base implementation of that service in other client Repos.
You have `web` covered, but there's also `desktop`, `browser` and `directory-connector`. Can you please do a quick scan and see if those would be impacted and need quick updates for this interface/abstraction contract change?

eliykat:
Closing due to inactivity, see my explanation on https://github.com/bitwarden/web/pull/647

PR #239, opened 2020-12-25
Fix/csv export note newline

Goblin80:
During CSV export, notes with new lines break the CSV structure.

This PR aims at escaping every newline in the notes during export.

https://github.com/bitwarden/web/issues/752

### DIFF
--- a/src/models/export/cipher.ts
+++ b/src/models/export/cipher.ts
@@ -112,7 +112,7 @@ export class Cipher {
 
         if (o instanceof CipherView) {
             this.name = o.name;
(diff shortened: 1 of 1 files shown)


kspearrin:
I am not sure there is anything wrong with having new lines in a CSV as long as they are wrapped in quotes, which they are.

cscharf:
CSV specifications escape newlines, delimiters and text qualifiers (double double-quotes) with double quotes (text qualified). This works, for example, if you were to open that CSV in Excel or import into Sheets, etc. it would work as expected.

PR #243, opened 2021-01-04
Update revision date upon cipher restore

MGibson1:
# Overview

https://app.asana.com/0/1169444489336079/1199337616674667/f
related to bitwarden/web/issues/748

Cipher edit revision date validation is failing for ciphers which were restored from the trash. This is because restores update revisionDate on the server side, but do not send an updated cipher back to the client.

These changes force a call to the server after restore to get the new revisionDate of the cipher.

# Files Changed
* **api.service.ts**: Change signature to expect response from restore calls. (bitwarden/server#1072)
* **cipher.service.ts**: Update local cipher db to match ciphers returned from restore calls

### DIFF
--- a/src/abstractions/cipher.service.ts
+++ b/src/abstractions/cipher.service.ts
@@ -53,7 +53,7 @@ export abstract class CipherService {
     softDelete: (id: string | string[]) => Promise<any>;
     softDeleteWithServer: (id: string) => Promise<any>;
     softDeleteManyWithServer: (ids: string[]) => Promise<any>;
-    restore: (id: string | string[]) => Promise<any>;
+    restore: (cipher: { id: string, revisionDate: string; } | { id: string, revisionDate: string; }[]) => Promise<any>;
     restoreWithServer: (id: string) => Promise<any>;
     restoreManyWithServer: (ids: string[]) => Promise<any>;
 }
--- a/src/services/cipher.service.ts
+++ b/src/services/cipher.service.ts
@@ -876,7 +876,7 @@ export class CipherService implements CipherServiceAbstraction {
         await this.softDelete(ids);
     }
 
-    async restore(id: string | string[]): Promise<any> {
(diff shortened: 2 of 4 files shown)


src/services/cipher.service.ts
```diff
@@ -903,12 +905,18 @@ export class CipherService implements CipherServiceAbstraction {
...
-        await this.restore(id);
+        const response = await this.apiService.getCipher(id);
+        await this.restore({ id: id, revisionDate: response.revisionDate });
     }
 
     async restoreManyWithServer(ids: string[]): Promise<any> {
         await this.apiService.putRestoreManyCiphers(new CipherBulkRestoreRequest(ids));
-        await this.restore(ids);
+        const restores: { id: string, revisionDate: string; }[] = [];
+        for (const id of ids) {
```
MGibson1, line 915: this feels hacky to me. Should we just change the server signature to return the updated ciphers?

cscharf: I like that much better... shouldn't hurt reverse compatibility in that way since the client can ignore returned JSON (older client versions) but newer client versions can take advantage of it. Any time we're hitting an API (or database) in a loop it makes my skin crawl a bit

src/abstractions/cipher.service.ts
```diff
@@ -53,7 +53,7 @@ export abstract class CipherService {
     softDelete: (id: string | string[]) => Promise<any>;
     softDeleteWithServer: (id: string) => Promise<any>;
     softDeleteManyWithServer: (ids: string[]) => Promise<any>;
-    restore: (id: string | string[]) => Promise<any>;
```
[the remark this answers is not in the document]
MGibson1: Nope, it looks like this could be removed from the interface and made private in the service and no client would care.

Shall I do that?

src/services/cipher.service.ts
MGibson1: The server changes are in bitwarden/server#1072

cscharf (approved):
Looks good, may just want to wait until server update is approved/merged before merging this one.

PR #246, opened 2021-01-12
update package https-proxy-agent

cscharf:
See: https://github.com/bitwarden/cli/issues/181

This resolves the upstream error for the CLI being able to access via proxy with the error, `Protocol "https:" not supported`.

Current issue is: `jasmine` tests in the karma runner fail with the npm package update... haven't been able to trace down the reason, but this would need to be resolved/figured out before moving forward with this update 🙁


kspearrin (dismissed):

eliykat:
The problem seems to be:
- the karma spec uses the `@fluffy-spoon/substitute` module to create mock objects
- in turn, `substitute` imports node's `util` module.
- however, the `util` module does not consistently get bundled by `karma-typescript` when running browser tests. This causes the `require("util")` statement to silently fail, then it throws an error the first time it tries to call a method from `util`.

I think it's some kind of race condition, because it behaves really inconsistently - I thought I'd solved it 3 or 4 times today just by changing different things.

By importing `util` in the spec, it seems to have forced `karma-typescript` to include it in the bundle. It's not really the right way to fix it though, I'll have a look at karma's docs tomorrow.

eliykat:
Alright, this git history got a little messy 😅 Let me know if you want me to tidy it up somehow. 

The problem is definitely with resolving and/or bundling the `util` module. I noticed that in `master`, the `karma-typescript` bundler resolves `util` imports to `node_modules/util`, but in this branch, it had started resolving to  `node_modules/karma-typescript/node_modules/util`.

Adding an explicit resolution alias in `karma.conf.js`, so it resolves to the top-level `util` module, has fixed it.

addisonbeck (approved):

cscharf:
Gah!!! Build failed, https://github.com/bitwarden/jslib/runs/1781119657?check_suite_focus=true, @eliykat , would you mind please taking a peek?

eliykat:
Note for posterity: I suspect this may have been caused by merge conflicts in `package-lock.json`, but re-running the tests fixed it, so I'm not doing any further investigation at this stage.

PR #247, opened 2021-01-13
Update electron to 11.1.1

Hinton:
Update electron to 11.1.1.

Relevant electron deprecation:

 * process.windowStore is now read-only, which required some minor refactoring in `src/electron/utils.ts`.
 * `app.getName()` is deprecated and has been replaced by `app.name`.


src/electron/utils.ts
```diff
@@ -16,11 +16,12 @@ export function isMacAppStore() {
...
     const isWindows = process.platform === 'win32';
-    if (isWindows && !process.windowsStore &&
+    let windowsStore = process.windowsStore;
+    if (isWindows && !windowsStore &&
         process.resourcesPath.indexOf('8bitSolutionsLLC.bitwardendesktop_') > -1) {
-        process.windowsStore = true;
+        windowsStore = true;
     }
-    return isWindows && process.windowsStore && process.windowsStore === true;
+    return isWindows && windowsStore && windowsStore === true;
```
cscharf, line 24: Can't this be simplified down to `return isWindows && windowsStore === true;` ?

cscharf (commented):
Looks good, one boolean logic nit, but otherwise will hold until after Jan release.

src/electron/utils.ts
flying-sausages: is this nitpick the last thing holding off native possible Apple Silicon support?

cscharf: Hi @flying-sausages, no, this is not holding anything up, we just were waiting for our prior release before approving/merging this in. This nit was just general feedback.

cscharf (dismissed):

src/electron/utils.ts
flying-sausages: Sorry, my phrasing wasn't the most appropriate! Thanks for the clarification :)

Hinton: @cscharf Probably, can't see any case where this simplification would not work.

cscharf (approved):

PR #252, opened 2021-01-28
Add a flag for viewOnly to AttachmentsComponent

Hinton:
Needed since we only want to view attachments in emergency access.


kspearrin (approved):

kspearrin (changes requested):
Should this just be in the web project since we only do emergency access recovery there?

Hinton:
@kspearrin Hmm, we can probably keep it in the web vault for now.

PR #253, opened 2021-01-29
Add send to cli

MGibson1:
# Overview

These are changes related to Send in CLI (PR incoming). 

# Files Changed

* **api.service.ts**: CLI needs to be able to access api's other than those configured for it, so ApiService's PostSendAccess accepts an outside url.
* **send.service.ts**: CLI does not have a file object, so instead sends in an ArrayBuffer directly as the contents of any files uploaded.
* **login.command.ts/update.command.ts**: A result of upgrading CLI to commander 7.0.0, options are no longer on a command object, but an OptionsValues object. This is just a refactor and shouldn't have any impact.
* **Utils.ts**: These helpers allow for compile-time errors in case of property renames. It emulates what you get from c#'s `nameof(property)`. The assign helper limits possible assignments to overriding values already on the given object template, which is a very common use.
* **search.service.ts**: Add the ability to find sends by access id
* **nodeUtils.ts**: helper to conver Node Buffer to javascript BufferArray

# Testing Considerations

Testing will largely be handled in CLI PR, but It will be prudent to run basic testing on web vault Sends to ensure these are all benign changes there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1680,24 +1680,32 @@
           "dependencies": {
             "abbrev": {
               "version": "1.1.1",
+              "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+              "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
               "bundled": true,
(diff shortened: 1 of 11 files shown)


cscharf (approved):
I almost got a little lost in the refactor for `send.service.ts` but then saw the DRY pattern emerge around the file data and name encryption, had my ah-ha moment and then clicked Approve 😁

cscharf (approved):

PR #254, opened 2021-01-29
[Send] Port web based components

vincentsalucci:
## Code Changes
- send/add-edit.component.ts: Full port of existing web based component.  
- send/send.component.ts: Ported existing web base component to jslib. Added `onSuccessfulRemovePassword` and `onSuccessfulDelete` callbacks in case other platforms perform these actions within different components

## Testing Concerns
- Once patched with the web client, make sure existing send features work as they did initially

### DIFF
--- a/src/angular/components/send/send.component.ts
+++ b/src/angular/components/send/send.component.ts
@@ -1,0 +1,195 @@
+import {
+    NgZone,
+    OnInit,
+} from '@angular/core';
+
+import { SendType } from '../../../enums/sendType';
+
+import { SendView } from '../../../models/view/sendView';
+
+import { EnvironmentService } from '../../../abstractions/environment.service';
+import { I18nService } from '../../../abstractions/i18n.service';
+import { PlatformUtilsService } from '../../../abstractions/platformUtils.service';
+import { SendService } from '../../../abstractions/send.service';
+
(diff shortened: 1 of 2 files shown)


src/angular/components/send/send.component.ts
```diff
@@ -0,0 +1,195 @@
...
+            this.filteredSends = this.sends.filter((s) => this.filter == null || this.filter(s));
+            return;
+        }
+        this.searchPending = true;
+        this.searchTimeout = setTimeout(async () => {
+            this.filteredSends = this.sends.filter((s) => this.filter == null || this.filter(s));
+            this.searchPending = false;
+        }, timeout);
+    }
+
```
cscharf, line 110: Missing `addSend()` and `editSend()` methods, or at least interfaces/base impl here? Not sure how modal components or child components would act here, but perhaps those really don't belong in jslib (they may not, just not sure why they're omitted).

cscharf (approved):

src/angular/components/send/send.component.ts
vincentsalucci: I originally had them here but empty (with a comment) since it appears almost all actions for both of those will be handled entirely by the platform. What's the better route here?

vincentsalucci: Leaving these methods out for now since the child classes will handle all implementation.

PR #256, opened 2021-02-01
Enable subpixel antialiasing

Lombra:
Enables subpixel font antialiasing by explicitly setting Electron background color.

https://www.electronjs.org/docs/faq#the-font-looks-blurry-what-is-this-and-what-can-i-do

### DIFF
--- a/src/electron/window.main.ts
+++ b/src/electron/window.main.ts
@@ -114,6 +114,7 @@ export class WindowMain {
             icon: process.platform === 'linux' ? path.join(__dirname, '/images/icon.png') : undefined,
             titleBarStyle: this.hideTitleBar && process.platform === 'darwin' ? 'hiddenInset' : undefined,
             show: false,
+            backgroundColor: '#fff',
             alwaysOnTop: this.enableAlwaysOnTop,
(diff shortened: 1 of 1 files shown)


src/electron/window.main.ts
```diff
@@ -114,6 +114,7 @@ export class WindowMain {
             icon: process.platform === 'linux' ? path.join(__dirname, '/images/icon.png') : undefined,
             titleBarStyle: this.hideTitleBar && process.platform === 'darwin' ? 'hiddenInset' : undefined,
             show: false,
+            backgroundColor: '#fff',
```
cscharf, line 117: Will this override/overrule CSS rules based on theming (dark vs. light vs. other, etc.)? That's my only concern with this change.

Lombra: I have not noticed anything breaking in themes. The parameter does not seem to actually have a visible effect on any background color. (even with a completely styleless app, setting the background to black using this parameter doesn't visibly do so) Setting CSS background color on html, body, etc works as usual.

cscharf (approved):

PR #258, opened 2021-02-03
Send search

addisonbeck:
It looks like fuzzy searching sends isn't working, even in web. This PR wires that up.

### DIFF
--- a/src/angular/components/send/send.component.ts
+++ b/src/angular/components/send/send.component.ts
@@ -59,7 +59,6 @@ export class SendComponent implements OnInit {
         });
 
         await this.load();
-
     }
 
     ngOnDestroy() {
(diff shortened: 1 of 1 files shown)


src/angular/components/send/send.component.ts
```diff
@@ -10,6 +10,7 @@ import { SendView } from '../../../models/view/sendView';
 import { EnvironmentService } from '../../../abstractions/environment.service';
 import { I18nService } from '../../../abstractions/i18n.service';
 import { PlatformUtilsService } from '../../../abstractions/platformUtils.service';
+import { SearchService } from '../../../abstractions';
```
cscharf: Need to use the full path here to the `search.service` module to follow convention.

cscharf (changes requested):

src/angular/components/send/send.component.ts
addisonbeck: Fixed, along with the other linter error

cscharf (approved):
😏 , nice way to steal 3rd base there at the last minute Addison

PR #259, opened 2021-02-04
Implement disable send policy

MGibson1:
# Overview

Add disable send logic to Send components and define `DisableSend` policy type.

This policy will be used by organizations to disable organization users from creating or editing Bitwarden Sends. Delete is still allowed.

# Files Changed

* **add-edit.component.ts/send.component.ts**: If any policy on any organization the user belongs to is enabled and that user cannot manage policies on that organization, send is disabled.
* **policyType.ts**: Add enum number for DisableSend policy

# Testing concerns

Logic for the disable will be tested in the first client to implement this policy, which will be web.

### DIFF
--- a/src/angular/components/send/send.component.ts
+++ b/src/angular/components/send/send.component.ts
@@ -4,6 +4,8 @@ import {
 } from '@angular/core';
 
 import { SendType } from '../../../enums/sendType';
+import { PolicyType } from '../../../enums/policyType';
+import { OrganizationUserStatusType } from '../../../enums/organizationUserStatusType';
 
 import { SendView } from '../../../models/view/sendView';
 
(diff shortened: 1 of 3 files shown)


src/angular/components/send/send.component.ts
```diff
@@ -199,7 +215,7 @@ export class SendComponent implements OnInit {
 
     private applyTextSearch() {
         if (this.searchText != null) {
-           this.filteredSends = this.searchService.searchSends(this.filteredSends, this.searchText);
+            this.filteredSends = this.searchService.searchSends(this.filteredSends, this.searchText);
```
addisonbeck, line 218: this looks accidental

MGibson1: It's a linter fix. One of the old tabs had only 3 spaces

cscharf (approved):

PR #262, opened 2021-02-08
[Send] Add more flexibility to base component

vincentsalucci:
## Code Changes
- **send/send.component**: Added `hasSearched` tracking boolean that is used for showing search results in the same component, smoothly. Added anonymous function for `onSuccessfulLoad` in order to run some client-side calculations before considering the data `loaded`. Removed call to `load` in `ngOnInit` because additional client-side operations need to take place before calling `load` in the browser extension.

## Breaking Change
- The call to remove `load` from `onInit` will need to be updated and re-implemented for the child classes in the `web` and `desktop` repositories immediately following the merge of this PR

### DIFF
--- a/src/angular/components/send/send.component.ts
+++ b/src/angular/components/send/send.component.ts
@@ -38,7 +38,7 @@ export class SendComponent implements OnInit {
     searchPlaceholder: string;
     filter: (cipher: SendView) => boolean;
     searchPending = false;
-    hasSearched = false;
+    hasSearched = false; // search() function called - returns true if text qualifies for search
 
     actionPromise: any;
     onSuccessfulRemovePassword: () => Promise<any>;
--- a/src/angular/components/send/send.component.ts
+++ b/src/angular/components/send/send.component.ts
@@ -38,10 +38,12 @@ export class SendComponent implements OnInit {
     searchPlaceholder: string;
     filter: (cipher: SendView) => boolean;
     searchPending = false;
+    hasSearched = false;
 
     actionPromise: any;
     onSuccessfulRemovePassword: () => Promise<any>;
     onSuccessfulDelete: () => Promise<any>;
+    onSuccessfulLoad: () => Promise<any>;
 
     private searchTimeout: any;
 
@@ -73,8 +75,6 @@ export class SendComponent implements OnInit {
                 }
             });
         });
-
-        await this.load();
     }
 
     ngOnDestroy() {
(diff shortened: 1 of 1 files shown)


src/angular/components/send/send.component.ts
```diff
@@ -85,6 +85,9 @@ export class SendComponent implements OnInit {
         this.loading = true;
         const sends = await this.sendService.getAllDecrypted();
         this.sends = sends;
+        if (this.onSuccessfulLoad() != null) {
```
cscharf, line 88: This will execute `this.onSuccessfulLoad()` and evaluate the result (`!= null`), not sure that's intended since you're executing it again below. I believe you need to remove the `()` parens at the end to test if the delegate/method is `null` vs. the result of evaluation/execution of the same.

src/angular/components/send/send.component.ts
```diff
@@ -116,12 +119,14 @@ export class SendComponent implements OnInit {
             clearTimeout(this.searchTimeout);
         }
         if (timeout == null) {
+            this.hasSearched = this.searchService.isSearchable(this.searchText);
```
cscharf, line 122: Perhaps I misunderstood the purpose of `this.hasSearched`. My understanding is this was going to track whether or not a search was performed, where here it appears to be tracking whether or not a search _can_ be performed.

Perhaps this needs to go below `this.applyTextSearch();`? Not entirely sure.

src/angular/components/send/send.component.ts
```diff
@@ -116,12 +119,14 @@ export class SendComponent implements OnInit {
...
             this.filteredSends = this.sends.filter(s => this.filter == null || this.filter(s));
             this.applyTextSearch();
             return;
         }
         this.searchPending = true;
         this.searchTimeout = setTimeout(async () => {
+            this.hasSearched = this.searchService.isSearchable(this.searchText);
```
cscharf, line 129: Same.

src/angular/components/send/send.component.ts
vincentsalucci: Great catch! Updated this delegate and the others as well.

src/angular/components/send/send.component.ts
vincentsalucci: From what I can see - you're right on both accounts :D 

The variable name `hasSearched` is signifying that the function `search(...)` has been called but its context does indeed prove that the search text _can_ and _should_ be acted upon via the `searchService` (next line).

I'm okay with leaving it or changing the name to be more concise!

cscharf: I dunno, `canAndLikelyHasSearched` seems a bit too wordy. This is probably okay as-is now that I understand the context a bit more. As long as nobody modifies other code later in the UI to cascade UI changes based on this bit where the search function may fail, etc. and/or have unexpected results (timing/sequence). But probably good for now; just some code comments would be good I think where it's declared (in this instance).

vincentsalucci: I <3 comments

cscharf (approved):

PR #268, opened 2021-02-09
Move LogService to static singleton

MGibson1:
Use LogService to output all caught errors to console

# Overview

We have a lot of locations we are silently swallowing Errors. This PR is a first step toward better surfacing of those errors.

I've moved LogService to a static construct that is initialized with other Services to a singleton. If a client does not initialize, a default `ConsoleLogService` is instantiated.

Typescript client PRs incoming to use static `LogService` methods

# Files Changed

* **abstract/log.service.ts**: method signatures are changed to match `console` signatures and static methods are provided. Clients should use the `LogService.set()` method to set the LogService instance to use during whatever service instantiation they do.
* **auth.service/crypto.service/notifications.service/search.service**: Removed logService parameter dependency in favor of static methods.

Other file changes are adding log errors for caught exceptions of converting to static LogService methods

### DIFF
--- a/src/abstractions/log.service.ts
+++ b/src/abstractions/log.service.ts
@@ -4,7 +4,7 @@ import { ConsoleLogService } from '../services/consoleLog.service';
 export abstract class LogService {
 
     private static get instance() {
-        if (this._instance) {
+        if (this._instance == null) {
             this._instance = new ConsoleLogService(false);
         }
         return this._instance;
--- a/src/abstractions/log.service.ts
+++ b/src/abstractions/log.service.ts
@@ -1,11 +1,49 @@
 import { LogLevelType } from '../enums/logLevelType';
+import { ConsoleLogService } from '../services/consoleLog.service';
 
 export abstract class LogService {
-    debug: (message: string) => void;
-    info: (message: string) => void;
-    warning: (message: string) => void;
-    error: (message: string) => void;
-    write: (level: LogLevelType, message: string) => void;
+
+    private static get instance() {
+        if (this._instance) {
+            this._instance = new ConsoleLogService(false);
+        }
+        return this._instance;
+    }
+
+    static set(logService: LogService) {
+        this._instance = logService;
+    }
+
+    static debug(message?: any, ...optionalParams: any[]) {
+        this.instance.debug(message, optionalParams);
+    }
+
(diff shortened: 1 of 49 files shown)


src/electron/biometric.windows.main.ts
```diff
@@ -113,7 +118,10 @@ export default class BiometricWindowsMain implements BiometricMain {
                     module.UserConsentVerifierAvailability.deviceBusy,
                 ];
             }
-        } catch { /*Ignore error*/ }
+        } catch (e) {
+            LogService.error(e);
+            /*Ignore error*/
```
cscharf, line 123: We can remove this comment. We're no longer ignoring it 😁

src/misc/utils.ts
```diff
@@ -1,5 +1,7 @@
 import * as tldjs from 'tldjs';
 
+import { LogService } from '../abstractions/log.service';
+
```
cscharf, line 4: Why not combine into a group from `abstractions`?

src/abstractions/log.service.ts
```diff
@@ -1,11 +1,49 @@
...
 export abstract class LogService {
-    debug: (message: string) => void;
-    info: (message: string) => void;
-    warning: (message: string) => void;
-    error: (message: string) => void;
-    write: (level: LogLevelType, message: string) => void;
+
+    private static get instance() {
+        if (this._instance == null) {
+            this._instance = new ConsoleLogService(false);
```
cscharf, line 8: I know this provides a really nice fallback, and I get that, however I'm struggling with having a Singleton pattern implementation of this abstract type which has a circular reference to a descendent type/default implementation of itself. Kind of outside of established OOD principals of encapsulation and inheritance (although a great example of polymorphism 😁 ). 

It seems sucky, I know, but I would have more warm & fuzzies if we simply left `this._instance` as `null` if not explicitly set anywhere by a downstream runtime/client, and following the "quasi" DI pattern, each client app is responsible for and **must** set this in order for logging to be "a thing"; otherwise calls to `.debug`, `.error`, etc. are simply ignored within the static methods if no logging instance is set. I would simply like to avoid circular references in the dependency chain.

MGibson1:
Not going to do before jslib more for sure

PR #272, opened 2021-02-12
refactored a few properties into getters

addisonbeck:
Refactored two simple Send properties (`title` and `editMode`) into getters for better reuse & change detection. This update shouldn't impact any clients unless they want it to.

### DIFF
--- a/src/angular/components/send/add-edit.component.ts
+++ b/src/angular/components/send/add-edit.component.ts
@@ -35,9 +35,7 @@ export class AddEditComponent implements OnInit {
 
     copyLink = false;
     disableSend = false;
-    editMode: boolean = false;
     send: SendView;
-    title: string;
     deletionDate: string;
     expirationDate: string;
     hasPassword: boolean;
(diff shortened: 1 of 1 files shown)


src/angular/components/send/add-edit.component.ts
```diff
@@ -94,15 +92,19 @@ export class AddEditComponent implements OnInit {
...
-    async load() {
-        this.editMode = this.sendId != null;
-        if (this.editMode) {
-            this.editMode = true;
-            this.title = this.i18nService.t('editSend');
-        } else {
-            this.title = this.i18nService.t('createSend');
-        }
+    get editMode(): boolean {
+        return this.sendId != null;
```
cscharf, line 96: Is this something that changes within the same context of this screen/dialog? When it's saved will the dialog stay open where this would be prevalent (create and then immediately edit)? Trying to understand the reason this is needed.

cscharf (commented):
Looks good, just a quick question

src/angular/components/send/add-edit.component.ts
addisonbeck: In desktop since there isn't a modal that is constantly being created/destroyed it makes sense to reuse the same `SendAddEditComponent` if possible, so yes. If I click on a Send to edit it and then pivot to the Add Send button the only change that really needs to be made to reuse the same component that is already displayed is clearing the form and letting Angular update these display variables. For web where there is a modal being created & destroyed on every run this is not really changing anything besides cleaning out stuff that doesn't have to be in `load()` method.

cscharf: Okay, thanks for that @addisonbeck , sometimes it's easy to forget about that, appreciate the follow-up.

cscharf (approved):

PR #273, opened 2021-02-12
Lock lowdb storage file to avoid dirty data collisions

MGibson1:
# Overview

linked to bitwarden/directory-connector/issues/41.
linked to biwarden/directory-connector/

There appear to be issues with lowdb colliding in cron with previous jobs and wiping out databases (typicode/lowdb/issues/277). I have been unable to force this myself, but locking the lowdb `data.json` file should resolve this if it's what's causing our wiping issue.

The method of locking is to create a file, `data.json.lock`, to indicate the file is being accessed. This implementation creates this lock on all db reads and writes regardless of whether it is a cached read. proper-lockfile updates the modified time of the lock file periodically to signal that the lock is still live and valid. Locks older than a stale lock time (10s default) are assumed to be from dead instances and ignored.

Cached read performance impact should be minimal as no data is being written to disk.

If a lock is not able to be acquired in 3 attempts, an exception is thrown which must be handled by clients.

# Files changed

* **package-lock.json/package.json**: Added proper-lockfile to handle locking of `data.json`
* **lowdbStorage.service.ts**: lockDbFile helper method ensures the file lock is held when accessing the db. If this option is enabled, it ensures only one instance can write at a time while preserving the current ability to run multiple instances.
  * locking is optional and defaults off
  * corrected spelling of info log message
  * added debug messages to indicate database state change.


cscharf (approved):
Nice! I think this looks good.

cscharf (changes requested):
Whoops, too soon. https://github.com/bitwarden/jslib/pull/273/checks?check_run_id=1890497729#step:8:15

MGibson1:
It turns out proper-lockfile is incompatible with non-node builds so it can't go in jslib.

I've moved the dependency to the directory-connector repository in a subclass.  There are still quite a few changes to lowedbStorage necessary to enable this.

cscharf (approved):
good catch/call.

PR #277, opened 2021-02-19
[Send] Updated delete function for add/edit component

vincentsalucci:
## Code Changes
- **send/add-edit.component**: `delete` now returns `Promise<boolean>` so that the child class can properly handle navigation changes after a successful deletion (if necessary).

## Breaking Changes
- `Browser` -> Yes, the child's `delete` signature/operations will need to be updated.
- `Desktop` -> No
- `Web` -> No

### DIFF
--- a/src/angular/components/send/add-edit.component.ts
+++ b/src/angular/components/send/add-edit.component.ts
@@ -239,6 +239,8 @@ export class AddEditComponent implements OnInit {
             this.onDeletedSend.emit(this.send);
             return true;
         } catch { }
+
+        return false;
     }
 
     typeChanged() {
--- a/src/angular/components/send/add-edit.component.ts
+++ b/src/angular/components/send/add-edit.component.ts
@@ -219,16 +219,16 @@ export class AddEditComponent implements OnInit {
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/angular/components/send/add-edit.component.ts
```diff
@@ -237,6 +237,7 @@ export class AddEditComponent implements OnInit {
             this.platformUtilsService.showToast('success', null, this.i18nService.t('deletedSend'));
             await this.load();
             this.onDeletedSend.emit(this.send);
+            return true;
         } catch { }
```
cscharf, line 241: You need at least one fallback `return false;` at the end of the method since you have a `try { ... } catch { }`

cscharf (changes requested):
Looks good, just 1 change needed, awesome documenting the jslib other repo impacts, thank you!

src/angular/components/send/add-edit.component.ts
vincentsalucci: That feeling when you don't save your final changes before pushing them upstream....



Thank you :D

cscharf (approved):

PR #278, opened 2021-02-20
feat: Help enable copying of TOTP code when auto filling credentials

tomershvueli:
Enable [Extension PR #1626](https://github.com/bitwarden/browser/pull/1626) by extending the jslib with proper constant and `TotpService` method.

### DIFF
--- a/src/services/constants.service.ts
+++ b/src/services/constants.service.ts
@@ -7,6 +7,7 @@ export class ConstantsService {
     static readonly disableFaviconKey: string = 'disableFavicon';
     static readonly disableAutoTotpCopyKey: string = 'disableAutoTotpCopy';
     static readonly enableAutoFillOnPageLoadKey: string = 'enableAutoFillOnPageLoad';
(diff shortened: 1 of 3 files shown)


src/services/constants.service.ts
```diff
@@ -7,6 +7,7 @@ export class ConstantsService {
     static readonly disableFaviconKey: string = 'disableFavicon';
     static readonly disableAutoTotpCopyKey: string = 'disableAutoTotpCopy';
     static readonly enableAutoFillOnPageLoadKey: string = 'enableAutoFillOnPageLoad';
+    static readonly enableAutoTotpCopyOnAutoFill: string = 'enableAutoTotpCopyOnAutoFill';
```
eliykat, line 10: Should be called `enableAutoTotpCopyOnAutoFillKey` to be consistent with the naming convention

tomershvueli: Fixed! 👍

eliykat (approved):

PR #280, opened 2021-02-22
Improve import error messages

eliykat:
## Objective

Required by https://github.com/bitwarden/web/pull/841.

## Code changes

* `import.service` - parse the validationErrors returned by the server, match them up to the corresponding import items, and format some nice error messages for the user.


cscharf (dismissed):

eliykat:
I wasn't handling non-model validation errors properly, e.g. if the server just throws an `Error("You cannot import this much data at once")`. This is now fixed.

cscharf (approved):

PR #281, opened 2021-02-22
Add disable badge counter

vrnvorona:
https://community.bitwarden.com/t/disable-badge-counter/18377

### DIFF
--- a/src/angular/components/send/add-edit.component.ts
+++ b/src/angular/components/send/add-edit.component.ts
@@ -51,7 +51,6 @@ export class AddEditComponent implements OnInit {
     expirationDateSelect: number = null;
     canAccessPremium = true;
     premiumRequiredAlertShown = false;
-    showOptions = false;
 
     private webVaultUrl: string;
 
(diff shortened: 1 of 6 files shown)


cscharf (changes requested):
Thanks for the submission!

Outside of the `constants.service` for the constants key, I would be inclined to say this is specific to the `browser` project only, and all implementation of loading/saving this setting should be encapsulated in there; favicons are relevant to all clients who load these icons for vault entries whereas the badge counter is only used in the browser extension.

These are located in browser/src/popup/accounts/*.component.ts

src/angular/components/send/add-edit.component.ts
```diff
@@ -51,7 +51,6 @@ export class AddEditComponent implements OnInit {
     expirationDateSelect: number = null;
```
eliykat, line 51: Can you remove these unrelated changes to this file?

vrnvorona: Done, accidentally reverted it earlier

cscharf (approved):

PR #285, opened 2021-02-26
Upload Directly to Azure or Local storage

MGibson1:
# Overview

Companion PR to Bitwarden/server#1167 and Bitwarden/web#853
Receives Send file upload and download URLs from Server to directly upload to a specified endpoint. This allows for directly uploading to Azure.

# Files Changed
* **package.json/package-lock.json**: Included azure/storage-blob to handle upload to an azure url
* **api.service.ts**: Separates Api tasks from posting a Send that is of type file to Posting a send without data, receiving Identifiers and upload URLs and a post endpoint to allow for direct upload to a local storage service (PostSendFile). Also creates a GetDownloadURL method that returns the appropriate URL to download a blob/file from.
* **FileUploadType**: Enum defining the way to upload a file. Direct upload to Bitwarden requires encapsulating in FormData, Blob upload is handled by Azure code.
* **SendFileApi/SendFileData/SendFile/SendFileView**: these models no longer contain a download URL since we can require calling the getSendFileDownloadData endpoint.
* **SendRequest**: Now includes a trust-based hint on the file size we will be uploading. Required and checked only for File type sends. This file size hint is verified post file upload and the send is deleted if it is incorrect within 1MB.
* **SendFileDownloadDataResponse**: Download information. Useful bit is the URL to fetch from
* **SendFileUploadDataResponse**: Upload information. Holds:
  1. The generated Send as a SendResponse
  2. The URL to upload a file to
  3. The type of upload required (see FileUploadType)
* **SendService**: Implement the above changes. Direct uploads use form data and upload only the file content. Azure code handles blob upload. SendRequest is popuplated with a file size.

### DIFF
--- a/src/services/send.service.ts
+++ b/src/services/send.service.ts
@@ -11,7 +11,7 @@ import { SendFile } from '../models/domain/sendFile';
 import { SendText } from '../models/domain/sendText';
 import { SymmetricCryptoKey } from '../models/domain/symmetricCryptoKey';
 
-import { FileUploadType } from '../enums/FileUploadType';
+import { FileUploadType } from '../enums/fileUploadType';
 import { SendType } from '../enums/sendType';
 
 import { SendView } from '../models/view/sendView';
--- a/src/services/send.service.ts
+++ b/src/services/send.service.ts
@@ -274,6 +274,6 @@ export class SendService implements SendServiceAbstraction {
     }
 
     private async azureUploadFileToServer(url: string, data: ArrayBuffer) {
-        const r = await new BlockBlobClient(url).uploadData(data);
+        await new BlockBlobClient(url).uploadData(data);
     }
 }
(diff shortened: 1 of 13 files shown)


src/services/send.service.ts
```diff
@@ -254,4 +254,26 @@ export class SendService implements SendServiceAbstraction {
...
+                } as any);
+            } else {
+                throw e;
+            }
+        }
+        await this.apiService.postSendFile(sendResponse.id, sendResponse.file.id, fd);
+    }
+
+    private async azureUploadFileToServer(url: string, data: ArrayBuffer) {
+        const r = await new BlockBlobClient(url).uploadData(data);
```
cscharf, line 277: Remove the unnecessary declaration and assignment of `const r`

cscharf (approved):
This looks good, although I would caution merging it before the server changes are ready and merged.

cscharf (approved):

MGibson1:
Closing due to split into #288 & #296

PR #288, opened 2021-03-01
Direct download for send

MGibson1:
# Overview 

Limits changes from #285 to download links only. Its companion is bitwarden/server#1174.

# Files changed
* **api.service.ts**: GetSendFileDownloadData. Request is re-sent to allow for reverification of access upon download of the file. This is a first step toward solving an access-count race and closes the issue of saving the request downlod link url and having unfettered downloads.
* **SendFileApi/SendFileData/SendFile/SendFileView**: Remove Url from models
* **SendFileDownloadDataResponse**: Response to download link reuquest. Currently holds SendResponse and url.

### DIFF
--- a/src/services/api.service.ts
+++ b/src/services/api.service.ts
@@ -420,8 +420,8 @@ export class ApiService implements ApiServiceAbstraction {
     }
 
 
-    async getSendFileDownloadData(send: SendAccessView): Promise<SendFileDownloadDataResponse> {
-        const r = await this.send('GET', '/sends/' + send.id + '/access/file/' + send.file.id, null, false, true);
+    async getSendFileDownloadData(send: SendAccessView, request: SendAccessRequest): Promise<SendFileDownloadDataResponse> {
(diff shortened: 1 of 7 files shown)


src/services/api.service.ts
```diff
@@ -416,6 +419,12 @@ export class ApiService implements ApiServiceAbstraction {
         return new SendAccessResponse(r);
     }
 
+
+    async getSendFileDownloadData(send: SendAccessView, request: SendAccessRequest): Promise<SendFileDownloadDataResponse> {
+        const r = await this.send('POST', '/sends/' + send.id + '/access/file/' + send.file.id, request, false, true);
```
cscharf, line 424: Isn't `send.id` supposed to be sent base64 encoded or is it already stored this way when concat'd here?

cscharf (commented):
Just 1 question

src/services/api.service.ts
cscharf: No worries, that's why I ask, it's good

cscharf (approved):

PR #289, opened 2021-03-01
Include File size in SendRequest

MGibson1:
# Overview

Client-side changes coupled to bitwarden/server#1175

# Files Changed
* **sendRequest.ts/send.service.ts**: include optional file length. This will be required for File type sends

### DIFF
--- a/src/services/send.service.ts
+++ b/src/services/send.service.ts
@@ -127,7 +127,7 @@ export class SendService implements SendServiceAbstraction {
     }
 
     async saveWithServer(sendData: [Send, ArrayBuffer]): Promise<any> {
-        const request = new SendRequest(sendData[0]);
+        const request = new SendRequest(sendData[0], sendData[1]?.byteLength);
         let response: SendResponse;
         if (sendData[0].id == null) {
             if (sendData[0].type === SendType.Text) {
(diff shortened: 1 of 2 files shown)


src/services/send.service.ts
```diff
@@ -127,7 +127,7 @@ export class SendService implements SendServiceAbstraction {
     }
 
     async saveWithServer(sendData: [Send, ArrayBuffer]): Promise<any> {
-        const request = new SendRequest(sendData[0]);
+        const request = new SendRequest(sendData[0], sendData[1]?.byteLength);
```
cscharf, line 130: If there's no `ArrayBuffer` provided, wouldn't this throw an index out of bounds exception vs. just `null`? Maybe not, not sure but figured I would bring it up as a question.

MGibson1: This data ultimately comes from the SendService encrypt method, which always provides an ArrayBuffer for all sends. It's just null unless a file was given to encrypt as well.

We're also protected by Typescript since you can't call the method without explicitly providing an array with a `[Send, ArrayBuffer]` type signature. This is the reason I'm OK with this design, otherwise I'd say it's too flimsy.

cscharf (approved):

PR #290, opened 2021-03-02
Date time fallbacks

addisonbeck:
Added date/time fallbacks on Send add/edit using separate props for browsers that don't support `<input type=datetime-local>`


cscharf (approved):
Sad day in browser compatibility land 🙁

PR #292, opened 2021-03-02
Date time fallback fixes

addisonbeck:
1. Null checked the array that is returned when trying to split a datetime. This was throwing a console error when creating sends.
2. Initialized expirationDate to null instead of undefined for reliability 
3. Corrected the conditional that applies the datetime fallback to account for expirationdate not being required conditionally.

### DIFF
--- a/src/angular/components/send/add-edit.component.ts
+++ b/src/angular/components/send/add-edit.component.ts
@@ -39,7 +39,7 @@ export class AddEditComponent implements OnInit {
     deletionDate: string;
     deletionDateFallback: string;
     deletionTimeFallback: string;
-    expirationDate: string;
+    expirationDate: string = null;
     expirationDateFallback: string;
     expirationTimeFallback: string;
     hasPassword: boolean;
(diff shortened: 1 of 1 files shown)


cscharf (approved):

src/angular/components/send/add-edit.component.ts
```diff
@@ -156,24 +156,28 @@ export class AddEditComponent implements OnInit {
...
     async submit(): Promise<boolean> {
-        if (!this.isDateTimeLocalSupported && this.expirationDateTimeFallback !== null) {
-            this.expirationDate = this.expirationDateTimeFallback;
-        }
-
-        if (!this.isDateTimeLocalSupported && this.deletionDateTimeFallback !== null) {
+        if (!this.isDateTimeLocalSupported) {
             this.deletionDate = this.deletionDateTimeFallback;
+            if ((this.editMode && this.expirationDateFallback != null) || this.expirationDateSelect === 0) {
+                this.expirationDate = this.expirationDateTimeFallback;
```
MGibson1, line 179: `this.expirationDateTimeFallback` return `'T'` when you clear out the expiration parts.

Probably the parts need to be linked so that clearing one clears the other. Also there is a clear link which does not work right now.

I'm seeing some stuff to get rid of the clear `x` entirely for date and time inputs, that might be the way to go

MGibson1 (changes requested):

cscharf:
Good catches @MGibson1 , sorry I missed that stuff Addison

PR #293, opened 2021-03-03
fixed up the send date fallbacks again

addisonbeck:
**Addresses:** 
> for a Send with an expiration, you can go in there and clear that out, but you can’t save because expirationDateFallback will never be null again. Should probably also check expirationTimeFallback since they’re both required.
> last issue is the clear <a> tag doesn’t work and it’d be nice if we could hide the clear out buttons in the input or link the two fields

1. Check if either expiration date or time is null in `expirationDateTimeFallback()` and return null if true are instead of "T"
2. Throw an error if only one of the two is null informing the user to add both.
3. Added expirationDate and Time fallbacks to `clearExpiration()`

### DIFF
--- a/src/angular/components/send/add-edit.component.ts
+++ b/src/angular/components/send/add-edit.component.ts
@@ -114,7 +114,9 @@ export class AddEditComponent implements OnInit {
     }
 
     get expirationDateTimeFallback() {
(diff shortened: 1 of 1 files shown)


MGibson1 (approved):
Looks good. Nice job, this was finicky!

PR #296, opened 2021-03-05
Upload to Azure strorage blobs

MGibson1:
# Overview

Utilizes Server endpoints to grab SAS-enabled upload URLs and either upload directly to Server or to Azure, depending on the upload type.

Azure uploads are handled by a new Service, which PUTs single blobs where possible; Blocks where necessary, also alters the block size depending on Azure version. 

Unfortunately, Azure Block PUTs require Write access, so we need to allow for a way to update SAS Tokens when they're about to expire. That's handled by a renew callback.

The plan on the Server side is to hand out renewed upload links until an Azure blob create event is called (triggered by block list PUT). We'll then mark a flag on the Send to specify it's been written and deny further upload URL requests.

# Files Changed
* **api.Service.ts**: Endpoints for
  * Creating a file type send: returns UploadData object. which specifies the method and url to upload a file to.
  * Posting a file to local storage
  * Renewing a download URL that has or is about to expired
* **add-edit.compontent**: remove 100MB limit
* **FileUploadType**: specifies the upload scheme to use while uploading file data.
* **Utils**: useful methods while uploading files. We need to be able to always retrieve a URL object, and encode UTF string to base64 for blockIds.
* **FileUploadDataResponse**: Response to creation of a file type send. Constains the send response, a url to upload data to, and the type of file upload.
* **azureStorage.service.ts**: handles uploading to Azure blob storage from a SAS-tokenized URL.
* **send.service.ts**: Use upload data response and different upload types.

# Test Requirements

Upload and download to Local Storage and Azure of various sizes.

Critical sizes are: 
1. <256MiB
2. \>256MiB
3. upload >256MiB taking longer than 1 min to upload

### DIFF
--- a/src/abstractions/fileUpload.service.ts
+++ b/src/abstractions/fileUpload.service.ts
@@ -1,0 +1,7 @@
+import { CipherString } from '../models/domain';
+import { SendFileUploadDataResponse } from '../models/response/sendFileUploadDataResponse';
+
+export abstract class FileUploadService {
+    uploadSendFile: (uploadData: SendFileUploadDataResponse, fileName: CipherString,
+        encryptedFileData: ArrayBuffer) => Promise<any>;
+}
--- a/src/services/send.service.ts
+++ b/src/services/send.service.ts
@@ -150,7 +150,7 @@ export class SendService implements SendServiceAbstraction {
                                 const renewalResponse = await this.apiService.renewFileUploadUrl(response.id, response.file.id);
                                 return renewalResponse.url;
                             };
-                            await this.azureStorageService.azureUploadFileToServer(uploadDataResponse.url, sendData[1],
+                            await this.azureStorageService.uploadFileToServer(uploadDataResponse.url, sendData[1],
                                 renewalCallback);
                             break;
                         default:
--- a/src/services/send.service.ts
+++ b/src/services/send.service.ts
@@ -136,10 +136,14 @@ export class SendService implements SendServiceAbstraction {
             if (sendData[0].type === SendType.Text) {
                 response = await this.apiService.postSend(request);
             } else {
-                const uploadDataResponse = await this.apiService.postFileTypeSend(request);
-                response = uploadDataResponse.sendResponse;
+                try {
+                    const uploadDataResponse = await this.apiService.postFileTypeSend(request);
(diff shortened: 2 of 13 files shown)


cscharf (approved):
Looks good!

MGibson1:
@cscharf. I implemented the client-side changes discussed yesterday:
1. Fall back to original upload method for old server instances
2. Abstract file upload to a service. I'll be adding cipher attachment uploads to this service as well.

src/services/send.service.ts
```diff
@@ -133,23 +136,14 @@ export class SendService implements SendServiceAbstraction {
...
-                    if (Utils.isNode && !Utils.isBrowser) {
-                        fd.append('model', JSON.stringify(request));
-                        fd.append('data', Buffer.from(sendData[1]) as any, {
-                            filepath: sendData[0].file.fileName.encryptedString,
-                            contentType: 'application/octet-stream',
-                        } as any);
-                    } else {
-                        throw e;
-                    }
+                    response = await this.legacyServerSendFileUpload(sendData, request);
```
cscharf, line 145: Do we want to try this based on the error response HTTP status code (i.e. "404") vs. just a blanket fallback for any error (what if it's a validation error, for example)?

src/abstractions/fileUpload.service.ts
```diff
@@ -0,0 +1,7 @@
+import { CipherString } from '../models/domain';
+import { SendFileUploadDataResponse } from '../models/response/sendFileUploadDataResponse';
+
+export abstract class FileUploadService {
```
cscharf, line 4: This abstraction is missing from `abstractions/index.ts`.

don't forget the `services.module.ts` in downstream repos as well (although not relevant here)

MGibson1: There is a comment in index.ts informing its deprecation -- I believe we're just importing direct from files now rather than a one-stop-shop

https://github.com/bitwarden/jslib/blob/master/src/abstractions/index.ts#L1

cscharf: 🤦🏻 , you're right, good catch, thanks.

cscharf (approved):

cscharf (approved):
Awesome, thanks Matt, still reviewing server PR

PR #307, opened 2021-03-15
Enhance SafeInCloud import

PaulVII:
Hello everyone, 
I recently switched from SafeInCloud and found the import tool to be not very good. I made some adjustments for myself and though I'd give them back to the community and this great otherwise great program.
I understand that I had to make some choices here which may not be liked by everyone and might not be the right thing to use in Bitwarden, but at least importing favourite status seems universally useful to me :)
I'd be glad to get feedback from you

Paul

**Changes**
- Keep favourite status when importing
- in SafeInCloud, the trash doesn't ever get cleaned automatically.
  -  Currently, all deleted cards are imported, which can be a lot of very old ones
  - I exempt deleted cards from being imported (I wanted to set cipher.deleteDate, but this didn't work)
- SafeInCloud allows for multiple type="password" fields. 
  - Currently, the importer doesn't account for this and therefore just didn't import all but the last password field
  -  Therefore, I add all type="password" fields as hidden FieldViews and try to find the best candidate to be added as the autofill login value
- SafeInCloud has type="secret" fields which are treated like passwords with hidden values
  - Currently, these just stored as plaintext values
  - I treat them the same as type="password" (and therefore add them as a hidden FieldView)

### DIFF
--- a/src/importers/safeInCloudXmlImporter.ts
+++ b/src/importers/safeInCloudXmlImporter.ts
@@ -9,6 +9,9 @@ import { SecureNoteView } from '../models/view/secureNoteView';
 import { CipherType } from '../enums/cipherType';
 import { SecureNoteType } from '../enums/secureNoteType';
 
+import { FieldType } from '../enums';
+import { CipherView, FieldView } from '../models/view';
+
 export class SafeInCloudXmlImporter extends BaseImporter implements Importer {
     parse(data: string): Promise<ImportResult> {
         const result = new ImportResult();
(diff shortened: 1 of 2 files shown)


src/importers/safeInCloudXmlImporter.ts
```diff
@@ -39,7 +42,7 @@ export class SafeInCloudXmlImporter extends BaseImporter implements Importer {
         });
 
         Array.from(doc.querySelectorAll('database > card')).forEach(cardEl => {
-            if (cardEl.getAttribute('template') === 'true') {
+            if (cardEl.getAttribute('template') === 'true' || cardEl.getAttribute('deleted') === 'true') {
```
MGibson1, line 45: Just to be sure, is there an easy way in SafeInCloud to un-delete items so if someone notices something missing they can recover? Sorry to use you as a go-between here ;)

MGibson1 (approved):
This is really nice work, thank you!

With regard to deleted items, we are behaving the same way for our 1pw pif importer right now, so I think the right call is to ignore them.

Above and beyond ask:
Is there any way I can get you to add in a mocked out csv line or two to `spec/common/importers/testData/safeInCloud`? It's a pet project of mine to get better coverage in general and some test data makes matters a lot easier on the importer fronts.

src/importers/safeInCloudXmlImporter.ts
PaulVII: > Just to be sure, is there an easy way in SafeInCloud to un-delete items so if someone notices something missing they can recover? Sorry to use you as a go-between here ;)

Yes, you can just right-click and press "restore". It's possible to restore multiple deleted entries or all via CTRL+A

PaulVII:
> This is really nice work, thank you!
> 
> With regard to deleted items, we are behaving the same way for our 1pw pif importer right now, so I think the right call is to ignore them.
> 
> Above and beyond ask:
> Is there any way I can get you to add in a mocked out csv line or two to `spec/common/importers/testData/safeInCloud`? It's a pet project of mine to get better coverage in general and some test data makes matters a lot easier on the importer fronts.

Thanks! I created and exported a small database that should cover most of what the importer does. Is this what you intended?

MGibson1 (approved):
Yup, the idea is it'll be easier when we find time to truck through these importers' test cases to have some sample data. Thank you!

PR #309, opened 2021-03-17
Fixes linting error introduced with PR #307

djsmith85:
Just a semicolon missing at the end of spec/common/importers/testData/safeInCloud/testData.xml.ts


cscharf (approved):
Thanks!

PR #310, opened 2021-03-23
Hide email address in Sends

eliykat:
Required for PR: https://github.com/bitwarden/web/pull/895

### DIFF
--- a/src/angular/components/send/add-edit.component.ts
+++ b/src/angular/components/send/add-edit.component.ts
@@ -19,10 +19,11 @@ import { PolicyService } from '../../../abstractions/policy.service';
 import { SendService } from '../../../abstractions/send.service';
 import { UserService } from '../../../abstractions/user.service';
 
-import { Send } from '../../../models/domain/send';
 import { SendFileView } from '../../../models/view/sendFileView';
 import { SendTextView } from '../../../models/view/sendTextView';
 import { SendView } from '../../../models/view/sendView';
+
+import { Send } from '../../../models/domain/send';
 
 // TimeOption is used for the dropdown implementation of custom times
 // Standard = displayed time; Military = stored time
@@ -47,7 +48,6 @@ export class AddEditComponent implements OnInit {
     copyLink = false;
     disableSend = false;
     disableHideEmail = false;
-    disableThisSend = false;
     send: SendView;
     deletionDate: string;
     deletionDateFallback: string;
@@ -69,7 +69,6 @@ export class AddEditComponent implements OnInit {
     canAccessPremium = true;
     premiumRequiredAlertShown = false;
     showOptions = false;
-
 
     safariDeletionTime: string;
     safariExpirationTime: string;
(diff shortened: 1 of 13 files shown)


src/angular/components/send/add-edit.component.ts
```diff
@@ -151,13 +154,15 @@ export class AddEditComponent implements OnInit {
...
-            return o.enabled &&
-                o.status === OrganizationUserStatusType.Confirmed &&
-                o.usePolicies &&
-                !o.canManagePolicies &&
-                policies.some(p => p.organizationId === o.id && p.enabled);
-        });
+        const disableSendPolicies = policies.filter(p =>
+            p.enabled && organizations.some(o =>
+                o.enabled &&
+                    o.status === OrganizationUserStatusType.Confirmed &&
```
cscharf, line 160: reduce the indentation of lines 160~163 to align with line 159.

src/angular/components/send/add-edit.component.ts
```diff
@@ -178,6 +183,12 @@ export class AddEditComponent implements OnInit {
             }
         }
 
+        this.disableAnonymousSend = disableSendPolicies.some(p =>
+            p.data?.disableSend === DisableSendType.DisableAnonymous);
+        this.disableSend = (this.disableAnonymousSend && this.editMode && this.send.hideEmail) ||
```
cscharf, line 188: This logic doesn't make sense, either Send is disabled by policy or not, shouldn't make a difference whether anonymous is selected in this scenario.

cscharf (commented):
Will come back to this after re-routing to the server change with updates there.

src/angular/components/send/add-edit.component.ts
```diff
@@ -183,13 +189,9 @@ export class AddEditComponent implements OnInit {
...
         }
 
-        this.disableAnonymousSend = disableSendPolicies.some(p =>
-            p.data?.disableSend === DisableSendType.DisableAnonymous);
-        this.disableSend = (this.disableAnonymousSend && this.editMode && this.send.hideEmail) ||
-            disableSendPolicies.some(p =>
-                (p.data?.disableSend ?? DisableSendType.DisableAll) === DisableSendType.DisableAll);
-
         this.hasPassword = this.send.password != null && this.send.password.trim() !== '';
+        this.disableThisSend = this.disableThisSend ||
```
cscharf: What is the purpose of this? We should be allowing a user, assuming a policy disabling anonymous email was created or added to their org after the Send's created, to edit the Send after the policy comes into effect vs. disabling everything.

The user should be able to edit the Send, disable the checkbox, but ensure it is unchecked from this perspective. So if they save it, it will save w/o the value checked.

eliykat: I think you're looking at an old commit - I changed this so that it works how you describe, unless something weird is going on with my commits

cscharf: 🤦🏻

cscharf (approved):

PR #324, opened 2021-04-05
Use iterator compatible Array instantiation

MGibson1:
# Overview

It appears that there are some issues with instantiating an Array by expanding a Set object through `[...set]`. This hopefully fixes this issue in other clients (`web` observed so far) requiting `downlevelIteration` option set in `tsconfig.json` in order to compile.

`downlevelIteration` comes with a warning that it will impact performance and increase verbosity for all javascript produces for any iteration. This doesn't seem worth it for this one line. Hopefully this instantiation works and we can get past this...

# Files changed:
*  **Cipher.ts**: instantiate the collectionId array in a different way


cscharf (approved):
bummer... `[...set]` was way sexier.

PR #330, opened 2021-04-06
Optimize isSearchable for Chinese.

iineva:
In Chinese, one character is also  meaning.
Such as A char `长` to `long`, `短` to `short`.
So we need to input one character to trigger searching.

### DIFF
--- a/src/services/search.service.ts
+++ b/src/services/search.service.ts
@@ -3,10 +3,10 @@ import * as lunr from 'lunr';
 import { CipherView } from '../models/view/cipherView';
 
 import { CipherService } from '../abstractions/cipher.service';
+import { I18nService } from '../abstractions/i18n.service';
 import { LogService } from '../abstractions/log.service';
 import { SearchService as SearchServiceAbstraction } from '../abstractions/search.service';
 
-import { I18nService } from '../abstractions';
 import { CipherType } from '../enums/cipherType';
 import { FieldType } from '../enums/fieldType';
 import { UriMatchType } from '../enums/uriMatchType';
--- a/src/services/search.service.ts
+++ b/src/services/search.service.ts
@@ -10,12 +10,17 @@ import { CipherType } from '../enums/cipherType';
 import { FieldType } from '../enums/fieldType';
 import { UriMatchType } from '../enums/uriMatchType';
 import { SendView } from '../models/view/sendView';
+import { I18nService } from '../abstractions';
 
 export class SearchService implements SearchServiceAbstraction {
     private indexing = false;
     private index: lunr.Index = null;
+    private searchMinEdge = 2;
 
-    constructor(private cipherService: CipherService, private logService: LogService) {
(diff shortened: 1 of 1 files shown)


src/services/search.service.ts
```diff
@@ -10,21 +10,26 @@ import { CipherType } from '../enums/cipherType';
 import { FieldType } from '../enums/fieldType';
 import { UriMatchType } from '../enums/uriMatchType';
 import { SendView } from '../models/view/sendView';
+import { I18nService } from '../abstractions';
 
 export class SearchService implements SearchServiceAbstraction {
     private indexing = false;
     private index: lunr.Index = null;
+    private searchMinEdge = 2;
```
cscharf, line 18: Please consider variable name of, `searchableMinLength` instead of "searchMinEdge".

src/services/search.service.ts
```diff
@@ -10,21 +10,26 @@ import { CipherType } from '../enums/cipherType';
 import { FieldType } from '../enums/fieldType';
 import { UriMatchType } from '../enums/uriMatchType';
 import { SendView } from '../models/view/sendView';
+import { I18nService } from '../abstractions';
```
cscharf, line 13: Imports are grouped and should be in alphabetical order within each group. Please move this import to the other `abstractions` group, in the proper order.

src/services/search.service.ts
```diff
@@ -10,21 +10,26 @@ import { CipherType } from '../enums/cipherType';
...
 
 export class SearchService implements SearchServiceAbstraction {
     private indexing = false;
     private index: lunr.Index = null;
+    private searchMinEdge = 2;
 
-    constructor(private cipherService: CipherService, private logService: LogService) {
+    constructor(private cipherService: CipherService, private logService: LogService, private i18nService: I18nService) {
```
cscharf, line 20: Please wrap `private i18nService: I18nService` onto the next line, constructor arguments should be 2 per line when exceeding 2 arguments.

cscharf (changes requested):
Thank you for your submission. This should be okay, but just a few formatting and style issues. Also please ensure you run the linter before final submission.

iineva:
@cscharf All done.

src/services/search.service.ts
```diff
@@ -6,20 +6,21 @@ import { CipherService } from '../abstractions/cipher.service';
 import { LogService } from '../abstractions/log.service';
 import { SearchService as SearchServiceAbstraction } from '../abstractions/search.service';
 
+import { I18nService } from '../abstractions';
```
cscharf, line 9: Please move this up to the other `abstractions` block above this block, in alphabetical order (so between lines 5 and 6 above). 

Also, missed this last time, you should import the actual service, we don't use general directory imports, so `from '../abstractions/i18n.service';` instead of just `from '../abstractions';`.

cscharf (changes requested):

iineva:
My bad, It's ok now.

cscharf (approved):
Looks good, thanks! Will run this by the team and queue it up for QA.

PR #331, opened 2021-04-07
Require user to verify email to use file Send

eliykat:
jslib logic for https://github.com/bitwarden/web/pull/915

### DIFF
--- a/src/angular/components/send/add-edit.component.ts
+++ b/src/angular/components/send/add-edit.component.ts
@@ -69,8 +69,7 @@ export class AddEditComponent implements OnInit {
     expirationDateSelect: number = null;
     canAccessPremium = true;
     emailVerified = true;
-    premiumRequiredAlertShown = false;
-    emailVerificationRequiredAlertShown = false;
+    alertShown = false;
     showOptions = false;
 
     safariDeletionTime: string;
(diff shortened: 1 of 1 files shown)


src/angular/components/send/add-edit.component.ts
```diff
@@ -355,9 +360,16 @@ export class AddEditComponent implements OnInit {
...
-        if (!this.canAccessPremium && this.send.type === SendType.File && !this.premiumRequiredAlertShown) {
-            this.premiumRequiredAlertShown = true;
-            this.messagingService.send('premiumRequired');
+        if (this.send.type === SendType.File)
+        {
+            if (!this.canAccessPremium && !this.premiumRequiredAlertShown) {
+                this.premiumRequiredAlertShown = true;
+                this.messagingService.send('premiumRequired');
+            } else if (!this.emailVerified && !this.emailVerificationRequiredAlertShown &&
+                !this.premiumRequiredAlertShown) {
```
Hinton, line 369: I assume the `!this.premiumRequiredAlertShow` is to prevent multiple dialogs from appearing every time the user toggles it? If so I would suggest we switch to having a single property `alertShown` which both checks can use?

eliykat: Good point!

Hinton (approved):

PR #333, opened 2021-04-09
Use web workers for bulk cipher decryption

eliykat:
UPDATED 12 JULY - this is now on the roadmap board, so I've updated all the `web-workers` branches with the latest `master`. Ready for review.

## Objective

Partial solution to #1597: 
> If your account has access to more than 2000 items, the login and search process with the browser extension gets strongly degraded.

In particular, when logging into the web vault with a large number of items, the browser tab can become unresponsive and the browser will offer to kill it. This can last for up to 30 seconds depending on the number of vault items.

This PR uses a web worker to decrypt all personal vault contents on login. It works on all Angular clients (see links to other PRs below).

With these changes, the web vault UI is fully usable while its contents are being decrypted, and the browser does not try to kill the tab. I was testing with ~8k items and noticed no slowdown, except for a brief pause during indexing (which I do not propose to deal with at this time). Decryption is also significantly faster - see some rough benchmarking in my comment below.

This is now ready for review! I'm sure there'll be lots to fix but I think I have the basic structure hammered out.

## Code changes

Related PRs:
* web: https://github.com/bitwarden/web/pull/920
* browser: https://github.com/bitwarden/browser/pull/1803
* desktop: https://github.com/bitwarden/desktop/pull/858

Note: it looks like I've changed a lot in this PR (jslib), but many of the changes are just to support de/serialization. You can probably go through and tick off these changes to the view models first before diving into the substantive changes.

### Background

General info about web workers is available [here](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API), but here are the basics.

* web workers execute code on a separate thread, which avoids locking up the UI when performing CPU-intensive tasks
* web workers can only communicate with the main program/thread via messaging
* messages to/from web workers contain a shallow copy of objects. Object references to instantiated classes will not survive the trip. All objects need to be serialized at one end, and then deserialized and instantiated at the other end
* web workers execute in their own context outside `ngZone` and do not have access to Angular services (which are on the main thread anyway)

### Implementation

The basic flow is this:

1. `cipherService` checks whether the number of ciphers to be decrypted exceeds a certain threshold - currently 250. If so, it will use a web worker for decryption. If not, it will decrypt on the main thread (current behaviour).
2. If using a web worker, it will call `webWorkerService` to create the worker and get a reference to it.
3. `cipherService` then serializes and sends the data required to the worker. This consists of the `cipherData`, plus the encryption keys etc. required.
2. The web worker receives and deserializes the `cipherData`. It also sets up its environment, starts its own services, etc.
3. The web worker decrypts the `cipher` objects, then serializes them and returns them to `cipherService`.
5. `cipherService` deserializes the objects and then resumes the normal flow.

Note that this maintains the same `cipherService` interface, so these changes are abstracted away from all other Angular components and services.

### Notes

* the web worker cleans up its own state before terminating to try to get rid of sensitive data. I followed what we do currently, which just seems to be nulling out object references given that we don't have any direct memory access. It will also do this on lock and logout.
* `workerLogService` is fairly redundant, but I couldn't get `hrtime()` working in the web worker context and didn't want to overcomplicate things with subclassing `consoleLogService`.
* Electron has [some warnings about multithreading](https://www.electronjs.org/docs/tutorial/multithreading), but they relate to loading native Node modules. I don't *think* we use native node modules in this part of the app but would appreciate someone running their eyes over this.
* Webpack 5 has native web worker bundling/loading, but we're still on Webpack 4, so for now we're using `worker-loader` for this work. It's a plugin that gets added to the webpack config.
* the web worker has its own webpack bundle, because it runs in its own context. Therefore, we give `worker-loader` a separate `AngularCompilerPlugin` instance, which uses a dedicated `tsconfig` just for workers. (So that they don't bundle the entire app again unnecessarily.) Unfortunately, this does not seem to play nicely with `tslint-loader`, which (best I can tell) reads the transpiled javascript file and throws a bunch of errors. I've excluded `.worker.ts` files from `tslint-loader`, which is not a great solution, but they should be checked by `npm run lint` anyway.

## Testing

I've done basic testing on the following platforms:

* web - Chrome, FF, Safari
* browser - Chrome, FF
* desktop - Electron (MacOS)


src/workers/services/workerCryptoService.ts
```diff
@@ -0,0 +1,57 @@
+import { EncryptionType } from '../../enums';
+import { CipherString, SymmetricCryptoKey } from '../../models/domain';
+import { NodeCryptoFunctionService } from '../../services/nodeCryptoFunction.service';
+import { WorkerLogService } from './workerLogService';
+
+export class WorkerCryptoService {
+    cryptoFunctionService = new NodeCryptoFunctionService();
```
kspearrin, line 7: So we can run node APIs in a web worker?

src/workers/services/workerCryptoService.ts
```diff
@@ -0,0 +1,57 @@
...
+    worker: Worker;
+    logService: WorkerLogService;
+    key: any;
```
kspearrin, line 10: I worry about cleaning up state in this web worker. We would need to make sure locked app states are not holding onto sensitive data.

kspearrin:
> (Again, we can't use the real cipherService because it's shot through with Angular services and dependencies that the web worker can't resolve.) 

cipherService (or any service) should have nothing to do with angular. We use it (and other services) in our CLI which is not an angular app.

src/workers/services/workerCryptoService.ts
eliykat: Yes

src/workers/services/workerCryptoService.ts
eliykat: Noted, I'll look into how to handle this

eliykat:
I was thinking that services required Angular's DI framework to resolve their dependencies (e.g. CryptoService requires StorageService etc), but I see we can just populate those by hand. Makes sense.

With this in mind, I've refactored the worker to use the standard CryptoService. This is much nicer and immediately made everything a lot easier. It also meant that I was able to expand it to organizations really easily, so it's working for orgs now as well.

Do you have any view as to whether we use workers for all bulk decryption operations, or only over a certain threshold? Spinning up a worker may be excessive for a small vault.

eliykat:
I skimmed a couple of articles on web worker cost/latency, they suggest that despite warnings of web workers being "expensive", we're still talking in the tens-of-milliseconds range for spinning up and passing messages. (e.g. [here](https://dassur.ma/things/is-postmessage-slow/), [here](https://dassur.ma/things/is-postmessage-slow/)). 

I figured benchmarking would be more useful, so I ran some very rough tests using data generated by Mockaroo. I didn't average out the results, but I did try each test a couple of times to make sure the result wasn't an obvious outlier. Using Chrome.

number of items | web worker time (ms) | main thread time (ms)
--- | --- | ---
1200 | 2066 | 5015
1000 | 2058 | 4002
800 | 1061 | 3076
600 | 1075 | 2019
400 | 1002 | 1055
200 | 86 | 79

Thoughts:
* not sure what happened between 200 & 400, maybe quirks in the dataset
* A worker is much faster as the vault grows larger
* Not much cost to using workers for smaller vaults, compared to using the main thread
* I think we still specify a threshold where the worker kicks in, but we can be fairly conservative on it, e.g. 250 items.

kspearrin:
> I think we still specify a threshold where the worker kicks in, but we can be fairly conservative on it, e.g. 250 items.

Agreed.

src/services/cipher.service.ts
```diff
@@ -68,13 +70,16 @@ const DomainMatchBlacklist = new Map<string, Set<string>>([
 export class CipherService implements CipherServiceAbstraction {
     // tslint:disable-next-line
     _decryptedCipherCache: CipherView[];
+    workerThreshold: number = 250;
```
kspearrin, line 73: should this be private or a const declared outside the class?

src/electron/window.main.ts
```diff
@@ -118,6 +118,7 @@ export class WindowMain {
             alwaysOnTop: this.enableAlwaysOnTop,
             webPreferences: {
                 nodeIntegration: true,
+                nodeIntegrationInWorker: true,
```
kspearrin, line 121: is this still necessary? looks like you stopped using node in the worker.

src/services/cipher.service.ts
eliykat: It's only referenced in this class, so I think it can just be private.

src/electron/window.main.ts
eliykat: Good point - I've now removed this and cleaned up some node dependencies that were accidentally being used in the worker. I think the new boolean logic in `utils.ts` (`isWebWorker` etc) will need to be looked at closely. But this is great because we avoid Node threadsafety issues, afaik.

src/misc/utils.ts
```diff
@@ -24,17 +25,19 @@ export class Utils {
...
+        Utils.global = Utils.isNativeScript ?
+            global : (Utils.isNode && !Utils.isBrowser && !Utils.isWebWorker ? global : self);
     }
 
     static fromB64ToArray(str: string): Uint8Array {
         if (Utils.isNode || Utils.isNativeScript) {
             return new Uint8Array(Buffer.from(str, 'base64'));
         } else {
-            const binaryString = window.atob(str);
+            const binaryString = self.atob(str);
```
kspearrin, line 40: What is going on here with self? Is there a local atob function somewhere?

src/workers/crypto.worker.ts
```diff
@@ -77,7 +79,7 @@ class CryptoWorker {
             this.platformUtilsService, this.logService);
 
         this.containerService = new ContainerService(this.cryptoService);
-        this.containerService.attachToGlobal(global);
+        this.containerService.attachToGlobal(self);
```
kspearrin, line 82: again, im a little confused by this change

src/misc/utils.ts
eliykat: `self` refers to the global object. It's useful because workers have their own global object instead of `window` - the MDN docs explain it well [here](https://developer.mozilla.org/en-US/docs/Web/API/Window/self):

>The `Window.self` read-only property returns the window itself, as a WindowProxy. It can be used with dot notation on a window object (that is, `window.self`) or standalone (`self`). The advantage of the standalone notation is that a similar notation exists for non-window contexts, such as in Web Workers. By using `self`, you can refer to the global scope in a way that will work not only in a window context (`self` will resolve to `window.self`) but also in a worker context (`self` will then resolve to `WorkerGlobalScope.self`).

So this is just another way of accessing the `atob`/`btoa` functions, which is compatible on the main thread (via `window`) and in workers (via `WorkerGlobalScope`).

src/workers/crypto.worker.ts
eliykat: See previous comment

common/src/services/cipher.service.ts
```diff
@@ -303,22 +309,95 @@ export class CipherService implements CipherServiceAbstraction {
...
         });
 
         await Promise.all(promises);
-        decCiphers.sort(this.getLocaleSortingFunction());
-        this.decryptedCipherCache = decCiphers;
-        return this.decryptedCipherCache;
+        return decryptedCiphers;
+    }
+
+    async decryptManyCiphersWithWorker(workerName: string, cipherData: CipherData[]): Promise<CipherView[]> {
```
kspearrin, line 356: should this be private?

common/src/services/cipher.service.ts
```diff
@@ -68,12 +71,15 @@ export class CipherService implements CipherServiceAbstraction {
     // tslint:disable-next-line
     _decryptedCipherCache: CipherView[];
 
+    private workerThreshold: number = 250;
```
kspearrin, line 74: Seems like this should be a const outside the class like Keys, DomainMatchBlacklist, etc

common/src/services/cipher.service.ts
```diff
@@ -303,22 +309,95 @@ export class CipherService implements CipherServiceAbstraction {
...
 
-        const promises: any[] = [];
         const ciphers = await this.getAll();
+
+        this.consoleLogService.time('getAllDecrypted');
+        let decryptedCiphers;
+        if (this.platformUtilsService.supportsWebWorkers() && ciphers.length > this.workerThreshold) {
+            this.consoleLogService.info('getAllDecrypted started using worker');
+            const userId = await this.userService.getUserId();
+            const cipherData = ciphers.map(c => c.toCipherData(userId));
```
kspearrin, line 324: Why do we have to turn it back to cipherData? because it needs to be serializable or something?

common/src/services/cipher.service.ts
```diff
@@ -437,18 +516,24 @@ export class CipherService implements CipherServiceAbstraction {
...
-            return decCiphers;
+        const ciphersResponse = await this.apiService.getCiphersOrganization(organizationId);
+        if (ciphersResponse != null && ciphersResponse.data != null && ciphersResponse.data.length) {
+            const consoleLogKey = 'getAllFromApiForOrganization (' + organizationId + ')';
+            this.consoleLogService.time(consoleLogKey);
+            let decryptedCiphers;
+            if (this.platformUtilsService.supportsWebWorkers() && ciphersResponse.data.length > this.workerThreshold) {
+                this.consoleLogService.info('getAllFromApiForOrganization started using worker');
+                const cipherData: CipherData[] = ciphersResponse.data.map(r => new CipherData(r));
+                decryptedCiphers = await this.decryptManyCiphersWithWorker(organizationId, cipherData);
```
kspearrin, line 527: workerName is just org id?

common/src/services/cipher.service.ts
```diff
@@ -303,22 +309,95 @@ export class CipherService implements CipherServiceAbstraction {
...
+                platformUtilsData: JSON.stringify(platformUtilsData),
+            });
+            worker.addEventListener('message', event => {
+                if (event.data.type !== 'decryptManyResponse') {
+                    return;
+                }
+
+                const decryptedCiphers: CipherView[] = event.data.ciphers.map((v: any) => {
+                    const cipherView = new CipherView();
+                    cipherView.buildFromObj(JSON.parse(v));
```
kspearrin, line 393: Do we have to `JSON.parse` each object individually? We can't parse once from one array or objects and iterate that?

common/src/workers/crypto.worker.ts
```diff
@@ -0,0 +1,116 @@
...
+        workerApi.postMessage(message);
+    }
+
+    async clearCache() {
+        await Promise.all([
+            this.cryptoService.clearKey(),
+            this.cryptoService.clearOrgKeys(false),
+            this.cryptoService.clearKeyPair(false),
+            this.cryptoService.clearEncKey(false),
+        ]);
```
kspearrin, line 101: Can we clear the whole memory storage when the worker is terminated too? Maybe that already happens somehow?

kspearrin (changes requested):

eliykat:
This is too out of date, I'm working on a separate branch now and will submit a new PR when ready.

PR #341, opened 2021-04-10
feat: add hidden char count toggle

mkanavakatini:
- Added password count toggles to password input & hidden value custom input
- Added `color-password-count` pipe to render password with count & character type style

[Browser PR is here](https://github.com/bitwarden/browser/pull/1780)


src/enums/eventType.ts
```diff
@@ -25,6 +25,7 @@ export enum EventType {
     Cipher_ClientAutofilled = 1114,
     Cipher_SoftDeleted = 1115,
     Cipher_Restored = 1116,
+    Cipher_ClientToggledHiddenFieldCount = 1117,
```
cscharf, line 28: Hi, please update/rebase latest and note that there's a new enum taking this value, you'll need to go up to the next. Also, this event type is an "audit" style event for logging and not something we would likely track (such as the same as showing a hidden field value) and if it automatically shows the hidden field value, it should simply re-use the existing event for that. Plus this would need to also exist via the API/server; so best to revert this change here.

src/models/view/fieldView.ts
```diff
@@ -22,4 +22,8 @@ export class FieldView implements View {
     get maskedValue(): string {
         return this.value != null ? '••••••••' : null;
     }
+
+    get arrayValue(): string[] {
+        return this.value != null ?  Array.from(this.value) : [];
```
cscharf, line 27: Formatting: remove double space after `?  Array.from` to just `? Array.from`

src/angular/components/view.component.ts
```diff
@@ -201,6 +201,14 @@ export class ViewComponent implements OnDestroy, OnInit {
         }
     }
 
+    toggleFieldCount(field: FieldView) {
+        const f = (field as any);
+        f.showCount = !f.showCount;
+        if (f.showCount) {
```
cscharf, line 207: This `if` block can be removed in that we don't need to collect this particular event, unless showing the count takes the state from hidden to visible, otherwise this shouldn't be necessary (and if it is necessary, please re-use the same event for toggling a hidden field view action).

cscharf (commented):
Hi @mkanavakatini , thanks for the contribution, much appreciated! I left some feedback, please let me know if you have any questions.

angular/src/pipes/color-password-count.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+    transform(password: string) {
+        const passwordArray = Array.from(password);
+        let colorizedPasswordCount = '';
+        for (let i = 0; i < passwordArray.length; i++) {
+            const [character, type] = this.getCharacterAndType(
+                passwordArray[i]
+            );
+            colorizedPasswordCount +=
+                `<span class="password-character password-${type}
+                        ${i % 2 === 0 ? ' bg-grey' : ''}">
```
cscharf, line 18: On the `.bg-grey` class application here, I would prefer that we actually create a new class definition for the alternating items, OR, simply use pseudo classes on the stylesheet side to identify even/odd items in the list for coloration so they're "themeable" vs. hard-coding the color class here.

cscharf (commented):
Just one really minor feedback item but overall looks great!

angular/src/pipes/color-password-count.pipe.ts
```diff
@@ -0,0 +1,24 @@
+import { Pipe } from '@angular/core';
+import { ColorPasswordPipe } from './color-password.pipe';
+
+/*
+ An updated pipe that extends ColourPasswordPipe to include a character count
+*/
+@Pipe({ name: 'colorPasswordCount' })
+export class ColorPasswordCountPipe extends ColorPasswordPipe {
+    transform(password: string) {
+        const passwordArray = Array.from(password);
```
eliykat, line 10: It looks like this logic is the same as the base class. If the only difference is the template literal, could you just make the template literal a property that you override in the subclass, and leave the logic unchanged?

(That said - I like the `getCharacterAndType` refactor either way, it makes everything more readable)

angular/src/components/view.component.ts
```diff
@@ -246,6 +256,16 @@ export class ViewComponent implements OnDestroy, OnInit {
         }
     }
 
+    toggleFieldCount(field: FieldView) {
+        const f = (field as any);
```
eliykat, line 260: Do you need to type cast this to `any`? You added `showCount` as a property in `fieldView`.

eliykat (changes requested):
A couple of questions, but a really nice use of the existing pipe. Keeps it nice and simple.

angular/src/components/view.component.ts
mkanavakatini: Was following convention here but you're right - it's unnecessary - so I've removed all instances.

angular/src/directives/show-more.directive.ts
```diff
@@ -0,0 +1,23 @@
+import { Directive, ElementRef, Input, Renderer2, } from '@angular/core';
```
eliykat, line 1: I'm not sure what this file is, it looks like you've copied the existing `A11yTitleDirective` but haven't done anything more with it. Should it be removed?

angular/src/components/view.component.ts
```diff
@@ -191,11 +193,29 @@ export class ViewComponent implements OnDestroy, OnInit {
...
+        if (!await this.promptPassword()) {
+            return;
+        }
+
+        this.showPasswordCount = !this.showPasswordCount;
+        this.showPassword = this.showPasswordCount;
+        if (this.showPassword) {
+            this.eventService.collect(EventType.Cipher_ClientToggledPasswordVisible, this.cipherId);
+        }
+
```
eliykat, line 212: Delete whitespace

eliykat (commented):
Thanks @mkanavakatini, please:
* address feedback
* merge with master
* make sure you've run `npm run lint` and fixed any feedback

Then we can merge 👍

eliykat:
I've made some changes requested by our design team after some preliminary feedback:
* add accessibility properties to button and add missing UI text
* instead of adding the chevron to the action buttons (showing/hiding the additional action buttons), we decided to revert to the current layout, and instead show the "character count" button when viewing the password (this simplified a lot of the template logic)
* update icon to use our new icon set (requires internal access to our design materials)

I'm requesting a review from the team given that I shouldn't self-review my own code.

MGibson1 (approved):

PR #342, opened 2021-04-13
fix:  cmd + w doesn't close the window on macOS (bitwarden/desktop#714)

kapitainsky:
cmd + w doesn't close the window on macOS - bitwarden/desktop#714

Based on @cscharf [explanation](https://github.com/bitwarden/desktop/issues/714#issuecomment-810554674) - it seems that issue was introduced in response to erroneous Apple Store approver request. 

The code change reverses previously made changes.


cscharf (approved):
Looks good, thanks!

PR #343, opened 2021-04-13
Password reprompt

Hinton:
## Objective
Adds support for requiring master password re-prompt before accessing hidden fields on ciphers. Also marks card numbers as a secret to ensure it's protected.

### Code Changes
- **package.json**: Add SweetAlert as dependency, since reprompt service requires it.
- **src/abstractions/passwordReprompt.service.ts**: New interface for the password re-prompt service.
- **src/angular/components/add-edit.component.ts**: Add support for toggling visibility of credit card number.
- **src/angular/components/view.component.ts**: Add support for password reprompt, make card number hidden by default.
- **src/enums/eventType.ts**: Add new event type for when card number is toggled.
- **src/models/\* & src/services/cipher.service.ts**: Plumbing needed to propagate `passwordPrompt`.
- **src/models/view/cardView.ts**: Add support for hiding card number.
- **src/services/BrowserPasswordReprompt.service.ts**: Implementation of password reprompt service which keeps track of the fields to protect and prompts for confirmation.


src/services/BrowserPasswordReprompt.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export class BrowserPasswordRepromptService implements PasswordRepromptService {
+    constructor(private i18nService: I18nService, private cryptoService: CryptoService) { }
+
+    protectedFields() {
+        return ['TOTP', 'Password', 'H_Field', 'Card Number', 'Security Code'];
+    }
+
+    async showPasswordPrompt() {
+        const result = await Swal.fire({
```
kspearrin, line 15: This seems weird to me to be putting in jslib services. I would expect Swal to be a client-specific implementation of a UI prompt. Perhaps abstracted out into a platform utils function that is called directly here or with some kind of messaging.

Hinton: Yea, I initially wanted to put it in the `PlatformUtilsService` but since it requires the `cryptoService` that creates a cyclic dependency. Hence I extracted it to a separate service which was supposed to be client specific, but it turned out that each of the jsclient could share the same implementation using Swal.

Using messages to communicate with `platformUtilsService` would be nice, but might be tricky using the `inputValidator`.

Hinton: We could refactor it into:

```typescript
async showPasswordPrompt() {
  const passwordValidation = (value: string): Promise<boolean> => {
    return password == expectedhash
  }
  
  this.platformUtilsService.passwordDialog(title, text, validator);
}
```

Hinton: @kspearrin I refactored it, had to resort to messaging in the browser extension https://github.com/bitwarden/browser/pull/1784/files#diff-027b4bec7e9f506ae15598d89d8a2f7754776435e2545a8e868d61f40a3de158 but all in all it seems to work fine.

kspearrin (approved):

PR #359, opened 2021-04-23
Allow external awaits of indexing

MGibson1:
# Overview 

Fix for issue I missed while developing #356. We were getting stuck in an infinite load loop where we were basing logic on a dirty state of search service. This await enables us to wait until an index is complete, then update it rather than being kicked out of indexing early because it is in progress.

This allows `CipherService` to await indexing and restart with the full ciphers list.

# Files Changed
* **search.service**: move indexing hard work to a private method so we can store its promise and allow for outside awaits on it. Add await and clear method that will block until indexing is completed then clear out the index. Also reset indexedEntity value when clearing index.
* **cipher.service**: await and clear index prior to re-indexing if the indexed entity is not the full vault.

### DIFF
--- a/src/services/cipher.service.ts
+++ b/src/services/cipher.service.ts
@@ -295,9 +295,10 @@ export class CipherService implements CipherServiceAbstraction {
     async getAllDecrypted(): Promise<CipherView[]> {
         if (this.decryptedCipherCache != null) {
             const userId = await this.userService.getUserId();
-            if ((this.searchService().indexedEntityId ?? userId) !== userId)
+            if (this.searchService().indexedEntityId ?? userId !== userId)
             {
-                await this.searchService().indexCiphers();
+                await this.searchService().awaitAndClearIndex();
+                this.searchService().indexCiphers();
             }
             return this.decryptedCipherCache;
         }
--- a/src/services/cipher.service.ts
+++ b/src/services/cipher.service.ts
@@ -295,10 +295,9 @@ export class CipherService implements CipherServiceAbstraction {
     async getAllDecrypted(): Promise<CipherView[]> {
         if (this.decryptedCipherCache != null) {
             const userId = await this.userService.getUserId();
-            if (this.searchService().indexedEntityId ?? userId !== userId)
+            if ((this.searchService().indexedEntityId ?? userId) !== userId)
             {
-                await this.searchService().awaitAndClearIndex();
-                this.searchService().indexCiphers();
+                await this.searchService().indexCiphers(userId, this.decryptedCipherCache);
             }
             return this.decryptedCipherCache;
         }
--- a/src/services/search.service.ts
+++ b/src/services/search.service.ts
@@ -17,6 +17,7 @@ export class SearchService implements SearchServiceAbstraction {
     private indexing = false;
     private index: lunr.Index = null;
     private searchableMinLength = 2;
+    private indexingPromise: Promise<void> = null;
 
     constructor(private cipherService: CipherService, private logService: LogService,
         private i18nService: I18nService) {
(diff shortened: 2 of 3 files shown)


src/services/cipher.service.ts
```diff
@@ -295,9 +295,10 @@ export class CipherService implements CipherServiceAbstraction {
     async getAllDecrypted(): Promise<CipherView[]> {
         if (this.decryptedCipherCache != null) {
             const userId = await this.userService.getUserId();
-            if ((this.searchService().indexedEntityId ?? userId) !== userId)
+            if (this.searchService().indexedEntityId ?? userId !== userId)
             {
-                await this.searchService().indexCiphers();
+                await this.searchService().awaitAndClearIndex();
+                this.searchService().indexCiphers();
```
cscharf, line 301: Wouldn't this also need to be `await`'ed since `.indexCiphers()` still returns a `Promise<void>`?

src/services/search.service.ts
```diff
@@ -26,9 +27,17 @@ export class SearchService implements SearchServiceAbstraction {
     }
 
     clearIndex(): void {
+        this.indexedEntityId = null;
         this.index = null;
     }
 
+    async awaitAndClearIndex(): Promise<void> {
+        if (this.indexing && this.indexingPromise != null) {
+            await this.indexingPromise;
```
cscharf, line 36: help me understand the scenario and perhaps it's my limited understanding of async await in the TypeScript world, but if multiple call stacks are awaiting on a promise, or when that scenario in this instance would actually occur in the Angular/SPA or Desktop/nodeJS realm. I guess I'm not understanding when this scenario would actually occur and why then we'd be calling this again before it completed the first time (is indexing done in a background thread, etc.?) as it seems the prior implementation still `await`'ed the result.

src/services/cipher.service.ts
MGibson1: yes, thank you

src/services/search.service.ts
MGibson1: I wasn't able to get it to happen in a debugger, but I think what's going on is navigation while indexing results in triggering more indexing.

This is the flow that this is supposed to prevent:
We are indexing an organization, we navigate to a personal vault before indexing finishes. We'll clear out the index, and start indexing again. Ciphers indexing has a quick return in we're already indexing, so the second call will be ignored and the first index will complete. Only organization items will be indexed.

This method allows us to wait until indexing is done, clear it, and start a new job.

The other option is to rewrite the indexing method to be abortable and make clear index trigger an abort, but I couldn't think of a clean way to do that -- I don't think it's supported natively by typescript es5.

I hope that's clear. It's a little fuzzy in my head too, so I'm afraid I'm not explaining well

src/services/cipher.service.ts
MGibson1: awaiting this method caused things to break again, pointing me off in a new direction again. The flow of this code is really intermingled. ☹️

MGibson1:
OK, so it turns out the infinite loop was because `indexCiphers` was referencing `getAllDecrypted` when no ciphers are specified and `getAllDecrypted` was calling `indexCiphers` with no ciphers specified. I thought the early return would solve this problem, but it wasn't.

Specifying the existing decrypted ciphers cache solves the loop by not calling `getAllDecrypted` again.

src/services/search.service.ts
cscharf: That's a great explanation and now it makes sense, pesky navigation in the JavaScript web world, ruining lives since 1999.

🤔 , thinking through this scenario, can you instead build indexes to a specified `key`, so a hashtable of index references which get used based on the `key` of that index you wish to search on? I'm thinking in terms of search domains, where `lunr` has maybe 5 indexes built (each one on-demand the first time/lazy-loaded), one is your personal, and then perhaps 4 other orgs' vaults. So when searching the appropriate `key` is passed in, referencing the appropriate index.

Or, perhaps build `all the things` in one index, but add an index facet or filter based on the search scope (like an org id)?

Just thinking of ways we can avoid the invalid index state or aborting an index build due to navigation, etc.

cscharf (approved):

PR #360, opened 2021-04-23
Add support to import from Nordpass(.csv)

djsmith85:
As requested by a new user. last week during the "Bitwarden for Beginners"-Crowdcast, Bitwarden is missing an importer for Nordpass. 

This PR changes that 😉

I also fixed a minor thing in the setCardExpiration-method in the baseImporter. I added a removal of all possible whitespace before continuing the extraction.

### DIFF
--- a/src/importers/nordpassCsvImporter.ts
+++ b/src/importers/nordpassCsvImporter.ts
@@ -1,0 +1,139 @@
+import { BaseImporter } from './baseImporter';
+import { Importer } from './importer';
+
+import { ImportResult } from '../models/domain/importResult';
+
+import { CipherView } from '../models/view/cipherView';
+import { LoginView } from '../models/view/loginView';
+
+import { CipherType } from '../enums/cipherType';
+import { SecureNoteType } from '../enums/secureNoteType';
+
+type nodePassCsvParsed = {
+    name: string;
+    url: string;
+    username: string;
+    password: string;
+    note: string;
+    cardholdername: string;
+    cardnumber: string;
+    cvc: string;
+    expirydate: string;
+    zipcode: string;
+    folder: string;
+    full_name: string;
+    phone_number: string;
+    email: string;
+    address1: string;
+    address2: string;
+    city: string;
+    country: string;
+    state: string;
+};
+
+export class NordPassCsvImporter extends BaseImporter implements Importer {
+    parse(data: string): Promise<ImportResult> {
+        const result = new ImportResult();
+        const results: nodePassCsvParsed[] = this.parseCsv(data, true);
+        if (results == null) {
+            result.success = false;
+            return Promise.resolve(result);
+        }
+
+        results.forEach(record => {
+
+            const recordType = this.evaluateType(record);
+            if (recordType === undefined) {
+                return;
+            }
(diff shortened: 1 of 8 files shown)


src/importers/nordpassCsvImporter.ts
```diff
@@ -0,0 +1,139 @@
...
+            this.moveFoldersToCollections(result);
+        }
+
+        result.success = true;
+        return Promise.resolve(result);
+    }
+
+    private evaluateType(record: nodePassCsvParsed): CipherType {
+
+        if (!this.isNullOrWhitespace(record.username)) return CipherType.Login;
```
cscharf, line 113: For control blocks in TypeScript, please include curly braces and nest the 1-liner on the next line, e.g.
```TypeScript
if (something) {
    return something.Else;
}
```

src/importers/nordpassCsvImporter.ts
```diff
@@ -0,0 +1,139 @@
...
+        if (!this.isNullOrWhitespace(record.full_name)) return CipherType.Identity;
+
+        if (!this.isNullOrWhitespace(record.note)) return CipherType.SecureNote;
+
+        return undefined;
+    }
+
+    private processName(cipher: CipherView, fullName: string) {
+
+        if (this.isNullOrWhitespace(fullName)) return;
```
cscharf, line 126: same here (curly nesting)

src/importers/nordpassCsvImporter.ts
```diff
@@ -0,0 +1,139 @@
...
+
+        const nameParts = fullName.split(' ');
+        if (nameParts.length > 0) {
+            cipher.identity.firstName = this.getValueOrDefault(nameParts[0]);
+        }
+        if (nameParts.length === 2) {
+            cipher.identity.lastName = this.getValueOrDefault(nameParts[1]);
+        } else if (nameParts.length === 3) {
```
cscharf, line 134: please consider `} else if (nameParts.length >= 3) {`, simply because of names which may include suffixes, titles or other info, `Gerald Edmond Richardson Jr. III`, we would not parse name in this instance when we could at least capture `Edmond` and `Richardson`; or consider re-joining all name parts past the `[2]`nd index with that value, concatenating the final last name/surname result (e.g. `Edmond` (middle) and `Richardson Jr. III` (lastname)).

cscharf (commented):
Awesome, thanks @djsmith85! Just some minor feedback and formatting nits.

src/importers/nordpassCsvImporter.ts
djsmith85: Added braces as requested

src/importers/nordpassCsvImporter.ts
djsmith85: Added missing braces

src/importers/nordpassCsvImporter.ts
djsmith85: @cscharf Good point and I've added support to parse more complex names and wrote tests to verify.

cscharf (approved):
Looks good, thanks @djsmith85 !

PR #363, opened 2021-04-26
Split jslib into multiple modules

Hinton:
## Objective
Experiment of importing jslib as a "real" dependency which provides an easier way to sync dependency versions across our projects.

- Split jslib into multiple projects, web depends on common and angular.
- Tests are still run from the root project (not sure if we want to change this in the future)
- Added `ttypescript` and `typescript-transform-paths` to resolve the path aliases in tests.

Related to https://github.com/bitwarden/web/pull/951


cscharf (commented):
So, if I’m understanding this correctly, this change would essentially split out jslib into 4 distinct packages/modules… while remaining a single sub-module (jslib) so we reference jslib as a sub-module as we always have (no change there), but then need to map the paths or import each appropriate sub-module where appropriate within the client app that needs them, which in turn pulls in the necessary dependencies/versions as well.. So CLI wouldn’t need the electron or angular modules, and browser wouldn’t need the electron or cli modules, etc., but they all would retain the 1 same jslib sub-module reference…
This in turn would lighten the load of what gets packaged in the delivered apps and of course reduce contention when updating things in the dependency chain.

And the jslib independent module refs wouldn’t necessarily be versioned themselves, they would simply just refer to the module within jslib in the respective package.json file as `0.0.0.0`?

Would there be a reason not to include the @bitwarden/jslib-common module in the other modules' dependencies within jslib (I noticed the module name refs in the paths for those assets)? Or is that already happening and I missed it or just not necessary the way things are split up?

Hinton:
Added `@bitwarden/jslib-common` as dependency to the other modules (had errors when I did that using npm 6 but works fine with 7).

cscharf (approved):
Okay, I think this looks good, let's pull the trigger

cscharf:
After the merge conflicts are handled of course.

cscharf (approved):

cscharf (approved):

PR #365, opened 2021-04-27
Change download to happen in the renderer

Hinton:
## Objective
In #332 I removed the remote module and moved the save file logic to happen in the main process instead. This had the unfortunate side effect of slowing down file saving quite significantly for large files. Changing the logic to use the same download method we use in the browser extensions resolves this since we no longer need to send content between processes.

### Code Changes
- **src/electron/services/electronMainMessaging.service.ts**: Remove `saveFile` ipc handler.
- **src/electron/services/electronPlatformUtils.service.ts**: Change download to happen in the renderer.

### Testing Considerations
Verify downloading attachments works similar as before on all platforms.


MGibson1 (approved):
Nice.

PR #366, opened 2021-04-28
Validate key before using it

Hinton:
## Objective
When running multiple desktop applications (portable and installed) using different accounts it was possible for the application to get confused and use the wrong key to unlock and decrypt items. To avoid this we validate the key prior to returning it and should it be invalid we delete it from the storage.

### Code Changes
- **src/abstractions/crypto.service.ts**: Added new method `validateKey` used to check if the key is valid for the stored encKey.
- **src/services/crypto.service.ts**: Implemented `validateKey` and refactored `shouldStoreKey` to be more readable.

### DIFF
--- a/src/services/crypto.service.ts
+++ b/src/services/crypto.service.ts
@@ -92,10 +92,9 @@ export class CryptoService implements CryptoServiceAbstraction {
         }
 
         const key = await this.secureStorageService.get<string>(Keys.key);
-
         if (key != null) {
             if (!await this.shouldStoreKey()) {
-                console.error("Throwing away stored key since settings have changed")
+                this.logService.warning('Throwing away stored key since settings have changed');
                 this.secureStorageService.remove(Keys.key);
                 return null;
             }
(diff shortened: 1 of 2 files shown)


src/services/crypto.service.ts
```diff
@@ -580,8 +591,35 @@ export class CryptoService implements CryptoServiceAbstraction {
...
     // Helpers
 
+    private async shouldStoreKey() {
+        const vaultTimeout = await this.storageService.get<number>(ConstantsService.vaultTimeoutKey);
+        const biometricUnlock = await this.storageService.get<boolean>(ConstantsService.biometricUnlockKey);
+
+        const biometricsEnabled = biometricUnlock && this.platformUtilService.supportsSecureStorage();
+        const noVaultTimeout = vaultTimeout == null;
+
+        return noVaultTimeout || biometricsEnabled;
```
cscharf, line 620: This hurt my brain, but boolean logic refactor is correct from what I can see as I walk through it.

cscharf (approved):

PR #367, opened 2021-05-04
Default values for cipher.reprompt

eliykat:
## Objective

I'm getting a few errors because the new `cipher.reprompt` property does not have a default value set in some places. e.g.
* if you create a new cipher and save it without changing the value of the reprompt checkbox, it tries to save a null value
* all imports fail because imported ciphers are not initialized with a default reprompt value

On the import/export question, should we export this value so that users can import & export without any change in behaviour? Currently I've just initialized it to `CipherRepromptType.None` on all imports, but users may not be expecting that if they're importing a vault backup. Happy to do this if required as I'm poking around in the import/export code at the moment.

@Hinton I know you're still working on this feature, let me know if this is helpful or if we've crossed paths.

EDIT: looks like unit tests need to be updated, or I've done something wrong... will wait for feedback before spending any more time on it anyway

### DIFF
--- a/src/importers/baseImporter.ts
+++ b/src/importers/baseImporter.ts
@@ -15,6 +15,7 @@ import { FolderView } from '../models/view/folderView';
 import { LoginView } from '../models/view/loginView';
 import { SecureNoteView } from '../models/view/secureNoteView';
 
+import { CipherRepromptType } from '../enums/cipherRepromptType';
 import { CipherType } from '../enums/cipherType';
 import { FieldType } from '../enums/fieldType';
 import { SecureNoteType } from '../enums/secureNoteType';
@@ -292,6 +293,7 @@ export abstract class BaseImporter {
         cipher.fields = [];
         cipher.login = new LoginView();
         cipher.type = CipherType.Login;
+        cipher.reprompt = CipherRepromptType.None;
         return cipher;
     }
 
--- a/src/importers/baseImporter.ts
+++ b/src/importers/baseImporter.ts
@@ -293,7 +293,6 @@ export abstract class BaseImporter {
         cipher.fields = [];
         cipher.login = new LoginView();
         cipher.type = CipherType.Login;
-        cipher.reprompt = CipherRepromptType.None;
         return cipher;
     }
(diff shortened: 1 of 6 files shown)


src/models/export/cipher.ts
```diff
@@ -42,6 +43,7 @@ export class Cipher {
         view.name = req.name;
         view.notes = req.notes;
         view.favorite = req.favorite;
+        view.reprompt = CipherRepromptType.None;
```
Hinton, line 46: Feels like we should be able to use the actual reprompt value here?
```suggestion
        view.reprompt = req.reprompt;
```

src/importers/baseImporter.ts
```diff
@@ -292,6 +293,7 @@ export abstract class BaseImporter {
         cipher.fields = [];
         cipher.login = new LoginView();
         cipher.type = CipherType.Login;
+        cipher.reprompt = CipherRepromptType.None;
```
Hinton, line 296: I believe it should be fine if we correctly handle Bitwarden exports correctly for now, not sure if it's the same mechanism as third party imports?

Hinton:
@eliykat Snap, seems I didn't test adding new ciphers properly after cleaning up the code.

Hinton:
@eliykat I hope you don't mind, but I pushed a commit to your branch with some export & import fixes since I found a bug on the server side and noticed it didn't handle imports correctly.

eliykat:
@Hinton your changes look good to me (pushing to my branch is fine). I added some extra checks on `cipher.toView` and `cipher.toDomain` because those are used by `bitwardenJsonImporter` to parse decrypted and encrypted imports respectively. Otherwise an older export file will overwrite this property with `null`. Let me know if you disagree.

Otherwise I think we can merge this if you want to give an approving review.

Hinton:
@eliykat It worked without setting any value since the import doesn't seem to include non existing values, and the server correctly handles no value. But we might as well keep it since it should prevent issues if we ever change the behavior.

Hinton (approved):

PR #369, opened 2021-05-05
Increase error checking on imported Login items

eliykat:
## Objective

Fix the following defect:
> When a user makes an encrypted export of their vault, rotates their encryption key, and then attempts to import the encrypted JSON, the web vault doesn't always throw an error. Sometimes, corrupted/blank vault items are imported instead.

The problem is that `importService` only checks the first, middle and last items of an import to make sure they parsed/decrypted correctly. All of these 3 items must be "bad" for the import process to be aborted. However, the error checking only works for Login items, any other items will pass through it. Therefore, if a Card, Identity or Secure Note is in the first, middle or last position, the import will always succeed even if decryption failed.

## Code changes

At first I was going to include a hash of the `encKey` in the encrypted JSON, so we could check it hadn't changed before starting the import. However, we salt our hashes with the email address + master password, which means that changing either of these after exporting would cause the import to fail. I couldn't think of any other secure way of including this information or doing this check.

Also note that if the MAC check fails on an `encString`, it just sets the value to `null`, it doesn't throw an error. So this is hard to check for specifically without making larger changes.

Instead, I increased the error checking using the existing logic. Instead of checking first, middle and last items, it just checks all imported ciphers. It allows a margin of error of up to 10 "bad" ciphers before aborting - mostly for other types of imports (e.g. from other password managers), because this change otherwise makes error checking much stricter, and I don't want to abort an import if there's just 1 or 2 malformed items.

I assume that the original intention was to avoid looping through the ciphers again, but really the performance impact from this is negligible, especially compared to the time cost of removing garbage ciphers from your vault.

### DIFF
--- a/src/services/export.service.ts
+++ b/src/services/export.service.ts
@@ -4,6 +4,7 @@ import { CipherType } from '../enums/cipherType';
 
 import { ApiService } from '../abstractions/api.service';
 import { CipherService } from '../abstractions/cipher.service';
+import { CryptoService } from '../abstractions/crypto.service';
 import { ExportService as ExportServiceAbstraction } from '../abstractions/export.service';
 import { FolderService } from '../abstractions/folder.service';
 
@@ -25,7 +26,7 @@ import { FolderWithId as FolderExport } from '../models/export/folderWithId';
 
 export class ExportService implements ExportServiceAbstraction {
     constructor(private folderService: FolderService, private cipherService: CipherService,
-        private apiService: ApiService) { }
+        private apiService: ApiService, private cryptoService: CryptoService) { }
 
     async getExport(format: 'csv' | 'json' | 'encrypted_json' = 'csv'): Promise<string> {
         if (format === 'encrypted_json') {
@@ -141,8 +142,12 @@ export class ExportService implements ExportServiceAbstraction {
 
         await Promise.all(promises);
 
+        const testValue = await this.cryptoService.randomNumber(2048, 65536);
+        const encKeyTest = await this.cryptoService.encrypt(testValue.toString());
+
         const jsonDoc: any = {
             encrypted: true,
+            encKeyTest: encKeyTest.encryptedString,
             folders: [],
             items: [],
         };
--- a/src/services/export.service.ts
+++ b/src/services/export.service.ts
@@ -23,6 +23,7 @@ import { CollectionDetailsResponse } from '../models/response/collectionResponse
 import { CipherWithIds as CipherExport } from '../models/export/cipherWithIds';
 import { CollectionWithId as CollectionExport } from '../models/export/collectionWithId';
 import { FolderWithId as FolderExport } from '../models/export/folderWithId';
+import { Utils } from '../misc/utils';
 
 export class ExportService implements ExportServiceAbstraction {
     constructor(private folderService: FolderService, private cipherService: CipherService,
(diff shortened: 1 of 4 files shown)


MGibson1:
If the existing key hash doesn't work, can't we use a known constant string instead of a key hash?

MAC should be able to be any value that's known, we just stick an encrypted constant in the import object
```
{
  "encrypted": true,
  "MAC": <<encryption of something constant across all exports>>,
  "collections": {...},
  ...
}
```

eliykat:
I considered that (as you mentioned it before), but wouldn't encrypting known cleartext make it more vulnerable to brute force attacks? Obviously you should keep your exports in a safe place, but I wouldn't want to do anything that decreases the overall strength of the encrypted export. (and therefore the user's current vault, which would be using the same encKey)

MGibson1:
I don't think so, but we could also use our implementation of HMAC on the export itself, which I'd be more confident in.

For that I'd recommend making an object of the export items other than `encrypted` and `hmac` and sending it through out `cryptoFunction.service`. That scheme repeats serializations, but maybe you can come up with a way to avoid that.

eliykat:
Hey @MGibson1, I've revisited this and gone with the "canary" approach we discussed on Slack:
* on export, a random string is encrypted (the easiest function I could find to do this was to generate a long random number and then convert it to a string)
* on import, the string is decrypted and its mac checked; if it doesn't match, it pops up an import error to the user
* this check is optional to preserve backwards-compatability with existing exports (i.e. if the test `encString` isn't found, it'll just go ahead with the import anyway)

This takes advantage of our existing mac check on `encStrings` without trying to do anything too dangerous/tricky. Best way I could think to do it.

I've also kept my previous commit because I think it improves error checking generally which isn't a bad thing.

eliykat:
@MGibson1 I implemented the approach we discussed on Slack of using a random "canary" `encString` to decrypt first as a test. If its MAC fails then we abort the import process.

However, I'm fairly indecisive on this, I'm not sure whether this is worth doing as a separate check given that the general `badData` checks (in my first commit) would pick this up anyway. The only advantage is that we can give a more descriptive error message. Thoughts?

src/services/export.service.ts
```diff
@@ -141,8 +142,12 @@ export class ExportService implements ExportServiceAbstraction {
 
         await Promise.all(promises);
 
+        const testValue = await this.cryptoService.randomNumber(2048, 65536);
```
MGibson1, line 145: just curious, why a number? I was expecting Utils.NewGuid() or something

src/services/export.service.ts
```diff
@@ -141,8 +142,12 @@ export class ExportService implements ExportServiceAbstraction {
...
+        const encKeyTest = await this.cryptoService.encrypt(testValue.toString());
+
         const jsonDoc: any = {
             encrypted: true,
+            encKeyTest: encKeyTest.encryptedString,
```
MGibson1, line 150: Just a thought since the import hinges on this MAC.

Perhaps we should call this property `encKeyValidation-DO_NOT_EDIT` since that will result in destroying the export.

MGibson1 (commented):
The reason we shouldn't go back to the `badData` approach in my mind is this: It makes assumptions about the contents of the imported vault. 

It first assumes that Logins will be in the first, middle, or last position. Then your edits assume Logins will be in the first 10 and that there will be 10 ciphers, or it can't fail.

Beyond that, the `badData` method appears to assume all Logins have passwords filled in, which isn't necessarily true.

src/services/export.service.ts
eliykat: Because that's a better idea and I didn't think of that ;)

src/services/export.service.ts
eliykat: Yep good point

eliykat:
I've fixed those couple of items you suggested, thanks!

So the `encKeyValidation` test will make sure that the encryption key is checked before importing - which fixes the defect that was reported.

Both `badData` tests (i.e. before & after my changes) are imperfect and have assumptions baked in. I don't like assuming that login items are in the first, middle and last positions in the import file. However, my changes could prevent imports that have Login items with empty passwords, and I tend to think we should err on the side of allowing the import to proceed - otherwise both CS and users will spend time manually editing their CSVs.

As we've solved the problem another way, I've just reverted those changes.

MGibson1 (dismissed):

MGibson1 (approved):

PR #372, opened 2021-05-07
 Fix null search service checking

MGibson1:
# Overview

#356 incorrectly assumes that the `searchService` generator in `cipherService` will always be available. CLI does not currently use SearchService so this is null. A null check is added where `searchService` is accessed.

# Files Changed
* **cipher.service**: Check null search service 
* **baseProgram.ts**: chalk has been updated and now exports two objects, set the default export as `chalk` for use in responses.

### DIFF
--- a/src/cli/baseProgram.ts
+++ b/src/cli/baseProgram.ts
@@ -1,4 +1,4 @@
-import * as chk from 'chalk';
+import * as chalk from 'chalk';
 
 import { Response } from './models/response';
 import { ListResponse } from './models/response/listResponse';
@@ -6,8 +6,6 @@ import { MessageResponse } from './models/response/messageResponse';
 import { StringResponse } from './models/response/stringResponse';
 
 import { UserService } from '../abstractions/user.service';
-
-const chalk = chk.default;
 
 export abstract class BaseProgram {
     constructor(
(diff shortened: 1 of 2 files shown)


kspearrin (approved):

MGibson1:
@Hinton it looks like these failures are due to the package updates. Does ubuntu not have the latest chalk version available or something?

src/cli/baseProgram.ts
```diff
@@ -7,6 +7,8 @@ import { StringResponse } from './models/response/stringResponse';
 
 import { UserService } from '../abstractions/user.service';
 
+const chalk = chk.default;
```
Hinton, line 10: Remove.

src/cli/baseProgram.ts
```diff
@@ -1,4 +1,4 @@
-import * as chalk from 'chalk';
+import * as chk from 'chalk';
```
Hinton, line 1: Revert

Hinton (changes requested):
@MGibson1 Looks like some old code managed to sneak into `baseProgram`. We no longer need to do `chalk = chk.default`.

Hinton (approved):

PR #380, opened 2021-05-14
Add functionality to search the notes field for partial matches.

gryffs:
This PR is to correct the issue [877](https://github.com/bitwarden/desktop/issues/877) reported on desktop but also confirmed on browser and web.

This will make notes partial match for all TYPES: Login, Card, Identity, and Secure Note. I've uploaded the video to show functionality after adding notes to the wild-card search list.

https://user-images.githubusercontent.com/16548994/118257658-747e2f00-b4e1-11eb-9025-d7b478d9d580.mov

If this looks good, I can update jslib once merged as well.

### DIFF
--- a/src/services/search.service.ts
+++ b/src/services/search.service.ts
@@ -135,6 +135,7 @@ export class SearchService implements SearchServiceAbstraction {
                     q.term(t, { fields: ['name'], wildcard: soWild });
                     q.term(t, { fields: ['subtitle'], wildcard: soWild });
                     q.term(t, { fields: ['login.uris'], wildcard: soWild });
+                    q.term(t, { fields: ['notes'], wildcard: soWild });
                     q.term(t, {});
                 });
             });


cscharf (approved):
Hi @gryffs , thanks for the contribution! The code looks great, thank you! I'm putting a `hold` tag on it for now, however, as need to have a conversation with Kyle and the product team... I'm sure there was a reason at one point notes were explicitly excluded so while it works and is technically correct, not sure we _want_ this change. I'll follow up.

I know in the meantime, advanced search techniques (wildcards) should be used in this instance:
https://bitwarden.com/help/article/searching-vault/

Example:
>notes:ging*

cscharf:
Hi @gryffs , after some discussion we feel this is better left out, for now, given that wildcards are more expensive operations to complete. Notes can be a large amount of data vs a name for other field and therefore it's a bit cumbersome on performance and resources to always perform a wildcard search on notes each time vs. the advanced search method which is more explicit and purposeful. We do have some customers and users with pretty large vaults that this would likely negatively impact performance for.

Really appreciate the PR, however going to close this one. Thanks again!

gryffs:
Hi @cscharf , thank you for the response and clarifying, I understand and appreciate the clarity.

PR #382, opened 2021-05-18
Use organization api key for auth

MGibson1:
# Overview

Related to bitwarden/directory-connector#121.

This PR merges three requirements:
1. Implements classes to allow for authentication via organization Api Keys to a limited `api.organization` scope
2. Implements [public API Organization Import classes](https://github.com/bitwarden/server/blob/master/src/Core/Models/Api/Public/Request/OrganizationImportRequestModel.cs)
3. Removes instances of multiple simultaneous `storageService` writes
   * This was causing issues with directory-connector's [file-locked version of `lowDbStorageService`](https://github.com/bitwarden/directory-connector/blob/master/src/services/lowdbStorage.service.ts)
   * It looks there's a relationship between the number of storage promises and the number of lock retries. When logging out, we were promising ~8 things total, which is too many, resulting in lock acquisition failures.

# Files Changed
* **api.service.ts**: Add public api endpoint for organization import
* **apiKey.service.ts**: Analogous to `UserService` this service exists to house information on api authentication which is not necessarily tied to a user. Makes fewer assumptions about the contents of an authentication token.
* **organizationImportRequest/organizationImportGroupRequest/organizationImportMemberRequest**: These are the public API request contracts for organization import. They are functionally the same as the private api, but have different property names. Constructors include conversions from private api classes to encourage tight coupling through compile-time errors.
* **toknRequest.ts**: organization api keys are more limited in scope than default api keys.
* **auth.service**: Expose necessary services to inheriting classes. This is used for an auth.service in directory-connector which authenticates only via api keys backed by `ApiKeyService` rather than `UserService`.
* **token.service/crypto.service/user.service/vaultTimeout.service**: Serialize `storageService` writes for the benefit of lowDb.


addisonbeck (dismissed):

src/services/apiKey.service.ts
```diff
@@ -0,0 +1,67 @@
...
+        this.clientId = clientId;
+        const idParts = clientId.split('.');
+
+        if (idParts.length !== 2 || !Utils.isGuid(idParts[1])) {
+            throw Error('Invalid clientId');
+        }
+        this.entityType = idParts[0];
+        this.entityId = idParts[1];
+
+        return this.storageService.save(Keys.clientId, this.clientId)
```
cscharf, line 31: Why not just make 3 separate calls using `await`?

src/services/apiKey.service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    async getEntityId(): Promise<string> {
+        if (this.entityId == null) {
+            this.entityId = await this.storageService.get<string>(Keys.entityId);
+        }
+        return this.entityId;
+    }
+
+    async clear(): Promise<any> {
+        await this.storageService.remove(Keys.clientId)
```
cscharf, line 51: Same... the promise return type, with `async` keyword is transpiled anyway to include the promise without the need for this type of chaining (just using `await`).
```TypeScript
await this.storageService.remove(Keys.clientId);
await this.storageService.remove(Keys.entityId);
await this.storageService.remove(Keys.entityType);
```

src/services/apiKey.service.ts
MGibson1: I liked how the language makes it clear things are sequential, but I think that's just in juxtaposition to the `Promise.all` language.

normally I'd go this route so I'll change it

cscharf (approved):

src/services/user.service.ts
```diff
@@ -96,14 +94,12 @@ export class UserService implements UserServiceAbstraction {
...
 
-        await Promise.all([
-            this.storageService.remove(Keys.userId),
-            this.storageService.remove(Keys.userEmail),
-            this.storageService.remove(Keys.stamp),
-            this.storageService.remove(Keys.kdf),
-            this.storageService.remove(Keys.kdfIterations),
-            this.clearOrganizations(userId),
-        ]);
+        await this.storageService.remove(Keys.userId)
```
cscharf, line 97: whoops, missing a semi-colon.

cscharf (changes requested):

addisonbeck (approved):

cscharf (approved):

PR #384, opened 2021-05-19
Fix lockGuard logic

eliykat:
## Objective

Fix https://github.com/bitwarden/web/issues/978:

> After closing bitwarden tab, and reopening, you are not logged in as your account, but rather logged in as $1

## Code changes

This was introduced by the `lockGuardService`, which is meant to only allow navigation to the lock screen if the vault is locked. Thanks to @Hinton for picking this up.

The problem is that we were allowing navigation if `vaultTimeoutService.isLocked() == true`. However, that returns true if the user is logged out as well as if the vault is locked, so it was allowing a logged out user to access the lock page.

I've changed the order of the logic so that it should work as intended. This closely follows `unauthGuardService`.

`web` and `jslib` will need to be bumped. `browser` doesn't use this.

### DIFF
--- a/src/angular/services/lock-guard.service.ts
+++ b/src/angular/services/lock-guard.service.ts
@@ -13,17 +13,17 @@ export class LockGuardService implements CanActivate {
         private router: Router) { }
 
     async canActivate() {
-        const locked = await this.vaultTimeoutService.isLocked();
-        if (!locked) {
-            const isAuthed = await this.userService.isAuthenticated();
-            if (!isAuthed) {
-                this.router.navigate(['login']);
+        const isAuthed = await this.userService.isAuthenticated();
+        if (isAuthed) {
+            const locked = await this.vaultTimeoutService.isLocked();
+            if (locked) {
+                return true;
             } else {
                 this.router.navigate(['vault']);
             }
             return false;
         }
 
-        return true;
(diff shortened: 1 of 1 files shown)


Hinton (commented):
I noticed one of the paths doesn't return a boolean. I also took a look at the documentation and it seems Angular 7.1+ supports returning an `UrlTree` for redirects. https://angular.io/api/router/CanActivate To me this feels a bit cleaner than calling `this.router.navigate`.

eliykat:
I've experimented with returning a `UrlTree`, because I agree that would be cleaner. However, it worked slightly differently and I wasn't happy with it - it's easiest to explain with an example:
* I am on the `/vault` page
* I try to navigate to `/lock`
* the guard will catch me and redirect me back to `/vault`; but because I'm already there, it won't actually navigate anywhere (fine) but will still leave the `/lock` URL in my address bar (confusing).

This only happens when it's redirecting back to the same page you're already on, so not all the time, but it still felt weird and buggy to me compared to the current behaviour. I looked at the options you can pass in to `createUrlTree` but couldn't find any way around this; maybe if it supported `replaceUrl`, but it doesn't.

Anyway, I thought I'd leave it as is but just insert the missing return values. Let me know if you think I'm missing something.

PR #386, opened 2021-05-20
Bulk confirm

Hinton:
## Objective
Confirming multiple users is currently time consuming and involves multiple steps. This PR adds two new endpoints, one for retrieving the public key of multiple organization users, and another for confirming multiple users.

Related: https://github.com/bitwarden/server/pull/1345


addisonbeck (approved):
🥳

PR #399, opened 2021-06-04
[Provider] Add initial support for providers

Hinton:
## Objective
Adds the necessary code for providers to work. Implements multiple APIs, hooks into userService, syncService and cryptoService for provider related operations.

## Noteworthy changes
* **common/src/abstractions/crypto.service.ts**: Added `providerOrgs` to `setOrgKeys`, added provider related operations.
* **common/src/services/sync.service.ts** Added sync logic for providers and their orgs.
* **common/src/services/user.service.ts**: Modified `getAllOrgs` to only return orgs the user has direct access to. Added provider related methods.
* **common/src/models/data/organizationData.ts**: Added provider ID and name to showcase whom is managing this org. Also added a flag `isProviderUser` which is used to identify if the user has access through a provider. (Hide from list)


common/src/services/sync.service.ts
```diff
@@ -286,15 +287,31 @@ export class SyncService implements SyncServiceAbstraction {
...
+            providers[p.id] = new ProviderData(p);
+        });
+
+        response.providerOrganizations.forEach(o => {
+            if (organizations[o.id] == null) {
+                organizations[o.id] = new OrganizationData(o);
+                organizations[o.id].isProviderUser = true;
+            }
+        });
+        return Promise.all([
```
cscharf, line 311: Just want to ensure that on the server-side, we're not going to have modification date collisions on the Organization table for any updates, or table locks if these aren't done in sequence (vs. parallel).

cscharf (commented):
Just a few small questions, otherwise looks good!

common/src/models/data/organizationData.ts
```diff
@@ -31,6 +31,9 @@ export class OrganizationData {
     resetPasswordEnrolled: boolean;
     userId: string;
     hasPublicAndPrivateKeys: boolean;
+    providerId: string;
+    providerName: string;
+    isProviderUser: boolean;
```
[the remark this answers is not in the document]
Hinton: `isProviderUser` is set explicitly by `SyncService`. I'm not quite happy with the approach but it seems the easiest way to allow us to get away with a single data structure.

common/src/services/sync.service.ts
Hinton: This only sets the values locally. We don't send any data to the server.

cscharf (approved):

PR #407, opened 2021-06-10
Add backwards compatability for new local hashing method

eliykat:
## Objective

Follow up work from #404. Changing the local password hashing method would have stopped users from unlocking after the update (at least until they logged out and back in again). This checks to see whether they're using an outdated hash and updates it gracefully.

### DIFF
--- a/common/src/services/crypto.service.ts
+++ b/common/src/services/crypto.service.ts
@@ -139,6 +139,7 @@ export class CryptoService implements CryptoServiceAbstraction {
                 return true;
             }
 
+            // TODO: remove serverKeyHash check in 1-2 releases after everyone's keyHash has been updated
             const serverKeyHash = await this.hashPassword(masterPassword, key, HashPurpose.ServerAuthorization);
             if (serverKeyHash != null && storedKeyHash === serverKeyHash) {
                 await this.setKeyHash(localKeyHash);
--- a/common/src/services/crypto.service.ts
+++ b/common/src/services/crypto.service.ts
@@ -131,6 +131,24 @@ export class CryptoService implements CryptoServiceAbstraction {
         return keyHash == null ? null : this.keyHash;
     }
 
+    async compareAndUpdateKeyHash(masterPassword: string, key: SymmetricCryptoKey): Promise<boolean> {
+        const storedKeyHash = await this.getKeyHash();
+        if (masterPassword != null && storedKeyHash != null) {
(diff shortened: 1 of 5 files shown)


common/src/services/crypto.service.ts
```diff
@@ -131,6 +131,24 @@ export class CryptoService implements CryptoServiceAbstraction {
...
 
+    async compareAndUpdateKeyHash(masterPassword: string, key: SymmetricCryptoKey): Promise<boolean> {
+        const storedKeyHash = await this.getKeyHash();
+        if (masterPassword != null && storedKeyHash != null) {
+            const localKeyHash = await this.hashPassword(masterPassword, key, HashPurpose.LocalAuthorization);
+            if (localKeyHash != null && storedKeyHash === localKeyHash) {
+                return true;
+            }
+
+            const serverKeyHash = await this.hashPassword(masterPassword, key, HashPurpose.ServerAuthorization);
```
kspearrin, line 142: May want to put a TODO comment here that we can remove this at some point in the future?

common/src/services/passwordReprompt.service.ts
```diff
@@ -16,13 +16,7 @@ export class PasswordRepromptService implements PasswordRepromptServiceAbstracti
...
     async showPasswordPrompt() {
         const passwordValidator = async (value: string) => {
-            const keyHash = await this.cryptoService.hashPassword(value, null, HashPurpose.LocalAuthorization);
-            const storedKeyHash = await this.cryptoService.getKeyHash();
-
-            if (storedKeyHash == null || keyHash == null || storedKeyHash !== keyHash) {
-                return false;
-            }
-            return true;
+            return await this.cryptoService.compareAndUpdateKeyHash(value, null);
```
kspearrin, line 19: async/await no longer necessary here. just return `compareAndUpdateKeyHash` unawaited.

kspearrin (changes requested):

common/src/services/passwordReprompt.service.ts
eliykat: Is this because `showPasswordDialog` already `await`s `passwordValidator` and will therefore `await` the promise it returns?

Should the same change be made in mobile? https://github.com/bitwarden/mobile/pull/1423/files#diff-e6b6724aaae6e1c5614946a04739af5d574b83805a6827452e7b0695c423f4cdR32

kspearrin (approved):

PR #409, opened 2021-06-14
Fix separate key storage for non desktop

MGibson1:
# Overview

#402 Splits out cryptographic secure storage into two key locations for Desktop clients. We intentionally put this implementation at the storage service level so other clients didn't need to do it. However, I failed to correctly identify when these items should be dealt with separately in `CryptoService`.

This checks if the service is running in a desktop client environment and clears keys accordingly.

Also fixed an issue where `VaultTimeoutService` wasn't being properly reset upon logout and was requesting unlock immediately following login.

# Files Changed
* **platformUtils.service**: Add boolean which specifies if the client is a desktop client
* **crypto.service**: handle desktop client key storage separately since we don't want to remove the sole key on non-desktop if biometric isn't to be stored.
* **vaultTimeout.service**: Reset state on `clear()`

### DIFF
--- a/common/src/abstractions/platformUtils.service.ts
+++ b/common/src/abstractions/platformUtils.service.ts
@@ -4,6 +4,7 @@ export abstract class PlatformUtilsService {
     identityClientId: string;
     getDevice: () => DeviceType;
     getDeviceString: () => string;
+    isDesktopClient: boolean;
     isFirefox: () => boolean;
     isChrome: () => boolean;
     isEdge: () => boolean;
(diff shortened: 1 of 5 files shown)


cscharf (commented):
This looks good to me, asked Oscar to take a quick look/review as well just to be sure for final approval.

cscharf (approved):

MGibson1:
Me too, thanks for the suggestion

Hinton (approved):

cscharf:
Oscar also felt it looked good, so we're good to go

PR #410, opened 2021-06-16
Add null checks for TrayMain.tray

eliykat:
## Objective

Fix bitwarden/desktop#934, which was caused by calling `trayMain.tray.setContextMenu` when `trayMain.tray` was null. This only occurs on Linux.

`trayMain.tray` may be null if:
* the user has not enabled minimise to tray in preferences - `init()` does not call `this.showTray()`, which initialises `this.tray`
* the OS does not have a tray and therefore `this.tray` may not be initialized properly

## Code changes
* early return from `showTray()` if we can't initialize the tray properly
* extra null check on `updateContextMenu()`

I'll need to bump `jslib` on `desktop` after this is merged.

### DIFF
--- a/electron/src/tray.main.ts
+++ b/electron/src/tray.main.ts
@@ -120,6 +120,10 @@ export class TrayMain {
         }
 
         this.tray = new Tray(this.icon);
+        if (this.tray == null) {
+            return;
+        }
+
         this.tray.setToolTip(this.appName);
         this.tray.on('click', () => this.toggleWindow());
         this.tray.on('right-click', () => this.tray.popUpContextMenu(this.contextMenu));
@@ -133,7 +137,7 @@ export class TrayMain {
     }
 
     updateContextMenu() {
-        if (this.contextMenu != null && this.isLinux()) {
+        if (this.contextMenu != null && this.tray != null && this.isLinux()) {
             this.tray.setContextMenu(this.contextMenu);
         }
     }


Hinton (dismissed):

WAPEETY (approved):
makes sense this fix

cscharf:
@bitwarden/dept-qa , FYI - regression candidate; please evaluate for inclusion or not in upcoming release

eihli:
```
        this.tray = new Tray(this.icon);
        if (this.tray == null) {
            return;
         }
```

Doesn't `new` always return `this` and would never be `null`?

Also, I tried adding the null check in `updateContextMenu` in my own code and it stopped the app from throwing that error but it also *never* become responsive (compared to previously when it would be unresponsive *until* the error was thrown).

cscharf:
> ```
>         this.tray = new Tray(this.icon);
>         if (this.tray == null) {
>             return;
>          }
> ```
> 
> Doesn't `new` always return `this` and would never be `null`?
> 
> Also, I tried adding the null check in `updateContextMenu` in my own code and it stopped the app from throwing that error but it also _never_ become responsive (compared to previously when it would be unresponsive _until_ the error was thrown).

That's a really good point @eliykat , this doesn't necessarily seem like the appropriate fix.

eliykat:
@eihli I was unable to reproduce the issue in StumpWM on Ubuntu. I did get the error message, but like other platforms, I didn't experience any freezing/unresponsiveness. Very strange that the null check makes the problem worse, not better - if `this.tray != null` is false, then it would only shortcut the call to `this.isLinux()` and `this.tray.setContextMenu`, and I don't understand why either of those would _solve_ the unresponsiveness.

Any other debug info you can provide would be welcome.

PS. I spent a good amount of time learning StumpWM hotkeys!

WAPEETY:
> @eihli I was unable to reproduce the issue in StumpWM on Ubuntu. I did get the error message, but like other platforms, I didn't experience any freezing/unresponsiveness. Very strange that the null check makes the problem worse, not better - if `this.tray != null` is false, then it would only shortcut the call to `this.isLinux()` and `this.tray.setContextMenu`, and I don't understand why either of those would _solve_ the unresponsiveness.
> 
> Any other debug info you can provide would be welcome.
> 
> PS. I spent a good amount of time learning StumpWM hotkeys!

Instead I had a lot of problem, the application won't start:
https://github.com/bitwarden/desktop/issues/956
Now I think that my problem is different...

eihli:
Sorry for making notes in multiple places. I've been updating the issue with my notes.

Here's the result of a few console.logs I threw around. This includes a console.log in showTray, which you'll notice is never called.
```
➜  desktop git:(rc) ✗ ELECTRON_IS_DEV=1 ./dist/Bitwarden-1.26.5-x86_64.AppImage
2021-06-17T13:16:29.425Z  messagingMain initialized from main.ts
2021-06-17T13:16:29.965Z trayMain about to init from main.ts
2021-06-17T13:16:29.966Z tray.main.ts init pre-await
2021-06-17T13:16:29.966Z tray.main.ts init post-await
2021-06-17T13:16:29.966Z awaiting storageService from main.ts#bootstrap
2021-06-17T13:16:30.034Z bootstrap complete
(node:71706) electron: The default of contextIsolation is deprecated and will be changing from false to true in a future release of Electron.  See https://github.com/electron/electron/issues/23506 for more information
2021-06-17T13:16:30.970Z received message { command: 'syncStarted' }
2021-06-17T13:16:31.013Z received message { command: 'syncCompleted', successfully: false }
2021-06-17T13:16:31.068Z received message { command: 'authBlocked' }
// **** ~35 second delay... app frozen.
2021-06-17T13:18:57.094Z received message { command: 'updateAppMenu', isAuthenticated: false, isLocked: true }
2021-06-17T13:18:57.096Z updateAppMenu message
updateTrayMenu
Thu Jun 17 2021 08:18:57 GMT-0500 (Central Daylight Time) updateContextMenu
(node:71706) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'setContextMenu' of undefined
```

The delay varies between ~20 seconds and ~40 seconds, so maybe something IO-bound rather than a hard-coded timeout?

Still narrowing in on what happens between bootstrapping and the first updateAppMenu message.

And most importantly...

I guess TrayMain was pulled out of this repo and into the shared repo https://github.com/bitwarden/jslib/blob/master/electron/src/tray.main.ts.

That's what main.ts is importing.

```
import { TrayMain } from 'jslib-electron/tray.main';
import { UpdaterMain } from 'jslib-electron/updater.main';
import { WindowMain } from 'jslib-electron/window.main';
```

eliykat:
This fix wasn't the right approach and the issue needs further review and evaluation. Closing.

PR #411, opened 2021-06-16
Improve URL parsing

eliykat:
## Objective

See more info here: https://app.asana.com/0/1169444489336079/1200474469788269/f.

## Code changes

* strip dollar signs from URI before calling `getDomain()`.

`browser/jslib` will need to be bumped after this is merged.

### DIFF
--- a/common/src/misc/utils.ts
+++ b/common/src/misc/utils.ts
@@ -221,6 +221,11 @@ export class Utils {
         if (httpUrl) {
             try {
                 const url = Utils.getUrlObject(uriString);
+                const validHostname = tldjs?.isValid != null ? tldjs.isValid(url.hostname) : true;
+                if (!validHostname) {
+                    return null;
+                }
+
                 if (url.hostname === 'localhost' || Utils.validIpAddress(url.hostname)) {
                     return url.hostname;
                 }
--- a/common/src/misc/tldjs.noop.ts
+++ b/common/src/misc/tldjs.noop.ts
@@ -1,3 +1,7 @@
 export function getDomain(host: string): string | null {
     return null;
 }
+
+export function isValid(host: string): boolean {
+    return true;
+}
--- a/spec/common/misc/utils.spec.ts
+++ b/spec/common/misc/utils.spec.ts
@@ -35,8 +35,8 @@ describe('Utils Service', () => {
         });
 
         it('should reject invalid hostnames', () => {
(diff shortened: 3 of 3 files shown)


Hinton:
What happens if you use another character like `!`? Tldjs has a `isValid` function could potentially be used (return null on invalid). https://github.com/oncletom/tld.js/blob/b758887c0d01a12b6c0f048a544377e8d44368db/lib/is-valid.js#L40

common/src/misc/utils.ts
```diff
@@ -203,7 +203,7 @@ export class Utils {
             return null;
         }
 
-        uriString = uriString.trim();
+        uriString = uriString.replace(/\$/g, '').trim();
```
cscharf, line 206: I agree with Oscar, please remove this line

common/src/misc/utils.ts
```diff
@@ -203,7 +203,7 @@ export class Utils {
...
         if (uriString === '') {
             return null;
         }
```
cscharf, line 209: Down below line 213 after the `data:` check. Add another  check `tldjs.isValid()` and if not valid, then return `null`. Follow the patterns for calling `tldjs` as also seen below.

cscharf (changes requested):

eliykat:
I've updated this to use `tldjs.isValid`. Note that the type definitions for `isValid` indicate that it's deprecated in favour of `isValidHostname`. However, `isValidHostname` isn't actually implemented in the installed npm package in `node_modules`. There seems to be [some confusion about this](https://github.com/oncletom/tld.js/issues/133). I think it's a mismatch between the type definitions and the current npm package that we can ignore.

Also, `isValid` only accepts a hostname (not full url), however we already have `Util.getHostname` which doesn't use `tldjs` and which appears to parse symbols correctly, so I thought that was ok to use for this purpose.

common/src/misc/utils.ts
```diff
@@ -221,6 +221,11 @@ export class Utils {
         if (httpUrl) {
             try {
                 const url = Utils.getUrlObject(uriString);
+                const validHostname = tldjs?.isValid != null ? tldjs.isValid(url.hostname) : true;
```
Hinton, line 224: Do we need to update our noop implementation as well? (Always return true perhaps)

cscharf (approved):

PR #412, opened 2021-06-17
Add org name to moved to org success toast

MGibson1:
# Overview

The term "Share" is confusing to users because it implies maintaining ownership and control of a cipher while allowing others to access it as well. It also implies the ability to revoke access or to stop sharing -- neither of these things is true.

When a cipher is "shared" it is irrevocably taken over by an organization. This occurs for good cryptographic and security reasons. The better solution to this is to revise the communication around "sharing." To that end, this work changes the term "share" to "move to organization." 

Most of these changes will occur in specific clients. PRs will reference this one.

# Files Changed
* **share.component.ts**: Renamed translation item and added cipher and org names

# Screen shots
image


eliykat (approved):
I like it! Much clearer.

PR #414, opened 2021-06-21
Persist API key creds for token refresh.

MGibson1:
# Overview

Linked to bitwarden/directory-connector#135

#382 and bitwarden/directory-connector#121 switched authentication to organization API key flow. However, this authentication does not include refresh tokens.

This is groundwork for persisting api key credentials and overriding token refresh with simply re-authenticating with api key creds.

> Note: This PR will need to be cherry-picked into `rc` and subsequently into the clients. Please evaluate accordingly

# Files Changed
* **apiKey.service.ts**: Persist and retrieve api key credentials
* **api.service.ts**: Update `doRefreshToken` private method to return nothing, and allow for inherited override. Return signature change is due to auth flow not receiving an `IdentityResponse` from the `apiService`. At any rate, we were only interested in grabbing the new token, which can be done from `TokenService`.

### DIFF
--- a/common/src/abstractions/apiKey.service.ts
+++ b/common/src/abstractions/apiKey.service.ts
@@ -1,6 +1,8 @@
 export abstract class ApiKeyService {
-    setInformation: (clientId: string) => Promise<any>;
+    setInformation: (clientId: string, clientSecret: string) => Promise<any>;
     clear: () => Promise<any>;
+    getClientId: () => Promise<string>;
+    getClientSecret: () => Promise<string>;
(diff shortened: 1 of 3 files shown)


Hinton (approved):

Hinton:
LGTM, but would be nice to get a 2nd pair of eyes with someone more knowledgeable of the token/refresh system.

cscharf (approved):
I also _think_ this looks good.

kspearrin (approved):

PR #419, opened 2021-06-29
[macOS] Don't enable secure input when app is not in focus

eliykat:
See https://github.com/bitwarden/desktop/pull/970


MGibson1 (approved):
This looks good to me, I commented on bitwarden/desktop#970 with, I think, some issue with the implementation.

If this won't require more changes, feel free to merge.

PR #423, opened 2021-07-03
Make `login` behaviour identical to `unlock` if `--passwordenv` is given

Philser:
Follow-up PR to https://github.com/bitwarden/cli/pull/347

- Added a log message when given `passwordenv` var is not set

### DIFF
--- a/node/src/cli/commands/login.command.ts
+++ b/node/src/cli/commands/login.command.ts
@@ -21,6 +21,7 @@ import { MessageResponse } from '../models/response/messageResponse';
 
 import { NodeUtils } from 'jslib-common/misc/nodeUtils';
 import { Utils } from 'jslib-common/misc/utils';
+import { ConsoleLogService } from 'jslib-common/services/consoleLog.service';
 
 // tslint:disable-next-line
 const open = require('open');
@@ -32,6 +33,7 @@ export class LoginCommand {
     protected clientId: string;
 
     private ssoRedirectUri: string = null;
+    private logService: ConsoleLogService;
 
     constructor(protected authService: AuthService, protected apiService: ApiService,
         protected i18nService: I18nService, protected environmentService: EnvironmentService,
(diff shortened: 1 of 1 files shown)


Philser:
@MGibson1 Bump, while the PR is still relevant

node/src/cli/commands/login.command.ts
```diff
@@ -32,13 +33,15 @@ export class LoginCommand {
...
     private ssoRedirectUri: string = null;
+    private logService: ConsoleLogService;
 
     constructor(protected authService: AuthService, protected apiService: ApiService,
         protected i18nService: I18nService, protected environmentService: EnvironmentService,
         protected passwordGenerationService: PasswordGenerationService,
         protected cryptoFunctionService: CryptoFunctionService, protected platformUtilsService: PlatformUtilsService,
         clientId: string) {
         this.clientId = clientId;
+        this.logService = new ConsoleLogService(false);
```
MGibson1, line 44: This should be passed in as a parameter so the application can determine if it's a dev instance and alter log filtering as appropriate.

I missed this in your work on the `unlock` command, sorry 😞

MGibson1 (changes requested):
Sorry for the long lead time -- I took some time off last week.

This looks almost there to me. Unfortunately I missed in the last PR that the `ConsoleLogService` should be passed in as a dependency rather than instantiated in-class.

There will be constructor changes in [cli](https://github.com/bitwarden/cli/blob/master/src/program.ts#L138-L141) and [directory-connector](https://github.com/bitwarden/cli/blob/master/src/program.ts#L138-L141) due to this dependency change.

node/src/cli/commands/login.command.ts
Philser: Will make the appropriate changes and then create a new PR for the bitwarden cli as well

Philser:
@MGibson1 Applied your feedback. Also, I don't know what went wrong but there's a second version of me listed as one of the committers and I cannot figure out how to tell your licensing tool that this is actually the same account.

MGibson1:
@bitwarden/dept-devops anything we can do to get the cla bot to get it together?

Hinton:
> @bitwarden/dept-devops anything we can do to get the cla bot to get it together?

Two of the commits uses a different email that has to be added to the GH account. (Squashing the commits into a single one should also work since the new commit will contain the proper email and be identified as belonging to the correct account.

Hinton:
Hi,

PR #424, opened 2021-07-05
Transpile unicode property escape regex for backwards compatibility

eliykat:
## Objective

Fix https://github.com/bitwarden/browser/issues/1931 by reverting [this commit](https://github.com/bitwarden/jslib/commit/b6f102938fe7c17631cb1b2e356438c5e4456529).

That commit added emoji support in Angular clients, but broke compatibility with Firefox 77 and earlier. That's because it used a unicode property escape, which [isn't supported before Firefox 78](https://caniuse.com/mdn-javascript_builtins_regexp_property_escapes).  Firefox 77 is only just more than 12 months old (released 2 June 2020) and may still be used by enterprise customers. 

Usually we'd fix this by testing for support first before calling the new API, e.g. `if (chrome.api.fancyNewFeature != null)`, but in this case I don't know how to test support for a regex syntax. As soon as it's used by a non-supporting browser, it throws an error and prevents the Angular client from loading. In the short term I think it's better to revert it and revisit the feature later.

I suggest this is cherry-picked to `rc` for any upcoming hotfix, but that's open for discussion.

jslib will also need to be updated in browser (and browser `rc` if applicable).


eliykat:
Reverting to draft while I investigate a transpiling workaround. Would be nice not to throw out this work altogether.

eliykat:
I've transpiled this to a backwards-compatible regex string, per @Hinton's recommendation.

The line is obscenely long which breaks our usual conventions. However, the alternative would be to break it up into multiple regex expressions and join them at runtime, but to me that's just adding cost and wouldn't necessarily make it any clearer. We could do something like [this](https://stackoverflow.com/a/34755045) if anyone disagrees though.

Hinton (approved):
Well, it's not quite as pretty as `Emoji_presentation` but should do the trick 👍

PR #428, opened 2021-07-15
Resolved Safari Date/Time Issues In Send

addisonbeck:
### Changes
* Split the `deletionDate` and `expirationDate` inputs for Sends into a separate component - `efflux-dates.component`
* Refactored the backing logic for the `efflux-dates` component to use Reactive Forms and to consolidate a lot of code that was poorly written (It was mine I can say that). These changes address the bug in the attached ticket, cut down on lines of code, improve type safety and readability, and consolidate logic that previously was split based on which browser is being used whenever possible.

### Related Work
https://github.com/bitwarden/web/pull/1084
https://github.com/bitwarden/browser/pull/1955
https://github.com/bitwarden/desktop/pull/992

### DIFF
--- a/angular/src/components/send/efflux-dates.component.ts
+++ b/angular/src/components/send/efflux-dates.component.ts
@@ -44,10 +44,10 @@ enum DatePreset {
 }
 
 // TimeOption is used for the dropdown implementation of custom times
-// Standard = displayed time; Military = time used in logic
+// Twelve = displayed time; TwentyFour = time used in logic
 interface TimeOption {
-    standard: string;
-    military: string;
+    Twelve: string;
+    TwentyFour: string;
 }
 
 @Directive()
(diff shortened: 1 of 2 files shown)


angular/src/components/send/efflux-dates.component.ts
```diff
@@ -0,0 +1,325 @@
...
+    ThirtyDays = 720,
+    Custom = 0,
+    Never = null,
+}
+
+// TimeOption is used for the dropdown implementation of custom times
+// Standard = displayed time; Military = time used in logic
+interface TimeOption {
+    standard: string;
+    military: string;
```
cscharf, line 50: Small nit... Should we instead call this `twentyFourHour` or `24hour` if possible? Military time is a US construct in the auspice that only the US military uses it within the US, but the rest of the world commonly uses it and calls it 24-hour time or just "time".

cscharf (dismissed):
Looks good, one suggestions/comment, up to you if you want to implement it or not, but otherwise I _think_ this looks good, of course pending that is fixes the issue 😁

angular/src/components/send/efflux-dates.component.ts
addisonbeck: Changed to `twelveHour` and `twentyFourHour`, since `12hour`/`24hour` aren't an option.

I did learn that some folks call the 24 hour clock "railroad time" and in the U.K. it is called "continental time" sometimes, so there's some trivia for the future.

cscharf (approved):

PR #430, opened 2021-07-17
Feature/use hcaptcha if bot

MGibson1:
# Overview

Adds captcha iframe to login component. The actual iframe connector is a part of the [web](bitwarden/web) repo (**PR incoming**). 

The captcha is displayed when the server responds with a [captcha required](https://github.com/bitwarden/server/blob/master/src/Core/IdentityServer/ResourceOwnerPasswordValidator.cs#L68) response. Successful captcha is required to proceed. Order of auth is 1) user/pass 2) (ifbot) user/pass/captcha 3) two-factor

We also now have two iframe connectors, so I extracted the messaging and initialization stuff into an `iframe_component` class.

# Files Changed
* **login.component**: Add Captcha component and handle the captcha required invalid grant error.
* **api.service**: handle the captcha required identity response
* **auth.service**: Add captchaToken to user/pass login.
* **captcha**: Handle initialization of captcha iframe component
* **webauthn/iframe_component**: Extract iframe messaging from webauthn to a base class

### DIFF
--- a/angular/src/components/login.component.ts
+++ b/angular/src/components/login.component.ts
@@ -19,7 +19,7 @@ import { StorageService } from 'jslib-common/abstractions/storage.service';
 
 import { ConstantsService } from 'jslib-common/services/constants.service';
 
-import { Captcha } from 'jslib-common/misc/captcha';
+import { CaptchaIFrame } from 'jslib-common/misc/captcha_iframe';
 import { Utils } from 'jslib-common/misc/utils';
 
 const Keys = {
@@ -36,7 +36,7 @@ export class LoginComponent implements OnInit {
     showPassword: boolean = false;
     captchaSiteKey: string = null;
     captchaToken: string = null;
-    captcha: Captcha;
+    captcha: CaptchaIFrame;
     formPromise: Promise<AuthResult>;
     onSuccessfulLogin: () => Promise<any>;
     onSuccessfulLoginNavigate: () => Promise<any>;
@@ -71,7 +71,7 @@ export class LoginComponent implements OnInit {
         if (webVaultUrl == null) {
             webVaultUrl = 'https://vault.bitwarden.com';
         }
-        this.captcha = new Captcha(window, webVaultUrl,
+        this.captcha = new CaptchaIFrame(window, webVaultUrl,
             this.i18nService, (token: string) => {
                 this.captchaToken = token;
             }, (error: string) => {
--- a/angular/src/components/login.component.ts
+++ b/angular/src/components/login.component.ts
@@ -41,7 +41,6 @@ export class LoginComponent implements OnInit {
     onSuccessfulLogin: () => Promise<any>;
     onSuccessfulLoginNavigate: () => Promise<any>;
     onSuccessfulLoginTwoFactorNavigate: () => Promise<any>;
-    onCaptchaRequired: () => Promise<any>;
 
     protected twoFactorRoute = '2fa';
     protected successRoute = 'vault';
@@ -111,9 +110,6 @@ export class LoginComponent implements OnInit {
             if (!Utils.isNullOrWhitespace(response.captchaSiteKey)) {
                 this.captchaSiteKey = response.captchaSiteKey;
                 this.captcha.init(response.captchaSiteKey);
-                if (this.onCaptchaRequired != null) {
-                    this.onCaptchaRequired();
-                }
             } else if (response.twoFactor) {
                 if (this.onSuccessfulLoginTwoFactorNavigate != null) {
                     this.onSuccessfulLoginTwoFactorNavigate();
--- a/angular/src/components/login.component.ts
+++ b/angular/src/components/login.component.ts
@@ -19,6 +19,7 @@ import { StorageService } from 'jslib-common/abstractions/storage.service';
 
 import { ConstantsService } from 'jslib-common/services/constants.service';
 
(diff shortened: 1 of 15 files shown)


angular/src/components/login.component.ts
```diff
@@ -33,10 +34,14 @@ export class LoginComponent implements OnInit {
...
     masterPassword: string = '';
     showPassword: boolean = false;
+    captchaSiteKey: string = null;
+    captchaToken: string = null;
+    captcha: Captcha;
     formPromise: Promise<AuthResult>;
     onSuccessfulLogin: () => Promise<any>;
     onSuccessfulLoginNavigate: () => Promise<any>;
     onSuccessfulLoginTwoFactorNavigate: () => Promise<any>;
+    onCaptchaRequired: () => Promise<any>;
```
Hinton, line 44: Any particular reason this is handled as a callback? If not would it be appropriate to change `LoginComponent` to be an abstract class forcing this method to be implemented?

common/src/misc/iframe_component.ts
```diff
@@ -0,0 +1,81 @@
...
+            data: this.base64Encode(JSON.stringify(data)),
+            parent: encodeURIComponent(this.win.document.location.href),
+            v: '1',
+        });
+    }
+
+    protected initComponent(params: URLSearchParams): void {
+        this.connectorLink.href = `${this.webVaultUrl}/${this.path}?${params}`;
+        this.iframe = this.win.document.getElementById(this.iframeId) as HTMLIFrameElement;
+        this.iframe.allow = 'publickey-credentials-get ' + new URL(this.webVaultUrl).origin;
```
Hinton, line 51: This is only needed for `WebAuthn`. While I don't see how this could be used negatively it's nice if we don't give more access than we absolutely need.

Hinton (commented):
A few thoughts.

angular/src/components/login.component.ts
MGibson1: It's because it was only needed for Desktop, where the login component width is too small to show the full captcha, so the `OnCaptchaRequired` is used to expand the `container` div to an appropriate width

Hinton: In that case, we could perhaps extend submit, and check if `captchaSiteKey` is set?

common/src/misc/webauthn_iframe.ts
```diff
@@ -20,6 +20,7 @@ export class WebAuthnIFrame extends IFrameComponent {
             this.platformUtilsService.launchUri(`${this.webVaultUrl}/webauthn-fallback-connector.html?${params}`);
         } else {
             super.initComponent(params);
+            this.iframe.allow = 'publickey-credentials-get ' + new URL(this.webVaultUrl).origin;
```
Hinton, line 23: Did we test this? I'm curious if we can set the `src` prior to setting `allow`?

MGibson1: It worked for me, yeah. I was kind of surprised by that too, but I just re-tested in an incognito window so I don't think it's a caching false-positive

angular/src/components/login.component.ts
```diff
@@ -61,6 +65,20 @@ export class LoginComponent implements OnInit {
         if (Utils.isBrowser && !Utils.isNode) {
             this.focusInput();
         }
+
+        let webVaultUrl = this.environmentService.getWebVaultUrl();
+        if (webVaultUrl == null) {
+            webVaultUrl = 'https://vault.bitwarden.com';
```
Hinton, line 71: I really dislike these defaults. While probably out of scope for this change, It's my opinion we should open a ticket to move those to the `environmentService` to keep things consistent.

common/src/misc/iframe_component.ts
```diff
@@ -60,6 +37,22 @@ export class WebAuthn {
         this.win.removeEventListener('message', this.parseFunction, false);
     }
 
+    protected createParams(data: any) {
+        return new URLSearchParams({
+            data: this.base64Encode(JSON.stringify(data)),
+            parent: encodeURIComponent(this.win.document.location.href),
+            v: '1',
```
Hinton, line 44: Feels like version should be a property set by the extending class since we probably won't bump version for both at the same time. (My understanding is that it can be used should we ever change the protocol while retaining backward compatibility with existing clients).

Hinton (changes requested):
One comment for future consideration. The `version` thing should probably be fixed now though.

common/src/misc/iframe_component.ts
MGibson1: I guess you're right, but the jslib version and web should be in lock step.

angular/src/components/login.component.ts
MGibson1: I've thought that before too, At the time when I looked into it there was a reason, but I can't remember it right now.

I think it was something along the lines of needing to override null values for development environments.

Hinton (approved):
Approved, just make sure we include backwards compatibility for older clients in the updated WebAuthn connector.

angular/src/components/login.component.ts
MGibson1: Ah yes, it's for Send: https://github.com/bitwarden/jslib/blob/master/angular/src/components/send/send.component.ts#L174-L177

Could just make a new helper for that too https://app.asana.com/0/1169444489336079/1200631805316488/f

PR #434, opened 2021-07-21
Feature/use hcaptcha on register if bot

MGibson1:
# Overview

Extract captcha component and request into separate abstract classes.

The lastest server PR bitwarden/server#1480 for captcha adds an attribute which responds with a `BadRequest` `ErrorResponse` if captcha is required. We need to not show an error on catpcha required, that will be handled in components. we also need to parse out captcha requirements from all validation issues.

# Files Changed
* **captchaProtected.component/login.component**: extract captcha processing stuff to an abstract component to extend.
* **register.component**: Extend captchaProtected component and initialize captcha on captcha required error responses to the register form submission. 
* **api-action.directive**: Do not display an error if the error is just a captcha required. Still log it, though.
* **captchaProtectedRequest/registerRequest/tokenRequest**: Extract and extend the required data for a captcha protected api endpoint.
* **errorResponse**: Extract out captchta error information from api `ErrorResponse`s for easier handling.


angular/src/components/register.component.ts
```diff
@@ -7,14 +8,17 @@ import { RegisterRequest } from 'jslib-common/models/request/registerRequest';
...
 import { AuthService } from 'jslib-common/abstractions/auth.service';
 import { CryptoService } from 'jslib-common/abstractions/crypto.service';
+import { EnvironmentService } from 'jslib-common/abstractions/environment.service';
 import { I18nService } from 'jslib-common/abstractions/i18n.service';
 import { PasswordGenerationService } from 'jslib-common/abstractions/passwordGeneration.service';
 import { PlatformUtilsService } from 'jslib-common/abstractions/platformUtils.service';
 import { StateService } from 'jslib-common/abstractions/state.service';
 
 import { KdfType } from 'jslib-common/enums/kdfType';
+import { CaptchaProtectedComponent } from './captchaProtectedComponent';
```
cscharf, line 18: Should be in a new import group?

common/src/models/request/registerRequest.ts
```diff
@@ -2,29 +2,18 @@ import { KeysRequest } from './keysRequest';
 import { ReferenceEventRequest } from './referenceEventRequest';
 
 import { KdfType } from '../../enums/kdfType';
+import { CaptchaProtectedRequest } from './captchaProtectedRequest';
```
cscharf, line 5: Should either be in a new import group or moved up above `KeysRequest` in the import group with the other request types

common/src/models/request/tokenRequest.ts
```diff
@@ -1,23 +1,20 @@
 import { TwoFactorProviderType } from '../../enums/twoFactorProviderType';
+import { CaptchaProtectedRequest } from './captchaProtectedRequest';
```
cscharf, line 2: Move down into the other import group below with `DeviceRequest`

common/src/models/response/errorResponse.ts
```diff
@@ -1,9 +1,12 @@
+import { Utils } from '../../misc/utils';
```
cscharf, line 1: Separate into 2 import groups

common/src/models/response/errorResponse.ts
```diff
@@ -20,6 +23,8 @@ export class ErrorResponse extends BaseResponse {
         if (errorModel) {
             this.message = this.getResponseProperty('Message', errorModel);
             this.validationErrors = this.getResponseProperty('ValidationErrors', errorModel);
+            this.captchaSiteKey = this.validationErrors?.HCaptcha_SiteKey?.[0];
```
cscharf, line 26: [Question] I know we abstracted the `HCaptcha_SiteKey` server-side, I wonder if there's a way to do the same here or not, or if the fact we're using HCaptcha unavoidable in the front-end anyway?

cscharf (commented):
Some minor style things and a question.

common/src/models/response/errorResponse.ts
MGibson1: I thought about that but ultimately decided that it'd be premature generalization.

If we really think we might use another provider I think the most prudent thing to do right now is set the `HCaptcha_SiteKey` key name in another property so we can sync it with what the server is requesting. Do you think there's value in that?

cscharf: Not at this time, no, thanks for the follow up!

cscharf (approved):
Looks good, thanks!

PR #435, opened 2021-07-22
Add helper methods to EnvironmentService for retrieving urls

Hinton:
## Objective
It was difficult to figure out which urls were used through the application, since there were no central place to store/retrieve them. To resolve this, I refactored the `EnvironmentService` to keep the state private, and instead expose helper functions for retrieving the appropriate urls.

### Noteworthy Code Changes
- **angular/src/components/lock.component.ts**: Special case where we convert `vault.bitwarden.com` to `bitwarden.com`.
- **common/src/services/api.service.ts**: Removed `setUrls`, and instead injects the `EnvironmentService` to determine the urls.
- **common/src/services/environment.service.ts**: Added helper functions, removed call to `apiService.setUrls`, Removed dependency on ApiService. No longer calls `notificationService.init`.
- **common/src/services/notifications.service.ts**: Changed to inject `EnvironmentService` as a regular dependency, and `init` no longer requires the service to be provided.

### Testing Considerations
We should verify all api calls, links, etc still works as expected in our different environments. As well as ensuring notification service gets initialized, since the logic no longer happens in `environmentService`.

Related to:
https://app.asana.com/0/1169444489336079/1200631805316488


angular/src/components/environment.component.ts
```diff
@@ -43,6 +47,8 @@ export class EnvironmentComponent {
             enterprise: this.enterpriseUrl,
         });
 
+        this.notificationService.init();
```
Hinton, line 50: I'm not really happy with this requirement. I could add a callback to `environmentService` which gets triggered on changes, or perhaps we should add `rxjs` as a dependency to `jslib-common` and let interested services subscribe to changes instead.

common/src/services/environment.service.ts
```diff
@@ -2,31 +2,112 @@ import { EnvironmentUrls } from '../models/domain/environmentUrls';
...
+        if (this.baseUrl) {
+            return this.baseUrl + '/api';
+        }
+
+        return 'https://api.bitwarden.com';
+    }
+
+    getIdentityUrl() {
+        if (this.identityUrl != null) {
+            return this.apiUrl;
```
cscharf, line 91: Shouldn't this return `this.identityUrl` here? I may be missing what's supposed to happen here if `this.identityUrl` is explicitly set.

common/src/services/environment.service.ts
```diff
@@ -2,31 +2,112 @@ import { EnvironmentUrls } from '../models/domain/environmentUrls';
...
+        if (this.baseUrl) {
+            return this.baseUrl + '/identity';
+        }
+
+        return 'https://identity.bitwarden.com';
+    }
+
+    getEventsUrl() {
+        if (this.eventsUrl != null) {
+            return this.apiUrl;
```
cscharf, line 103: Same here Re: `this.eventsUrl`?

common/src/services/environment.service.ts
```diff
@@ -90,22 +171,24 @@ export class EnvironmentService implements EnvironmentServiceAbstraction {
...
 
-        const envUrls = new EnvironmentUrls();
-        if (this.baseUrl) {
-            envUrls.base = this.baseUrl;
-        } else {
-            envUrls.api = this.apiUrl;
-            envUrls.identity = this.identityUrl;
-            envUrls.events = this.eventsUrl;
-        }
+        // TODO: Emit event to notify other services they should change urls?
```
cscharf, line 174: I really do like this idea (see comments in response to your earlier comment) now that I fully understand the scope of the changes and impact/needs around when those URLs change (through settings after logging out).

angular/src/components/environment.component.ts
cscharf: I know this setup is required to establish the appropriate SignalR connection for the client with the server so socket notifications can be received based on the `notificationsUrl`. I do wonder if there's a better way to do this outside of the environment component just knowing this is necessary, although if the user never configures their environment would this `init()` method ever get called? It feels like this needs to happen during client startup regardless and perhaps there's a better way to do that, I'm not sure.... is `setUrls` inside the `environmentService` called every time and/or when a user is logged in, etc.?

cscharf (commented):
Some feedback and thoughts provided.

angular/src/components/environment.component.ts
Hinton: It varies slightly between our clients. Web calls `setUrls` during the bootstraping phase to set `base` matching the origin. The other clients primarily calls `setUrlsFromStorage` and only uses `setUrls` in combination with `EnvironmentComponent`.

common/src/services/environment.service.ts
MGibson1: I might be slow in understanding this, but don't all the other services just grab the URL from this one every time?

angular/src/components/environment.component.ts
Hinton: If we add `rxjs` which is what Angular uses internally for their Observables to jslib-common, we could do something like `environmentService.urls.subscribe((urls) => { })` inside the `NotificationService`.

common/src/services/environment.service.ts
Hinton: They do, but `NotificationService` initializes a SignalR session, which would still call the old endpoint. Until it's been re-initialized hence the call on https://github.com/bitwarden/jslib/pull/435/files#diff-486e7d088c99f1fb43091643e50264a6272c5d14e3f572cfb04deb845130ded9R50

angular/src/components/environment.component.ts
cscharf: Yes, let's do that, I think that's a worthwhile dependency in this scenario and something we can likely start using elsewhere as well.

cscharf (dismissed):
Nice

cscharf (approved):

PR #437, opened 2021-07-23
[Provider] Refresh identity token on full sync

Hinton:
## Objective
Sometimes the `sync` response can be out of date with the identity token and contain organizations or providers which the user can't access yet. To resolve this I added a call to `refreshToken` while doing a full sync.

Note: I'm not certain if this will have side effects but from what I've managed to understand we usually refresh it every 5 min, so I don't see any harm in doing it more often sometimes.

### Testing Considerations
We should do some regression testing to verify the identity token can successfully be retrieved in all of our clients.

https://app.asana.com/0/1198901840263430/1200639990863109


cscharf (approved):
🤷🏻‍♂️ , seems like it should be fine.

PR #438, opened 2021-07-23
Use both instanceof and constructor name as type identification

MGibson1:
# Overview

`api-action.directive` does not display error messages on `captchaProtected` api endpoints requesting captcha verification. In the [browser](https://gitbhub.com/bitwarden/browser) project, the type of captcha-required responses was not being identified correctly. Adding a test to the constructor name as a fallback method.

### DIFF
--- a/angular/src/directives/api-action.directive.ts
+++ b/angular/src/directives/api-action.directive.ts
@@ -31,7 +31,7 @@ export class ApiActionDirective implements OnChanges {
         }, (e: any) => {
             this.el.nativeElement.loading = false;
 
-            if (e instanceof ErrorResponse && (e as ErrorResponse).captchaRequired) {
+            if ((e instanceof ErrorResponse || e.constructor.name === "ErrorResponse") && (e as ErrorResponse).captchaRequired) {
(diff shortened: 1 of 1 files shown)


angular/src/directives/api-action.directive.ts
```diff
@@ -31,7 +31,7 @@ export class ApiActionDirective implements OnChanges {
         }, (e: any) => {
             this.el.nativeElement.loading = false;
 
-            if (e instanceof ErrorResponse && (e as ErrorResponse).captchaRequired) {
+            if ((e instanceof ErrorResponse || e.constructor.name === 'ErrorResponse') && (e as ErrorResponse).captchaRequired) {
```
cscharf, line 34: I was wondering about this, but good catch!

cscharf (approved):

angular/src/directives/api-action.directive.ts
MGibson1: I just wish I knew when one would work over the other. Some light googling suggested ES3 or ES5 needing the constructor name, but browser (where the issue popped up) is targeting ES2016

PR #443, opened 2021-07-29
Update package lock for duo_web_sdk in angular

libertylocked:
Hi Bitwarden team!

I think recently @djsmith85 updated the `package.json` for jslib angular package (for #440), however it seems like `package-lock.json` wasn't updated.

I think it's best to save the integrity checks in the package-lock for such a crucial dependency. Please see diff.
(I double checked the checksum with `sha512sum duo_web_sdk-2.7.1.tgz | cut -f1 -d' ' | xxd -r -p | base64`)

### DIFF
--- a/angular/package-lock.json
+++ b/angular/package-lock.json
@@ -19,17 +19,20 @@
         "@angular/platform-browser-dynamic": "^11.2.11",
         "@angular/router": "^11.2.11",
         "@bitwarden/jslib-common": "file:../common",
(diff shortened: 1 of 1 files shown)


djsmith85 (approved):
My bad, forgot to add it in my commit. In your case (pulling building a separate package) this should suffice. Technically I would need to bump the jslib in all parent repos, for it to be included, but this will happen soon, with other changes anyway.

PR #451, opened 2021-08-10
Fix missing refresh method in ciphers.component

eliykat:
## Objective

I refactored the base `ciphers.component.ts` in #436, but was too aggressive in taking stuff out.

The `refresh` method is used by all clients, so I've added it back into `jslib`.

### DIFF
--- a/angular/src/components/ciphers.component.ts
+++ b/angular/src/components/ciphers.component.ts
@@ -42,6 +42,10 @@ export class CiphersComponent {
         await this.load(filter, deleted);
     }
 
+    async refresh() {
+        await this.reload(this.filter, this.deleted);
+    }
+
     async applyFilter(filter: (cipher: CipherView) => boolean = null) {
         this.filter = filter;
         await this.search(null);


angular/src/components/ciphers.component.ts
```diff
@@ -42,6 +42,10 @@ export class CiphersComponent {
         await this.load(filter, deleted);
     }
 
+    async refresh() {
```
MGibson1, line 45: should this have the same `try/finally` that was removed?

eliykat: yes, good point

eliykat: I don't think so, the `try/finally` was only to set `this.refreshing` in the `finally` block. `this.refreshing` is only used for paging and is not needed for virtual-scroll. (`web` has its own `refresh` method with the original logic, because it still uses paging, but other clients don't need it.)

MGibson1 (approved):

PR #452, opened 2021-08-10
Use custom virtual scroll strategy

eliykat:
## Objective

We've recently transitioned to using `cdk-virtual-scroll` for all cipher lists on desktop and browser.

`cdk-virtual-scroll` takes a "virtual scroll strategy", which provides the logic for rendering items. An important part of this is measuring the size of items in the list, so that the scrollbar can be correctly sized.

The default option is `fixedSizeVirtualScrollStrategy`, which takes an `itemSize` in `px` in the template, to tell it the height of the list items. That works fine for us in theory, because all our items are the same size, but any change to the actual rendered item height would require an update to the template. That could be caused by the runtime environment or by subsequent changes the code.

There is an `autoSizeVirtualScrollStrategy`, but that is still in `cdk-experimental` and is not rated for prod use. It's fairly complex because it doesn't assume that all items are the same height. But that's not our use case - we really just want an automatic way to set the fixed `itemSize`.

This PR creates a new scrolling strategy, `cipherListScrollStrategy`. It's basically a wrapper around `fixedSizeVirtualScrollStrategy`, but reads the actual rendered height of a sample list item to set the `itemSize`. This still assumes that all items are the same height, which I think is a valid assumption in our case. You can see the base class code [here](https://github.com/angular/components/blob/11.2.x/src/cdk/scrolling/fixed-size-virtual-scroll.ts).


angular/src/directives/CipherListVirtualScroll.directive.ts
```diff
@@ -0,0 +1,101 @@
...
+        const newItemSize = sampleItem?.offsetHeight;
+
+        if (newItemSize != null && newItemSize !== this.itemSize) {
+            this.itemSize = newItemSize;
+            this._scrollStrategy.updateItemAndBufferSize(this.itemSize, this.minBufferPx, this.maxBufferPx);
+        }
+    }
+
+    ngOnChanges() {
+        this._scrollStrategy.updateItemAndBufferSize(this.itemSize, this.minBufferPx, this.maxBufferPx);
```
MGibson1, line 95: Is this being re-calculated any time the list changes? wouldn't it be sufficient to just do it on load?

angular/src/directives/CipherListVirtualScroll.directive.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /** The scroll strategy used by this directive. */
+    _scrollStrategy = new CipherListVirtualScrollStrategy(this.itemSize, this.minBufferPx, this.maxBufferPx,
```
MGibson1, line 99: I think this should be at the top of the class

angular/src/directives/CipherListVirtualScroll.directive.ts
```diff
@@ -0,0 +1,101 @@
...
+    onContentRendered() {
+        if (this.timeout != null) {
+            clearTimeout(this.timeout);
+        }
+
+        this.timeout = setTimeout(this.checkItemSizeCallback, 500);
+    }
+}
+
+export function _cipherListVirtualScrollStrategyFactory(cipherListDir: CipherListVirtualScroll) {
```
MGibson1, line 37: does this need to be exported if we're just using it as a useFactory callback?

angular/src/directives/CipherListVirtualScroll.directive.ts
```diff
@@ -0,0 +1,101 @@
...
+    selector: 'cdk-virtual-scroll-viewport[defaultItemSize]',
+    providers: [{
+        provide: VIRTUAL_SCROLL_STRATEGY,
+        useFactory: _cipherListVirtualScrollStrategyFactory,
+        deps: [forwardRef(() => CipherListVirtualScroll)],
+    }],
+})
+export class CipherListVirtualScroll implements OnChanges {
+    // tslint:disable
+    static ngAcceptInputType_defaultItemSize: NumberInput;
```
MGibson1, line 51: does this just limit the types available to the scrolling strategy attributes? Isn't that already checked by typescript?

MGibson1 (commented):
This looks pretty streamlined. My questions all have to do with formatting or some fuzziness with angular

angular/src/directives/CipherListVirtualScroll.directive.ts
eliykat: Yes - angular throws an error otherwise. Not sure why to be honest, I agree with you in principle, but it seems like it's required

angular/src/directives/CipherListVirtualScroll.directive.ts
eliykat: This needs to be declared after `checkAndUpdateItemSize`. But I've now removed a lot of duplicate code here which may address your concerns

MGibson1 (approved):

MGibson1 (approved):

PR #453, opened 2021-08-11
Set up captcha iframe only on submit of login

MGibson1:
# overview

Fixes https://app.asana.com/0/0/1200768603643030/f.

This delay is necessary to make sure the location we're submitting the form to is the same as the webvault captcha iframe location.

This will be cherry picked to `rc` and all jslib clients will be updated to include these changes


cscharf (approved):

MGibson1:
🍒 ⛏️  to rc

PR #454, opened 2021-08-12
Use apikey client secret as captcha validation

MGibson1:
# Overview

Requires https://github.com/bitwarden/server/pull/1509

Handles captcha in login command as requiring a correct apikey client secret. This can be from environment variables in order to allow automated, un-monitored scripts a chance to still word despite captcha challenges.

# Files Changed
* **auth.service**: Add captcha token to `logInComplete` helper. This helper is used in CLI to allow pre-empting a 2fa.
* **login.command**: 
  * break out api client secret and client id 
  * if captcha is requires, request clientSecret
  * if another auth request fails, throw some instructions (**Note:** I've requested a help page for CLI captcha, the url will need updating)


MGibson1:
🍒 ⛏️  to rc

PR #455, opened 2021-08-12
Provide information to set webauthn allow in html template

MGibson1:
# Overview

Fixes: https://app.asana.com/0/0/1200719554428331/f

Chrome has begun refusing to load our webauthn iframe in browser. This appears to be due to either a race condition or an intentional block to adding an `allow` attribute to the iframe in javascript.

This PR provides a getter to provide the necessary allow directive to the `html` template.

# Files Changed
* **two-factor.ts**: Provide the necessary allow statement for webauthn
* **webauthn_iframe.ts**: only set allow if it's not already present.

# Upgrade path

I believe this requires no upgrade considerations. The `webauthn_iframe` component should be run on the client making the iframe and appending the `allow` attribute. This flow is complicated, though, so a second brain to validate, would be good.

### DIFF
--- a/angular/src/components/two-factor.component.ts
+++ b/angular/src/components/two-factor.component.ts
@@ -44,6 +44,10 @@ export class TwoFactorComponent implements OnInit, OnDestroy {
     onSuccessfulLogin: () => Promise<any>;
     onSuccessfulLoginNavigate: () => Promise<any>;
 
+    get webAuthnAllow(): string {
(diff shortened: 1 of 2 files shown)


djsmith85 (approved):
Looks good. Makes sense that the iframe isn't allowed to set policies itself

MGibson1:
🍒 ⛏️  to rc

PR #456, opened 2021-08-12
Generalize token refreshing to include reauth by api key

MGibson1:
# Overview

API Key-based authentication does not return a refresh token. Because of this, even after unlocking, it is not possible to work with your vault as syncing requires a refresh. Even if it didn't, auth tokens have a limited lifetime and should be valid until `bw logout` is called.

The fix is to generalize refresh to mean either refreshing via refresh token (the current method) OR refreshing via stored api key credentials (this PR).

# Files Changed
* **TokenService**: Add `clientId` and `clientSecret` to tokens being handled by the service. Treat them in the same way as `accessToken` and `refreshToken`.
* **api.service**: update referesh endpoint to `doAuthRefresh`, which determines the refresh method based on tokens stored in `tokenService`. Prefers refresh flow, but falls back to api-key based authentication to renew our access token. This is done through a callback provided  by the `NodeApiService`. If no callback is provided, error.
* **auth.service**: give clientIdClientSecret to token service for storage. These are only populated if api login is done.
* **nodeApi.service**: Set the api refresh callback. Only CLI is using api keys so this should be a good place to set to minimize changes in other repos.

# Testing

We'll need to validate login/sync/list/edit/lock/logout for both login methods (user+pass, api key)
* please verify that the keys `apikey_clientId` and `apikey_clientSecret` only appear in `data.json` if api key is used to log in.
* Please verify that the above keys are removed upon logout but **not** lock
  *  you will find `cat ~/Library/Application\ Support/Bitwarden\ CLI/data.json | grep client` helpful in the above tasks

### DIFF
--- a/common/src/services/auth.service.ts
+++ b/common/src/services/auth.service.ts
@@ -280,7 +280,7 @@ export class AuthService implements AuthServiceAbstraction {
 
         let emailPassword: string[] = [];
         let codeCodeVerifier: string[] = [];
-        let clientIdClientSecret: string[] = [];
+        let clientIdClientSecret: [string, string] = [null, null];
 
         if (email != null && hashedPassword != null) {
             emailPassword = [email, hashedPassword];
@@ -344,7 +344,7 @@ export class AuthService implements AuthServiceAbstraction {
             await this.tokenService.setTwoFactorToken(tokenResponse.twoFactorToken, email);
         }
 
-        await this.tokenService.setTokens(tokenResponse.accessToken, tokenResponse.refreshToken);
+        await this.tokenService.setTokens(tokenResponse.accessToken, tokenResponse.refreshToken, clientIdClientSecret);
         await this.userService.setInformation(this.tokenService.getUserId(), this.tokenService.getEmail(),
             tokenResponse.kdf, tokenResponse.kdfIterations);
         if (this.setCryptoKeys) {
--- a/common/src/abstractions/token.service.ts
+++ b/common/src/abstractions/token.service.ts
@@ -2,11 +2,15 @@ export abstract class TokenService {
     token: string;
     decodedToken: any;
     refreshToken: string;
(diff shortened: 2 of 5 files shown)


common/src/services/auth.service.ts
```diff
@@ -280,7 +280,7 @@ export class AuthService implements AuthServiceAbstraction {
 
         let emailPassword: string[] = [];
         let codeCodeVerifier: string[] = [];
-        let clientIdClientSecret: string[] = [];
+        let clientIdClientSecret: [string, string] = [null, null];
```
cscharf, line 283: Does this break the `TokenRequest` [constructor](https://github.com/bitwarden/jslib/blob/master/common/src/models/request/tokenRequest.ts#L18)? It seems like there are a few more down and upstream impacts/dependencies here on the clientId/secret structure, unless this type satisfies the token request server needs (an array that could be populated as `[null, null]` vs. `[]` or just `null`.

cscharf (commented):
Just one question

common/src/services/auth.service.ts
MGibson1: If you mean in terms of typescript, no. It's a more permissive to less permissive type.

If you mean in terms of communication with the server project. It shouldn't. All that I'm doing is removing `[]` from a possibility in your list, which isn't used as the code exists now. I just tested login out with `clientIdClientSecret = [null, null]` and everything worked as expected.

MGibson1: I'm fine either editing `tokenRequest` to expect a `[string, string]` or changing my code to take a `string[]` if you're worried about it.

cscharf: Nope, that's all good, thanks for the clarification and test/check!

cscharf (approved):

MGibson1:
cherry picked to `rc` (0a2ff12)

PR #461, opened 2021-08-20
Add null check to electronStorageService.Save

eliykat:
## Objective

Fix https://github.com/bitwarden/desktop/issues/1033. The latest desktop release cannot sync with older self-hosted servers. 

This is because the client is expecting a `ForcePasswordReset` property on the http sync response. Because that property doesn't exist and no other default value is provided, the client passes an `undefined` value to `electronStorageService`, which then throws an error when it tries to save.

## Code changes

In my brief testing, other Angular clients do not appear to be affected by this bug, so I assume they tolerate saving `null` or `undefined` values. Therefore, I've added a null check to `electronStorageService` to bring it into line with other storage services.

I've also added a default value, because we really should be doing that either way.

**Note:** this will need to be cherry-picked to `rc` and `desktop/rc` will need to be updated.

### DIFF
--- a/electron/src/services/electronStorage.service.ts
+++ b/electron/src/services/electronStorage.service.ts
@@ -50,10 +50,9 @@ export class ElectronStorageService implements StorageService {
             return Promise.resolve();
         }
         if (obj instanceof Set) {
-        	obj = Array.from(obj);
+            obj = Array.from(obj);
         }
         this.store.set(key, obj);
-        }
         return Promise.resolve();
     }
 
--- a/electron/src/services/electronStorage.service.ts
+++ b/electron/src/services/electronStorage.service.ts
@@ -46,10 +46,12 @@ export class ElectronStorageService implements StorageService {
     }
 
     save(key: string, obj: any): Promise<any> {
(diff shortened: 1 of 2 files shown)


electron/src/services/electronStorage.service.ts
```diff
@@ -46,10 +46,12 @@ export class ElectronStorageService implements StorageService {
     }
 
     save(key: string, obj: any): Promise<any> {
-        if (obj instanceof Set) {
-            obj = Array.from(obj);
+        if (obj != null) {
+            if (obj instanceof Set) {
+                obj = Array.from(obj);
+            }
+            this.store.set(key, obj);
```
Hinton, line 53: I kind of prefers to use "guard clauses" for these scenarios.

```suggestion
        if (obj == null) {
            return Promise.resolve();
        }
        if (obj instanceof Set) {
        	obj = Array.from(obj);
        }
        this.store.set(key, obj);
```

Hinton (dismissed):
Do we need to do the same change for the secure storage service?

vincentsalucci (approved):

PR #462, opened 2021-08-20
Revert WebAuthn iFrame handler

Hinton:
## Objective
Self-hosted users have trouble using WebAuthN with the latest clients, due to the clients expecting a newer connector protocol. Until we are confident self-hosted users have updated their servers we should revert the changes.

Note: This will be cherry picked to `rc`.


MGibson1 (approved):
Could you please create a tech debt ticket to re-revert this 1-2 releases down the line?

PR #465, opened 2021-08-23
Add BasePolicy and policies to policy service

Hinton:
## Objective
In an effort to simplify the current policy system, I've introduced a way for the `PolicyService` to keep track of possible policies.

### DIFF
--- a/common/src/abstractions/policy.service.ts
+++ b/common/src/abstractions/policy.service.ts
@@ -1,6 +1,7 @@
 import { PolicyData } from '../models/data/policyData';
 
 import { MasterPasswordPolicyOptions } from '../models/domain/masterPasswordPolicyOptions';
+import { Organization } from '../models/domain/organization';
 import { Policy } from '../models/domain/policy';
 import { ResetPasswordPolicyOptions } from '../models/domain/resetPasswordPolicyOptions';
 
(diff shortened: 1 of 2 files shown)


common/src/abstractions/policy.service.ts
```diff
@@ -9,8 +10,20 @@ import { PolicyType } from '../enums/policyType';
...
+    abstract name: string;
+    abstract description: string;
+    abstract type: PolicyType;
+    abstract component: any;
+
+    display(organization: Organization) {
+        return true;
+    }
+}
+
```
addisonbeck, line 23: We should put this in a dedicated file somewhere in /models

Hinton: It doesn't quite fit into the `models` folder, since it's an `abstract` class which is expected to be extended, and everything else in `models` are pure data storage. At the same time I also agree it doesn't quite fit into the `PolicyService`.

We are only using it in the web vault currently, so perhaps it would make sense to extend the `PolicyService` there and let it live in web.

addisonbeck: I see three decent paths forward here:
1. There is value in moving this to `web` since that is the only place that is going to be using this model. 
2. I could also see something like `/models/abstractions` being something we are going to want in the future anyway, maybe even with moving the current `abstractions` folder into `services` to keep service and model abstractions separate. 
3. Or just simply `/models/domain/policies.ts` that holds BasePolicy and anything that extends it. 

I'll leave that up to you, as ultimately I think all of these are good alternatives. Maybe even some other opinions from @bitwarden/dept-engineering if anybody has one.

Hinton:
Moved the logic to web.

PR #467, opened 2021-08-27
[Account Switching] Introduce AccountService 

addisonbeck:
### Related Work
[Asana Task](https://app.asana.com/0/1200578627114106/1200520645190695/f)
[Sibling task for details on services impact](https://app.asana.com/0/1200578627114106/1200520645190694/f)

### Changes
* Relocated the StorageServiceOptions model and renamed it SettingStorageOptions
    * Added a few flags here for storage options to skip memory, disk, or use secure storage.
* Created an AccountStorageKey enum for storing disparate keys from across services, components, and constants files. What you see here is a copy/paste of everything I found across jslib and desktop. 
    * I think there may be benefit in breaking these down into better scoped enums, and 
    * There is probably a better name to give this enum since not every key is really related to an "account".
* Created an `Account` object for in memory setting storage and noting when an account is active.
   * Current I have moved an implementation of `isAuthenticated()` into this class, but that may not stick around.
* Introduced an AccountService that ties all these other components together and exposes methods for managing setting storage and active account.

### Notes
* This PR is a draft, and I would recommend we don't merge it. There is still some discovery going on here and I think merging this now is just going to create a confusing diff later. This PR exists to gather feedback for the subtask linked above. This PR also contains breaking changes that I am not ready to fix in clients yet, so merging it and then updating clients would break them.


cscharf:
@addisonbeck , you made tslint angry:

```
common/src/abstractions/account.service.ts:1:35
ERROR: 1:35    quotemark               " should be '
ERROR: 3:25    quotemark               " should be '
ERROR: 4:39    quotemark               " should be '

common/src/abstractions/storage.service.ts:1:39
ERROR: 1:39    quotemark               " should be '

common/src/enums/accountStorageKey.ts:19:29
ERROR: 19:29   no-trailing-whitespace  trailing whitespace

common/src/models/domain/account.ts:1:35
ERROR: 1:35    quotemark               " should be '
ERROR: 2:25    quotemark               " should be '

common/src/services/account.service.ts:20:44
ERROR: 20:44   triple-equals           != should be !==
ERROR: 23:1    no-trailing-whitespace  trailing whitespace
ERROR: 27:6    semicolon               Unnecessary semicolon
ERROR: 53:56   no-trailing-whitespace  trailing whitespace
ERROR: 118:11  semicolon               Missing semicolon
```

common/src/services/account.service.ts
```diff
@@ -0,0 +1,124 @@
+import { AccountService as AccountServiceAbstraction } from '../abstractions/account.service';
+import { StorageService } from '../abstractions/storage.service';
+import { AccountStorageKey } from '../enums/accountStorageKey';
+
+import { Account } from '../models/domain/account';
+import { SettingStorageOptions } from '../models/domain/settingStorageOptions';
+
+export class AccountService implements AccountServiceAbstraction {
+    private accounts: Account[] = [];
+    private activeUserId: string;
```
cscharf, line 10: Why not just use an `[index]` (`int`) for `accounts`, and reference that way. Active Account, which seems like it is going to be called a LOT will reduce the cycles necessary having to use `find()` when it can instead use `[index]` to reference the appropriate account.

OR, use the `userId` as the indexer and make `accounts` a dynamic object graph where the `userId` is the property name of a property of type `Account`? This completely eliminates the use of `find()` as a pattern which I see in many of the methods here, reducing CPU overhead and being a little more straightforward. How that works in TypeScript, 🤷🏻 , but that's the pattern that may work better and then you can keep the `activeUserId` property as-is.

cscharf (commented):
Just a few bits of additional feedback. More left in Asana

common/src/services/account.service.ts
addisonbeck: I really appreciate this feedback, another thing I had not even begun to notice. I changed this up to use a [`Record`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype), removing all the `find()`s and indexing using `userId` instead.

addisonbeck:
Closing this as I get ready to open a PR for the next phase

PR #469, opened 2021-08-30
Smart Passwords - Using website annotations for password policies

mikibakaiki:
Hey everyone! 

As mentioned in [this post](https://community.bitwarden.com/t/compliant-random-passwords/28036), I set out to develop HTML annotations to describe password policies that could be easily read and applied to the password generator of Bitwarden. 

This PR includes the base implementation of this feature. 

It uses the same policy annotations - aka password rules -  as [Apple](https://developer.apple.com/password-rules/). 
It also uses an npm package - [@passcert/pwrules-annotations](https://www.npmjs.com/package/@passcert/pwrules-annotations) - that I developed and adapted to typescript, based on [Apple's own parser](https://github.com/apple/password-manager-resources).

There is a new option on the dropdown - **_Smart Password_** - that will try to read the field `passwordrules` in an input form. This field contains the password policies specified by the website, making it easier for the generator to generate compliant passwords.

bw_aug

In this example, the site had the following rules:

`passwordrules="required: upper; required: lower; required:digit; required: special; minlength: 10;"` 

So, the password must have at least 10 characters, and at least one lowercase letter, one uppercase letter, one digit, and one symbol. Since all character classes are required, the checkboxes are disabled and the minimum number of digits and symbols is also disabled because changing these values would generate a password that would be uncompliant. The minimum length of the password is also capped at 10, for the same reason.

You can read more about the [password policy generation options here](https://developer.apple.com/documentation/security/password_autofill/customizing_password_autofill_rules)

**TODO:**
- Get the list used in [apple's password quirks](https://github.com/apple/password-manager-resources/blob/main/quirks/password-rules.json) if a site does not have a `passwordrules` attribute.
- Overall checks and possible optimizations
- Test against company policies

Hoping to get some feedback and more ways to test this feature, which I think could be a great addition to Bitwarden. 😄


Hinton:
Hi,

PR #470, opened 2021-08-31
Fix 'Cannot access variable before initialization' error in Desktop

eliykat:
## Objective

Fix the following error, which occurs whenever you load the 2FA screen on Desktop:

Screen Shot 2021-08-20 at 4 14 59 pm

This is because we call `queryParamsSub.unsubscribe()` from within the `subscribe()` callback. I'm not sure why this doesn't work - we use this pattern elsewhere and it works fine, it just throws this error in this specific component. 

## Code changes

* `two-factor.component.ts` - use the rxjs `first` operator to get the first emitted value. This makes unsubscribing unnecessary and (in my opinion) is the better pattern to use here.

The original pattern appears in other components but they're not causing any trouble, so I haven't changed them at this stage.

### DIFF
--- a/angular/src/components/two-factor.component.ts
+++ b/angular/src/components/two-factor.component.ts
@@ -5,6 +5,8 @@ import {
     Router,
 } from '@angular/router';
 
+import { first } from 'rxjs/operators';
+
 import { TwoFactorProviderType } from 'jslib-common/enums/twoFactorProviderType';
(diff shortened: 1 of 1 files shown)


eliykat:
Tech debt ticket created: https://app.asana.com/0/1200578627114106/1200894245876095

PR #472, opened 2021-08-31
[SSO] Set password auto enroll update

vincentsalucci:
## Code Changes
- **set-password.component.ts**: Added new conditional that checks whether or not the user needs to be auto-enrolled in reset password.
- **user.service.ts**: Added new method that will retrieve an organization via `identifier`

### DIFF
--- a/angular/src/components/set-password.component.ts
+++ b/angular/src/components/set-password.component.ts
@@ -18,12 +18,16 @@ import { EncString } from 'jslib-common/models/domain/encString';
 import { SymmetricCryptoKey } from 'jslib-common/models/domain/symmetricCryptoKey';
 
 import { KeysRequest } from 'jslib-common/models/request/keysRequest';
+import { OrganizationUserResetPasswordEnrollmentRequest } from 'jslib-common/models/request/organizationUserResetPasswordEnrollmentRequest';
 import { SetPasswordRequest } from 'jslib-common/models/request/setPasswordRequest';
 
 import { ChangePasswordComponent as BaseChangePasswordComponent } from './change-password.component';
 
 import { HashPurpose } from 'jslib-common/enums/hashPurpose';
 import { KdfType } from 'jslib-common/enums/kdfType';
+import { PolicyType } from 'jslib-common/enums/policyType';
+
+import { Utils } from 'jslib-common/misc/utils';
 
 @Directive()
 export class SetPasswordComponent extends BaseChangePasswordComponent {
@@ -31,6 +35,8 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
     showPassword: boolean = false;
     hint: string = '';
     identifier: string = null;
+    orgId: string;
+    resetPassswordAutoEnroll = false;
 
     onSuccessfulChangePassword: () => Promise<any>;
     successRoute = 'vault';
@@ -57,6 +63,13 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
             }
         });
 
+        // Automatic Enrollment Detection
+        const org = await this.userService.getOrganizationByIdentifier(this.identifier);
+        this.orgId = org?.id;
+        const policyList = await this.policyService.getAll(PolicyType.ResetPassword);
+        const policyResult = this.policyService.getResetPasswordPolicyOptions(policyList, this.orgId);
+        this.resetPassswordAutoEnroll = policyResult[1] && policyResult[0].autoEnrollEnabled;
+
         super.ngOnInit();
     }
 
@@ -69,6 +82,7 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
 
     async performSubmitActions(masterPasswordHash: string, key: SymmetricCryptoKey,
         encKey: [SymmetricCryptoKey, EncString]) {
+        // Create Set Password Request
         const request = new SetPasswordRequest();
         request.masterPasswordHash = masterPasswordHash;
         request.key = encKey[1].encryptedString;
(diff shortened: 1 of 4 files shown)


angular/src/components/set-password.component.ts
```diff
@@ -57,6 +63,13 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
             }
         });
 
+        // Automatic Enrollment Detection
+        const org = await this.userService.getOrganizationByIdentifier(this.identifier);
+        this.orgId = org?.id;
+        const policyList = await this.policyService.getAll(PolicyType.ResetPassword);
+        const policyResult = this.policyService.getResetPasswordPolicyOptions(policyList, this.orgId);
+        this.resetPasswordAutoEnroll = policyResult[1] && policyResult[0].autoEnrollEnabled;
```
addisonbeck, line 71: `identifier` will still be null for non-SSO flows, right? If that is true, we should wrap this in a conditional to make sure we don't check a bunch of stuff we don't need to.

angular/src/components/set-password.component.ts
```diff
@@ -69,30 +82,51 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
 
     async performSubmitActions(masterPasswordHash: string, key: SymmetricCryptoKey,
         encKey: [SymmetricCryptoKey, EncString]) {
+        // Create Set Password Request
```
addisonbeck, line 85: I would say this makes for a better function than comment. We should create a constructor for SetPasswordRequest.

angular/src/components/set-password.component.ts
```diff
@@ -69,30 +82,51 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
...
         const request = new SetPasswordRequest();
         request.masterPasswordHash = masterPasswordHash;
         request.key = encKey[1].encryptedString;
         request.masterPasswordHint = this.hint;
         request.kdf = this.kdf;
         request.kdfIterations = this.kdfIterations;
         request.orgIdentifier = this.identifier;
-
+        // Create Key Pair
```
addisonbeck, line 93: Same, this comment's existence is indicative of an issue with the function. I would include keys in the above mentioned request constructor.

angular/src/components/set-password.component.ts
```diff
@@ -69,30 +82,51 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
...
+                    // RSA Encrypt user's encKey.key with organization public key
+                    const userEncKey = await this.cryptoService.getEncKey();
+                    const encryptedKey = await this.cryptoService.rsaEncrypt(userEncKey.key, publicKey.buffer);
+
+                    // Create request and execute enrollment
+                    const resetRequest = new OrganizationUserResetPasswordEnrollmentRequest();
+                    resetRequest.resetPasswordKey = encryptedKey.encryptedString;
+
+                    // Get User Id
+                    const userId = await this.userService.getUserId();
```
addisonbeck, line 119: Most of the rest of these I would just remove, since they describe what the code is doing more than why it is doing it. Feel free to extract anything else that makes sense and improves readability, but we shouldn't write comments like these in my opinion. To quote Uncle Bob:

> The proper use of comments is to compensate for our failure to express ourself in code. Note that I used the word failure. I meant it. Comments are always failures.

(I think he goes a little too crazy with it here but share the sentiment)

angular/src/components/set-password.component.ts
```diff
@@ -108,4 +142,17 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
...
+            this.kdf, this.kdfIterations);
+        await this.cryptoService.setKey(key);
+        await this.cryptoService.setEncKey(encKey[1].encryptedString);
+        await this.cryptoService.setEncPrivateKey(keys[1].encryptedString);
+
+        const localKeyHash = await this.cryptoService.hashPassword(this.masterPassword, key,
+            HashPurpose.LocalAuthorization);
+        await this.cryptoService.setKeyHash(localKeyHash);
+
+    }
```
addisonbeck, line 157: 💯

angular/src/components/set-password.component.ts
```diff
@@ -18,19 +18,25 @@ import { EncString } from 'jslib-common/models/domain/encString';
...
+import { Utils } from 'jslib-common/misc/utils';
 
 @Directive()
 export class SetPasswordComponent extends BaseChangePasswordComponent {
     syncLoading: boolean = true;
     showPassword: boolean = false;
     hint: string = '';
     identifier: string = null;
+    orgId: string;
+    resetPasswordAutoEnroll = false;
```
addisonbeck, line 39: Instead of having a class level variable for this set onInit we should wrap that init logic into a function that returns orgId if true, and just call when needed before making API calls.

addisonbeck (commented):
Great work Vince, just a few comments.

angular/src/components/set-password.component.ts
addisonbeck: Oh we should dump that blank line though

angular/src/components/set-password.component.ts
vincentsalucci: At this time, I don't believe there is a non-SSO scenario, but you make a great point about defensive programming - will update.

angular/src/components/set-password.component.ts
vincentsalucci: Hmm. Could go either way with this one. I'm not sure how I feel about the ambiguity of `orgId != null` instead of an explicit `boolean`. The boolean provided supplies the `ngIf` conditional for all downstream UI implementations (at this moment). Will think on this!

addisonbeck: Ah, okay, I was missing the context that these values were used in the UI (yay jslib). It makes sense to leave this as is.

angular/src/components/set-password.component.ts
addisonbeck: Might be fine as is. I was thinking this component had wider scope. Thanks Vince!

vincentsalucci: Updated with null check.

angular/src/components/set-password.component.ts
vincentsalucci: Created constructor.

angular/src/components/set-password.component.ts
vincentsalucci: Done!

angular/src/components/set-password.component.ts
vincentsalucci: Removed comments except for the one explaining the crypto operation.

cscharf (approved):

PR #474, opened 2021-08-31
Vault Timeout Policy

Hinton:
## Objective
Implements a *Vault Timeout Policy* which lets organizations restrict the maximum vault timeout. It also adds the new `VaultTimeoutInputComponent` which supports `Custom` timeouts.

The `VaultTimeoutInputComponent` is implemented using the `ControlValueAccessor` and `Validator` which provides a nice abstraction layer to encapsulate the complex logic of *Custom* timeouts. The bound model returns a nullable numeric value equal to the timeout in minutes. If the timeout exceeds the allowed timeout the `Validator` will trigger which sets the status to `valid = false`.

I also removed the deprecated `lockTimeout` from `PlatformUtilsService`.


common/src/services/vaultTimeout.service.ts
```diff
@@ -141,6 +138,25 @@ export class VaultTimeoutService implements VaultTimeoutServiceAbstraction {
...
+
+        if (await this.policyService.policyAppliesToUser(PolicyType.MaximumVaultTimeout)) {
+            const policy = await this.policyService.getAll(PolicyType.MaximumVaultTimeout);
+            // Remove negative values, and ensure it's smaller than maximum allowed value according to policy
+            const timeout = Math.max(0, Math.min(vaultTimeout, policy[0].data.minutes));
+
+            // We really shouldn't need to set the value here, but multiple services relies on this value being correct.
+            if (vaultTimeout != timeout) {
+                await this.storageService.save(ConstantsService.vaultTimeoutKey, timeout);
+            }
```
Hinton, line 152: I really don't like this, but the dependency chain for `VaultTimeoutService` would need to be reversed to clean this up. Which feels out of scope for this issue.

MGibson1 (approved):
This looks good to me

PR #479, opened 2021-09-08
Revert "Vault Timeout Policy"

Hinton:
Reverts bitwarden/jslib#474 since we're holding of merging.


MGibson1 (approved):
Not sure you actually need to revert. Let's hold on this till we find out if we're really frozen

PR #480, opened 2021-09-08
Add MaximumVaultTimeout policy type

Hinton:
## Objective
Implements a *Vault Timeout Policy* which lets organizations restrict the maximum vault timeout. It also adds the new `VaultTimeoutInputComponent` which supports `Custom` timeouts.

The `VaultTimeoutInputComponent` is implemented using the `ControlValueAccessor` and `Validator` which provides a nice abstraction layer to encapsulate the complex logic of *Custom* timeouts. The bound model returns a nullable numeric value equal to the timeout in minutes. If the timeout exceeds the allowed timeout the `Validator` will trigger which sets the status to `valid = false`.

I also removed the deprecated `lockTimeout` from `PlatformUtilsService`.

(Identical to #474 which got reverted by #479)


MGibson1 (approved):
same as last

PR #481, opened 2021-09-09
[Reset Password v1] Refactor ForcePasswordReset into AuthResult

vincentsalucci:
## Code Changes
- **login.component**: Added conditional/options for navigation when forcing a password reset is necessary
- **sso.component**:  Added conditional/options for navigation when forcing a password reset is necessary
- **two-factor.component**: Added conditional for when forcing a password reset is necessary
- **authResult**: Added `forcePasswordReset` boolean
- **identityTokenResponse**: Added `forcePasswordReset` boolean
- **auth.service**: Added `forcePasswordReset` to the `AuthResult` response object

### DIFF
--- a/angular/src/components/sso.component.ts
+++ b/angular/src/components/sso.component.ts
@@ -31,10 +31,12 @@ export class SsoComponent {
     onSuccessfulLoginNavigate: () => Promise<any>;
     onSuccessfulLoginTwoFactorNavigate: () => Promise<any>;
     onSuccessfulLoginChangePasswordNavigate: () => Promise<any>;
+    onSuccessfulLoginForceResetNavigate: () => Promise<any>;
 
     protected twoFactorRoute = '2fa';
     protected successRoute = 'lock';
     protected changePasswordRoute = 'set-password';
+    protected forcePasswordResetRoute = 'update-temp-password';
     protected clientId: string;
     protected redirectUri: string;
     protected state: string;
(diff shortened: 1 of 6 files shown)


angular/src/components/sso.component.ts
```diff
@@ -161,6 +163,12 @@ export class SsoComponent {
                         },
                     });
                 }
+            } else if (response.forcePasswordReset) {
```
Hinton, line 166: We are starting to have very many branches in this function. Which makes it a bit tricky to follow.

One thing that comes to mind that would clean it up a fair bit would be to convert the `on<hook>` to actual methods with a default implementation.

I.e. 

```ts
protected onSuccessfulLoginForceResetNavigate(orgIdFromState) {
    this.router.navigate([this.changePasswordRoute], {
        queryParams: {
            identifier: orgIdFromState,
        },
    });
}
```



 We could simplify things by using proper functions for the `onSuccessfulLoginForceResetNavigate` which by default does `this.router.navigate([this.forcePasswordResetRoute]);`.

We could then look into using guard clauses.

```ts
this.loggingIn = true;
try {
    this.formPromise = this.authService.logInSso(code, codeVerifier, this.redirectUri);
    const response = await this.formPromise;
    
    if (response.twoFactor) {
        this.onSuccessfulLoginTwoFactorNavigate();
        return;
    }

    ...
}
```

Hinton (approved):
In general it looks good, I wonder if we should take the opportunity to clean up the `sso.component` a bit, to avoid having nested if cases.

angular/src/components/sso.component.ts
vincentsalucci: Noted - will tackle some of this cleanup in another group of PRs.

vincentsalucci:
@Hinton - Sorry for the dismissal, the branch was getting a little stale (my fault).

Hinton (approved):

angular/src/components/sso.component.ts
Hinton: Feel free to make a tech debt item and throw it on Asana if you don't have time atm.

PR #482, opened 2021-09-09
Disable Private Vault Export Policy

Hinton:
## Objective
Adds support for the new policy which disables personal vault export.

Asana: https://app.asana.com/0/1198901840263430/1200900473757545

### DIFF
--- a/angular/src/components/export.component.ts
+++ b/angular/src/components/export.component.ts
@@ -23,7 +23,7 @@ export class ExportComponent implements OnInit {
     masterPassword: string;
     format: 'json' | 'encrypted_json' | 'csv' = 'json';
     showPassword = false;
-    disablePrivateVaultPolicyEnabled: boolean = false;
+    disabledByPolicy: boolean = false;
 
     constructor(protected cryptoService: CryptoService, protected i18nService: I18nService,
         protected platformUtilsService: PlatformUtilsService, protected exportService: ExportService,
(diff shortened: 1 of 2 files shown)


angular/src/components/export.component.ts
```diff
@@ -9,28 +10,43 @@ import { EventService } from 'jslib-common/abstractions/event.service';
...
 @Directive()
-export class ExportComponent {
+export class ExportComponent implements OnInit {
     @Output() onSaved = new EventEmitter();
 
     formPromise: Promise<string>;
     masterPassword: string;
     format: 'json' | 'encrypted_json' | 'csv' = 'json';
     showPassword = false;
+    disablePrivateVaultPolicyEnabled: boolean = false;
```
djsmith85, line 26: Looks good, I'd only pick another name for the flag. Maybe more like the policy itself. Disable and Enable in one word is kinda odd and mixing personal vs private
disablePersonalVaultExport ?

djsmith85 (changes requested):

angular/src/components/export.component.ts
MGibson1: maybe just `exportDisabledByPolicy` doesn't really matter where it comes from

Hinton: Renamed to `disabledByPolicy` since export is implied from the component.

MGibson1 (approved):

PR #484, opened 2021-09-09
Allow renaming by adding previous names field

MGibson1:
# Overview

Currently, there is a hurdle to jump when renaming messages due to delay in their translation. However, if it's a simple rename, the previous translation should still exist.

This adds optional information to a message enumerating the previous names of a message, most recent first. When translating, if the actual ID isn't found, we'll use these renames to help locate a translated previous version, if it exists.

This field should only be used when changing the name of a message without substantively changing its content.

## Open Questions

I'm not 100% sure if crowdin will remove unused translations when generating new locales for us. It's probably best to keep old message IDs in our `messages.json` until crowdin has a chance to catch up to the rename.

### DIFF
--- a/common/src/misc/utils.ts
+++ b/common/src/misc/utils.ts
@@ -292,6 +292,10 @@ export class Utils {
         return str == null || typeof str !== 'string' || str.trim() === '';
     }
 
+    static isNullOrEmpty(str: string): boolean {
+        return str == null || typeof str !== 'string' || str === '';
+    }
+
     static nameOf<T>(name: string & keyof T) {
         return name;
     }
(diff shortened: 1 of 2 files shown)


Hinton:
@MGibson1 After thinking about this for a while, I'm wondering if this will actually work. The localized files from Crowdin will always contain all keys from the English source of truth, just not translated.

MGibson1:
We'll need to update our crowdin settings to not copy untranslated strings.

We will also need to make changes to bitwarden/browser to use our i18n service rather than Chrome's.

I've added the hold label because for now it's really not very high priority, but will be useful when we because appropriately annoyed at this lack in freedom.

PR #485, opened 2021-09-10
[Account Switching] Allow clients to store data for more than one user

addisonbeck:
### Code Changes
* See commits & descriptions 😉


addisonbeck:
I found some issues with this that need to be addressed and will reopen soon

PR #487, opened 2021-09-14
Organization autoscaling

MGibson1:
# Overview

Asana Ticket: https://app.asana.com/0/0/1200970162914676/f

Service updates for organization autoscaling. See bitwarden/server#1585

### DIFF
--- a/common/src/abstractions/api.service.ts
+++ b/common/src/abstractions/api.service.ts
@@ -33,6 +33,7 @@ import { KeysRequest } from '../models/request/keysRequest';
 import { OrganizationCreateRequest } from '../models/request/organizationCreateRequest';
 import { OrganizationImportRequest } from '../models/request/organizationImportRequest';
 import { OrganizationKeysRequest } from '../models/request/organizationKeysRequest';
+import { OrganizationSubscriptionUpdateRequest } from '../models/request/organizationSubscriptionUpdateRequest';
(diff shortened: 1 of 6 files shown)


common/src/abstractions/api.service.ts
```diff
@@ -379,6 +380,7 @@ export abstract class ApiService {
     postOrganizationApiKey: (id: string, request: PasswordVerificationRequest) => Promise<ApiKeyResponse>;
     postOrganizationRotateApiKey: (id: string, request: PasswordVerificationRequest) => Promise<ApiKeyResponse>;
     postOrganizationUpgrade: (id: string, request: OrganizationUpgradeRequest) => Promise<PaymentResponse>;
+    postOrganizationUpdateSubscription: (id: string, request: OrganizationSubscriptionUpdateRequest) => Promise<void>;
```
Hinton, line 383: Any particular reason for using `void` here? I think we usually use `any` in the `ApiService`.

Hinton (changes requested):
Minor comment.

common/src/abstractions/api.service.ts
MGibson1: The reason is better type safety. 

Requests that do not have a return value should return void to indicate that fact. The way the `ApiService.send` method works is it returns a response only if `hasResponse == true`, otherwise it will either return `void` or `throw`. This way, no component using this api method could mistakenly think it's getting a response object.

Hinton (approved):

PR #489, opened 2021-09-16
Move custom fields to separate components

eliykat:
## Objective
@MGibson1 suggested in https://github.com/bitwarden/browser/pull/1963 that we move custom fields out of the `add-edit.component.ts` and into their own component. This tidies up `add-edit.component`, which already has a lot going on, and will also make the linked custom fields implementation cleaner. I'll rebase bitwarden/browser#1963 on top of this once merged.

There's less logic in `view.component.ts`, but I've split that out as well for consistency.

Note: this is a pure refactor, there are no changes to logic or behaviour in this PR.

## Code changes
* move relevant code and html from `add-edit.component` to new `add-edit-custom-field.component`
* move relevant code and html from `view.component` to new `view-custom-field.component`

Client changes:
* browser: https://github.com/bitwarden/browser/pull/2072
* web: https://github.com/bitwarden/web/pull/1192
* desktop: https://github.com/bitwarden/desktop/pull/1076

### DIFF
--- a/angular/src/components/add-edit-custom-fields.component.ts
+++ b/angular/src/components/add-edit-custom-fields.component.ts
@@ -25,6 +25,7 @@ import { FieldType } from 'jslib-common/enums/fieldType';
 @Directive()
 export class AddEditCustomFieldsComponent implements OnChanges {
     @Input() cipher: CipherView;
+    @Input() thisCipherType: CipherType;
     @Input() editMode: boolean;
 
     addFieldType: FieldType = FieldType.Text;
@@ -46,7 +47,7 @@ export class AddEditCustomFieldsComponent implements OnChanges {
     }
 
     ngOnChanges(changes: SimpleChanges) {
-        if (changes.cipher.currentValue != changes.cipher.previousValue) {
+        if (changes.thisCipherType.currentValue !== changes.thisCipherType.previousValue) {
             this.setLinkedFieldOptions();
         }
     }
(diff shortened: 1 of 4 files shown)


Hinton (commented):
👍, great to move away things from the quite bloated `add-edit` component.

I've got a couple of personal thoughts on this refactor since I recently worked with reactive forms for the custom vault timeout component. *Please feel free to ignore the below thoughts.*

Currently it looks like the `AddEditCustomFieldsComponent` depends heavily on `cipher` and directly manipulates it's fields. This could potentially be a bit fragile since `AddEditCipherComponent` no longer has sole ownership of that data.

In `VaultTimeoutInputComponent` I used Reactive Forms to combine internal fields to a single output. I suspect we could do something similar here with `AddEditCustomFieldsComponent` exposing something like:
```ts
fields = new FormArray([
  new FormGroup({
    fieldType: FormControl(FieldType.Text),
    value: FormControl(null),
    visible: FormControl(false),
  }),
]);
```
The `AddEditCipherComponent` would then listen to these field using the `[formControl]` binding and be responsible for converting from the internal data model back to `cipher`. Efficiently reversing the relationship.

https://angular.io/guide/reactive-forms#creating-nested-form-groups

eliykat:
@Hinton I think this is a good idea, however I couldn't find a nice way of implementing it without overhauling the entire `add-edit` form, which I don't want to do right now. For now I think let's keep the current structure unless it's actively causing problems.

Hinton (approved):

PR #497, opened 2021-09-23
Add theme enums and platformUtilsService helper

eliykat:
## Objective

Refactor how we handle themes in jslib, to assist the [dark mode PR](https://github.com/bitwarden/web/pull/1017).

## Code changes
* use an enum for themes, so that we don't have to keep typing out the string literals everywhere. This must be a string enum because its value is inserted into the `<html>` class in order to set the theme.
* add a new helper method, `getEffectiveTheme`. This reduces repetition where we need to (a) check the system theme if required, and (b) decide what theme is the default for that client. (Web will default to light theme, other clients default to the system theme.)

This should be reviewed in conjunction with the dark mode PR. If acceptable I will submit refactors for other clients to use the new enum type and helper method.


cscharf (approved):

common/src/abstractions/platformUtils.service.ts
```diff
@@ -28,7 +29,8 @@ export abstract class PlatformUtilsService {
     readFromClipboard: (options?: any) => Promise<string>;
     supportsBiometric: () => Promise<boolean>;
     authenticateBiometric: () => Promise<boolean>;
-    getDefaultSystemTheme: () => Promise<'light' | 'dark'>;
-    onDefaultSystemThemeChange: (callback: ((theme: 'light' | 'dark') => unknown)) => unknown;
+    getDefaultSystemTheme: () => Promise<ThemeType.Light | ThemeType.Dark>;
+    onDefaultSystemThemeChange: (callback: ((theme: ThemeType.Light | ThemeType.Dark) => unknown)) => unknown;
+    getEffectiveTheme: () => Promise<ThemeType>;
```
Hinton, line 34: We'll need to implement this for the `electronPlatformUtils` and `CliPlatformUtils` in the jslib project.

Hinton (changes requested):

common/src/abstractions/platformUtils.service.ts
eliykat: Good catch. Done

electron/src/services/electronPlatformUtils.service.ts
```diff
@@ -192,8 +195,17 @@ export class ElectronPlatformUtilsService implements PlatformUtilsService {
         return ipcRenderer.invoke('systemTheme');
     }
 
-    onDefaultSystemThemeChange(callback: ((theme: 'light' | 'dark') => unknown)) {
-        ipcRenderer.on('systemThemeUpdated', (event, theme: 'light' | 'dark') => callback(theme));
+    onDefaultSystemThemeChange(callback: ((theme: ThemeType.Light | ThemeType.Dark) => unknown)) {
+        ipcRenderer.on('systemThemeUpdated', (event, theme: ThemeType.Light | ThemeType.Dark) => callback(theme));
+    }
+
+    async getEffectiveTheme() {
```
Hinton, line 202: Would be nice to replace some of the logic in `services.module` in the various projects with this new logic. (To avoid duplicated code)

Hinton (approved):

PR #501, opened 2021-09-28
Resolve biometric error dissapearing before being read

Hinton:
## Objective

In the browser extension, primarily on Safari the error message for biometrics sometimes disappears before it can be read. To resolve this, I needed to only close the dialog if the action was successful, which requires to know the status of the unlock operation.

### DIFF
--- a/angular/src/components/lock.component.ts
+++ b/angular/src/components/lock.component.ts
@@ -166,6 +166,8 @@ export class LockComponent implements OnInit {
         if (success) {
             await this.doContinue();
         }
+
+        return success;
     }
 
     togglePassword() {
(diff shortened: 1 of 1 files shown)


angular/src/components/lock.component.ts
```diff
@@ -166,6 +166,8 @@ export class LockComponent implements OnInit {
         if (success) {
             await this.doContinue();
         }
+
```
djsmith85, line 169: As the function now returns a boolean, we should add the return type to the function signature.

djsmith85 (changes requested):

Hinton:
@djsmith85 They'll probably need updating but should be fine other than that.

PR #503, opened 2021-09-28
API service for fetching sso config

kspearrin:


Hinton:
Implemented in https://github.com/bitwarden/jslib/pull/520

PR #504, opened 2021-09-30
Feature/split manage collections permission

MGibson1:
# Overview

Asana: https://app.asana.com/0/0/1201089124980371/f

jslib work to split out collection permissions into create, edit, delete permissions.

# Files Changed
* **permissions.ts**: Deprecate old permissions, assigned numbers to permissions to allow for eventually removing deprecated permissions.
* **utils.ts**: Add helper to convert camelCase to PascalCase. Used to convert id attributes to name attributes.
* **permissionsApi.ts**: Add new permissions parsing. Keep the old, server provides them for backwards compatibility and if new permissions are null, we fallback to the manage permissions.
* **organization.ts**: Provide new permissions helpers. Fallback to manage permissions if new ones are null.

### DIFF
--- a/common/src/models/domain/organization.ts
+++ b/common/src/models/domain/organization.ts
@@ -108,12 +108,32 @@ export class Organization {
         return this.isAdmin || this.permissions.accessReports;
     }
 
-    get canManageAllCollections() {
-        return this.isAdmin || this.permissions.manageAllCollections;
+    get canCreateNewCollections() {
+        return this.isAdmin || (this.permissions.createNewCollections ?? this.permissions.manageAllCollections);
     }
 
-    get canManageAssignedCollections() {
-        return this.isManager || this.permissions.manageAssignedCollections;
+    get canEditAnyCollection() {
+        return this.isAdmin || (this.permissions.editAnyCollection ?? this.permissions.manageAllCollections);
+    }
+
+    get canDeleteAnyCollection() {
+        return this.isAdmin || (this.permissions.deleteAnyCollection ?? this.permissions.manageAllCollections);
+    }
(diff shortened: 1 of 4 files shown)


common/src/models/domain/organization.ts
```diff
@@ -108,12 +108,32 @@ export class Organization {
...
+
+    get canEditAssignedCollections() {
+        return this.isManager || (this.permissions.deleteAssignedCollections ?? this.permissions.manageAssignedCollections);
+    }
+
+    get canDeleteAssignedCollections() {
+        return this.isManager || (this.permissions.deleteAssignedCollections ?? this.permissions.manageAssignedCollections);
+    }
+
+    get canViewAssignedCollections() {
```
Hinton, line 135: Just a thought I had, would we gain anything from refactoring this to:

```
get canViewCollection(collection: Collection) {
	if (canEditAnyCollection || this.canDeleteAnyCollection) {
        return true;
    }
    if (collection.assigned == user) {
        return this.canDeleteAssignedCollections || this.canEditAssignedCollections;
    }
}
```

This might not quite be possible to do in the `Organization` domain model though and might belong in a permissions service or something similar.

Hinton (approved):

common/src/models/domain/organization.ts
MGibson1: I think that's definitely a good idea, but I want to do it in a permissions service.

In general, both on clients and server, I want to modify the way we handle permissions from asking for general permission to specific actions on specific models. What we're doing now uses knowledge of the guid id as implicit read permission, which I don't like.

PR #506, opened 2021-10-01
Remove Business Portal, add SSO configuration models

Hinton:
## Objective
Adds the necessary model changes to support setting up and configuring sso using the web vault.

### Noteworthy Code Changes
- **angular/src/components/environment.component.ts**: Remove `enterpriseUrl`, since the service is now gone.
- **common/src/services/environment.service.ts**: Remove enterprise related methods.
- **common/src/abstractions/api.service.ts**: Remove `getEnterprisePortalSignInToken`, add `getOrganizationSso` and `postOrganizationSso`.
- **common/src/enums/permissions.ts**: Remove `Permissions.AccessBusinessPortal`.
- **common/src/models/response/organization/organizationSsoResponse.ts**: Added response model for SSO configuration.
- **common/src/models/request/organization/organizationSsoRequest.ts**: Added request modal for SSO configuration.

### DIFF
--- a/common/src/enums/permissions.ts
+++ b/common/src/enums/permissions.ts
@@ -1,6 +1,5 @@
 export enum Permissions {
-    AccessBusinessPortal,
-    AccessEventLogs,
+    AccessEventLogs = 1,
     AccessImportExport,
     AccessReports,
     ManageAllCollections,
--- a/common/src/enums/permissions.ts
+++ b/common/src/enums/permissions.ts
@@ -1,27 +1,27 @@
 export enum Permissions {
-    AccessEventLogs = 1,
-    AccessImportExport = 2,
-    AccessReports = 3,
+    AccessEventLogs,
+    AccessImportExport,
+    AccessReports,
     /**
      * @deprecated Sep 29 2021: This permission has been split out to `createNewCollections`, `editAnyCollection`, and
      * `deleteAnyCollection`. It exists here for backwards compatibility with Server versions <= 1.43.0
      */
-    ManageAllCollections = 4,
+    ManageAllCollections,
     /**
      * @deprecated Sep 29 2021: This permission has been split out to `editAssignedCollections` and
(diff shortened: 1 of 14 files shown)


common/src/enums/permissions.ts
```diff
@@ -1,5 +1,4 @@
 export enum Permissions {
-    AccessBusinessPortal = 0,
```
MGibson1, line 2: We should leave this in commented out or deprecated as below (my vote is commented out if there are not upgrade concerns)

We can't reuse values in this enum, so it's best to leave reminders to what permission values were and what are available.

common/src/models/request/organization/organizationSsoRequest.ts
```diff
@@ -0,0 +1,4 @@
+export class OrganizationSsoRequest {
+    enabled: boolean = false;
+    data: any;
```
MGibson1, line 3: Is there some specific reason this needs to be `any` type? The type should be well-defined on the c# side, I'd prefer to use typescripts static-typing power whenever possible.

common/src/models/response/organization/organizationSsoResponse.ts
```diff
@@ -0,0 +1,22 @@
+import { BaseResponse } from '../baseResponse';
+
+export class OrganizationSsoResponse extends BaseResponse {
+    enabled: boolean;
+    data: any;
```
MGibson1, line 5: Same `any` concern

MGibson1 (changes requested):
Looks good, I'm just concerned about the data typing in the sso config request/response.

Nested types can be parsed similar to how we're doing it for `cipherResponses` in [`attachmentUploadDataResponse`](https://github.com/bitwarden/jslib/blob/master/common/src/models/response/attachmentUploadDataResponse.ts)

common/src/enums/permissions.ts
Hinton: I believe the enum is only used internally in the projects? So it should be possible to change the enum values.

MGibson1 (approved):

PR #527, opened 2021-10-25
Add support for requesting and using otp for verifying some requests

Hinton:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
Adds the missing functionality for using the Key Connector.
 * Password verification is handled using an email based OTP instead.
 * New component for removing the password of an account.

## Code changes

* **angular/src/components/add-edit.component.ts**: Disable reprompt checkbox.
* **angular/src/components/export.component.ts**: Changes the export form to use reactive forms since it simplified the template work a bit. Uses the new `UserVerificationService` to verify.
* **angular/src/components/lock.component.ts**: Changed lock component to log you out if you use the key connector without biometric or pin.
* **angular/src/components/sso.component.ts**: Update crypto agent references to key connector. Also fixed a spelling error in a method.
* **angular/src/components/verify-master-password.component.ts**: Added new component for verifying using either master password or OTP.
* **angular/src/services/passwordReprompt.service.ts**: Disable password reprompt when using key connector.
* **common/src/enums/eventType.ts**: Add new events for enabling/disabling sso/key connector.
* **common/src/enums/verificationType.ts**: Add new enum for identifying which operation is used to verify.
* **common/src/models/api/ssoConfigApi.ts**: Rename crypto agent to key connector.
* **common/src/models/request/account/verifyOtpRequest.ts**: New request for verifying OTP.
* **common/src/models/request/passwordVerificationRequest.ts**: Add otp as alternative verification method.
* **common/src/models/request/tokenRequest.ts**: Include orgIdentifier in TokenRequest. Used to determine users org when using api key.
* **common/src/models/response/identityTokenResponse.ts**: Rename crypto agent references.
* **common/src/models/response/profileOrganizationResponse.ts**: Include organization key connector info in sync.
* **common/src/models/response/profileProviderOrganizationResponse.ts**: Explicitly extend ProfileOrganizationResponse since that's the behavior services expects.
* **common/src/models/response/profileResponse.ts**: Include details if the user uses key connector.
* **common/src/services/api.service.ts**: Add new endpoints related to converting existing accounts to key connector, OTP and more.
* **common/src/services/auth.service.ts**: Update auth service to send along org identifier in `logInApiKey`, rename crypto agent to key connector.
* **common/src/services/sync.service.ts**: Add logic for redirecting to "Remove password" page when the account should be migrated to key connector.
* **common/src/services/token.service.ts**: Add helper for determining if the user logged in using SSO.
* **common/src/services/user.service.ts**: Set whether the user uses key connector or not.
* **common/src/services/userVerification.service.ts**: New service for handling user verification using either password or OTP.
* **common/src/services/vaultTimeout.service.ts**: Log out user if using key connector and no biometric/pin is set.
* **node/src/cli/commands/login.command.ts** - (see the CLI PR for more context)
  * moved the full sync here (from the child component) because it's required before we can check if the user needs to migrate to Key Connector.
  * add migration flow - see the `migrateToKeyConnector` method
  * in the SSO flow (using the `--sso` option), get the Organization Identifier back from the browser popup so that we can feed it through to `auth.service`.

## Before you submit

### DIFF
--- a/angular/src/components/export.component.ts
+++ b/angular/src/components/export.component.ts
@@ -5,6 +5,7 @@ import {
     Output,
     ViewChild,
 } from '@angular/core';
+import { FormControl } from '@angular/forms';
 
 import { CryptoService } from 'jslib-common/abstractions/crypto.service';
 import { EventService } from 'jslib-common/abstractions/event.service';
@@ -28,6 +29,7 @@ export class ExportComponent implements OnInit {
     format: 'json' | 'encrypted_json' | 'csv' = 'json';
     showPassword = false;
     disabledByPolicy: boolean = false;
+    secret = new FormControl();
 
     constructor(protected cryptoService: CryptoService, protected i18nService: I18nService,
         protected platformUtilsService: PlatformUtilsService, protected exportService: ExportService,
@@ -40,6 +42,9 @@ export class ExportComponent implements OnInit {
 
     async checkExportDisabled() {
         this.disabledByPolicy = await this.policyService.policyAppliesToUser(PolicyType.DisablePersonalVaultExport);
+        if (this.disabledByPolicy) {
+            this.secret.disable();
+        }
     }
 
     get encryptedFormat() {
--- a/angular/src/components/export.component.ts
+++ b/angular/src/components/export.component.ts
@@ -33,7 +33,7 @@ export class ExportComponent implements OnInit {
     exportForm = this.fb.group({
         format: ['json'],
         secret: [''],
-    })
+    });
 
     formatOptions = [
         { name: '.json', value: 'json' },
--- a/angular/src/components/export.component.ts
+++ b/angular/src/components/export.component.ts
@@ -21,7 +21,7 @@ import { VerificationType } from 'jslib-common/enums/verificationType';
 
 import { VerifyOtpRequest } from 'jslib-common/models/request/account/verifyOtpRequest';
 
-import { Verification } from './verify-master-password.component';
+import { Verification } from 'jslib-common/types/verification';
 
 @Directive()
 export class ExportComponent implements OnInit {
--- a/angular/src/components/export.component.ts
+++ b/angular/src/components/export.component.ts
@@ -4,7 +4,7 @@ import {
     OnInit,
     Output,
 } from '@angular/core';
-import { FormControl } from '@angular/forms';
+import { FormBuilder } from '@angular/forms';
 
 import { ApiService } from 'jslib-common/abstractions/api.service';
 import { CryptoService } from 'jslib-common/abstractions/crypto.service';
@@ -28,15 +28,23 @@ export class ExportComponent implements OnInit {
     @Output() onSaved = new EventEmitter();
 
     formPromise: Promise<string>;
-    format: 'json' | 'encrypted_json' | 'csv' = 'json';
-    showPassword = false;
     disabledByPolicy: boolean = false;
-    verificationForm = new FormControl();
+
+    exportForm = this.fb.group({
+        format: ['json'],
+        secret: [''],
+    })
+
+    formatOptions = [
+        { name: '.json', value: 'json' },
+        { name: '.csv', value: 'csv' },
+        { name: '.json (Encrypted)', value: 'encrypted_json' },
+    ];
 
     constructor(protected cryptoService: CryptoService, protected i18nService: I18nService,
         protected platformUtilsService: PlatformUtilsService, protected exportService: ExportService,
         protected eventService: EventService, private policyService: PolicyService, protected win: Window,
-        private logService: LogService, private apiService: ApiService) { }
+        private logService: LogService, private apiService: ApiService, private fb: FormBuilder) { }
 
     async ngOnInit() {
         await this.checkExportDisabled();
@@ -45,7 +53,7 @@ export class ExportComponent implements OnInit {
     async checkExportDisabled() {
         this.disabledByPolicy = await this.policyService.policyAppliesToUser(PolicyType.DisablePersonalVaultExport);
         if (this.disabledByPolicy) {
-            this.verificationForm.disable();
+            this.exportForm.disable();
         }
     }
 
@@ -74,6 +82,7 @@ export class ExportComponent implements OnInit {
             this.downloadFile(data);
             this.saved();
             await this.collectEvent();
+            this.exportForm.get('secret').setValue('');
         } catch (e) {
             this.logService.error(e);
         }
@@ -121,7 +130,7 @@ export class ExportComponent implements OnInit {
     }
 
     protected async verifySecret(): Promise<boolean> {
-        const verification: Verification = this.verificationForm.value;
+        const verification: Verification = this.exportForm.get('secret').value;
         if (verification?.secret == null || verification.secret === '') {
             return false;
         }
@@ -144,6 +153,10 @@ export class ExportComponent implements OnInit {
             }
         }
         return true;
+    }
+
+    get format() {
+        return this.exportForm.get('format').value;
     }
 
     private downloadFile(csv: string): void {
--- a/angular/src/components/export.component.ts
+++ b/angular/src/components/export.component.ts
@@ -6,7 +6,6 @@ import {
 } from '@angular/core';
 import { FormBuilder } from '@angular/forms';
 
-import { ApiService } from 'jslib-common/abstractions/api.service';
 import { CryptoService } from 'jslib-common/abstractions/crypto.service';
 import { EventService } from 'jslib-common/abstractions/event.service';
 import { ExportService } from 'jslib-common/abstractions/export.service';
@@ -14,14 +13,10 @@ import { I18nService } from 'jslib-common/abstractions/i18n.service';
 import { LogService } from 'jslib-common/abstractions/log.service';
 import { PlatformUtilsService } from 'jslib-common/abstractions/platformUtils.service';
 import { PolicyService } from 'jslib-common/abstractions/policy.service';
+import { UserVerificationService } from 'jslib-common/abstractions/userVerification.service';
 
 import { EventType } from 'jslib-common/enums/eventType';
 import { PolicyType } from 'jslib-common/enums/policyType';
-import { VerificationType } from 'jslib-common/enums/verificationType';
-
-import { VerifyOtpRequest } from 'jslib-common/models/request/account/verifyOtpRequest';
-
-import { Verification } from 'jslib-common/types/verification';
 
 @Directive()
 export class ExportComponent implements OnInit {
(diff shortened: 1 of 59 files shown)


angular/src/components/export.component.ts
```diff
@@ -123,6 +129,38 @@ export class ExportComponent implements OnInit {
         await this.eventService.collect(EventType.User_ClientExportedVault);
     }
 
+    protected async verifySecret(): Promise<boolean> {
+        const verification: Verification = this.exportForm.get('secret').value;
+        if (verification?.secret == null || verification.secret === '') {
+            this.platformUtilsService.showToast('error', this.i18nService.t('errorOccurred'),
+                this.i18nService.t('verificationCodeRequired'));
```
Hinton, line 136: Do we need to use `invalidMasterPassword` here if not using crypto agent?

common/src/services/user.service.ts
```diff
@@ -229,4 +249,29 @@ export class UserService implements UserServiceAbstraction {
     async clearProviders(userId: string): Promise<any> {
         await this.storageService.remove(Keys.providersPrefix + userId);
     }
+
+    async buildVerificationRequest<T extends PasswordVerificationRequest>
```
Hinton, line 253: I think we should move this to a different service.

angular/src/components/export.component.ts
eliykat: As discussed, we'll rely on the HTML5 `required` attribute for form validation. That is already implemented in the `VerifyMasterPasswordComponent` template.

This means that the error here does not have to be gracefully handled or user friendly, as it should never happen. I've changed it to just `throw` a non-translated message for debugging purposes.

common/src/services/user.service.ts
eliykat: Added `userVerificationService` as discussed - thanks for the suggestion.

common/src/services/keyConnector.service.ts
```diff
@@ -0,0 +1,86 @@
...
+        } catch (e) {
+            throw new Error('Unable to reach key connector');
+        }
+
+        await this.apiService.postConvertToKeyConnector();
+    }
+
+    async getAndSetKey(url?: string) {
+        if (url == null) {
+            url = this.environmentService.getKeyConnectorUrl();
```
eliykat, line 60: This environment variable is only used for CLI, I'm not sure whether it's better to use it as the default here, or just require it to be passed in.

eliykat (commented):
I've reviewed and left some questions on both @Hinton's code and my own.

common/src/services/constants.service.ts
```diff
@@ -65,4 +66,5 @@ export class ConstantsService {
     readonly biometricText: string = ConstantsService.biometricText;
     readonly biometricAwaitingAcceptance: string = ConstantsService.biometricAwaitingAcceptance;
     readonly biometricFingerprintValidated: string = ConstantsService.biometricFingerprintValidated;
+    readonly convertAccountToKeyConnector: string = ConstantsService.convertAccountToKeyConnector;
```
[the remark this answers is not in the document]
Hinton: I don't want to mix up storage and message strings. I think we need to add a new constant for messages which feels out of scope for this PR.

common/src/services/sync.service.ts
```diff
@@ -316,6 +320,11 @@ export class SyncService implements SyncServiceAbstraction {
                 organizations[o.id].isProviderUser = true;
             }
         });
+
+        if (this.tokenService.getIsExternal() && await this.keyConnectorService.userNeedsMigration()) {
```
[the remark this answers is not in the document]
Hinton: That the user logged in using SSO.

common/src/services/keyConnector.service.ts
```diff
@@ -0,0 +1,86 @@
...
+            const keyArr = Utils.fromB64ToArray(userKeyResponse.key);
+            const k = new SymmetricCryptoKey(keyArr);
+            await this.cryptoService.setKey(k);
+        } catch (e) {
+            this.logService.error(e);
+            throw new Error('Unable to reach key connector');
+        }
+    }
+
+    async getManagingOrganization() {
```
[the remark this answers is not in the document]
Hinton: It's probably a bit redundant to check if the user is a `providerUser` since those always gets the role owner. But should we ever change that logic it's good to have the check.

angular/src/components/remove-password.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    async convert() {
+        this.continuing = true;
+        this.actionPromise = this.keyConnectorService.migrateUser();
+
+        try {
+            await this.actionPromise;
+            this.platformUtilsService.showToast('success', null, this.i18nService.t('removedMasterPassword'));
+            await this.storageService.remove(ConstantsService.convertAccountToKeyConnector);
```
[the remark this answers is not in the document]
Hinton: I think we should set it on all clients to prevent the user from navigating away.

common/src/models/request/twoFactorEmailRequest.ts
```diff
@@ -2,10 +2,4 @@ import { PasswordVerificationRequest } from './passwordVerificationRequest';
 
 export class TwoFactorEmailRequest extends PasswordVerificationRequest {
     email: string;
-
-    constructor(email: string, masterPasswordHash: string) {
-        super();
-        this.masterPasswordHash = masterPasswordHash;
-        this.email = email;
-    }
```
addisonbeck, line 10: We hate constructors now or something?

angular/src/components/export.component.ts
```diff
@@ -21,22 +23,34 @@ export class ExportComponent implements OnInit {
...
     formPromise: Promise<string>;
-    masterPassword: string;
-    format: 'json' | 'encrypted_json' | 'csv' = 'json';
-    showPassword = false;
     disabledByPolicy: boolean = false;
 
+    exportForm = this.fb.group({
+        format: ['json'],
+        secret: [''],
+    });
```
addisonbeck, line 31: Me when somebody refactor an existing form to use reactive forms
giphy

common/src/services/userVerification.service.ts
```diff
@@ -0,0 +1,70 @@
...
+
+@Injectable()
+export class UserVerificationService implements UserVerificationServiceAbstraction {
+    constructor(private cryptoService: CryptoService, private i18nService: I18nService,
+        private platformUtilsService: PlatformUtilsService, private apiService: ApiService,
+        private logService: LogService) { }
+
+    async buildRequest<T extends PasswordVerificationRequest>(verification: Verification,
+        requestClass?: new () => T, alreadyHashed?: boolean) {
+        if (verification?.secret == null || verification.secret === '') {
```
addisonbeck, line 26: verification?.secret == null checks for '' here, doesn't it?

addisonbeck (commented):
Not as much feedback from me on this one. Great work Oscar and Thomas!

common/src/services/sync.service.ts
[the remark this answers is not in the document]
eliykat: CLI also uses that service method when logging in with the api key, so I think we need to leave it here.

common/src/services/constants.service.ts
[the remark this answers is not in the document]
eliykat: Yep good point

common/src/services/keyConnector.service.ts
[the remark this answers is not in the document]
eliykat: I didn't realise that, it's confusing to me that the `Owner` role does not just equal `OrganizationUserType.Owner`. But if that's the pattern that we're following, then I think we should follow it and not introduce the redundant check.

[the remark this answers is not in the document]
eliykat: Wait, `organization.type` returns `OrganizationUserType`, but `get isOwner()` returns `OrganizationUserStatusType.Owner || isProviderUser`? Seems bad to me; I will still refactor to use the getter though

common/src/services/keyConnector.service.ts
eliykat: No-one seems to care so I'm going to leave as is

angular/src/components/verify-master-password.component.ts
```diff
@@ -0,0 +1,81 @@
...
+            multi: true,
+            useExisting: VerifyMasterPasswordComponent,
+        },
+    ],
+})
+export class VerifyMasterPasswordComponent implements ControlValueAccessor, OnInit {
+    usesKeyConnector: boolean = false;
+    disableRequestOtp: boolean = false;
+
+    secret = new FormControl('');
```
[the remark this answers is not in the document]
eliykat: One of my fav gifs

common/src/models/request/twoFactorEmailRequest.ts
eliykat: It's to make it compatible with the factory pattern in `TwoFactorBaseComponent.buildRequestModel` (which in turn uses `UserVerificationService.buildRequest`). That factory assumes that the model you want to build does not take any args in the ctor, which was true for every other request model except for this one, so this seemed like the easiest fix.

angular/src/components/export.component.ts
eliykat: blame @Hinton ... but he's converting me

common/src/services/userVerification.service.ts
eliykat: I still assume the user can submit an empty string, unless reactive forms change this to a null value?

angular/src/components/verify-master-password.component.html
```diff
@@ -0,0 +1,17 @@
+<ng-container *ngIf="!usesKeyConnector">
+    <label for="masterPassword">{{'masterPass' | i18n}}</label>
+    <input id="masterPassword" type="password" name="MasterPasswordHash" class="form-control"
+        [formControl]="secret" required appAutofocus appInputVerbatim>
+</ng-container>
+<ng-container *ngIf="usesKeyConnector">
+    <div class="form-group">
+        <input type="button" class="btn btn-primary" value="Request one-time password" (click)="requestOtp()"
```
eliykat, line 8: * button text needs to be i18n'd
* not sure about `input type="button"`, why not just `<button>` ?

Hinton: Done.

common/src/services/userVerification.service.ts
Hinton: `null != ''`. We could change it to `if (verification?.secret)`, but that feels less explicit.

common/src/models/request/twoFactorEmailRequest.ts
addisonbeck: Thanks for the explanation, Thomas.

addisonbeck (dismissed):

addisonbeck (approved):

addisonbeck (approved):

PR #542, opened 2021-11-11
[Linked Fields] Fix QA feedback

eliykat:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix QA feedback on linked fields:
* CLI did not show the new `Field.LinkedId` property, because it was missing from the export models (which are used by CLI for json string representations)
* Angular clients were making all field types "linked" when the cipher type changed

## Code changes

* **common/src/models/export/field.ts** - add `linkedId` to the export model. I haven't added it to the template, because it's an optional field and I think it's unlikely to be set via CLI (it takes an enum, it's for use in browser, it's only used for 1 field type).
* **angular/src/components/add-edit-custom-fields.component.ts** - change `=` to `===` in the filter

jslib will need to be updated on all Angular clients.

## Testing requirements

No additional testing other than making sure the bugs are fixed.

## Before you submit

### DIFF
--- a/angular/src/components/add-edit-custom-fields.component.ts
+++ b/angular/src/components/add-edit-custom-fields.component.ts
@@ -118,7 +118,7 @@ export class AddEditCustomFieldsComponent implements OnChanges {
         }
 
         this.cipher.fields
-            .filter(f => f.type = FieldType.Linked)
+            .filter(f => f.type === FieldType.Linked)
             .forEach(f => f.linkedId = this.linkedFieldOptions[0].value);
     }
 }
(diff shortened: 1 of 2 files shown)


djsmith85 (approved):
Adding to the export model and especially to the `toDomain` method also includes the linkedId within an import of an exported Bitwarden vault 👍  Not quite sure what you meant about the CLI template, though?

eliykat:
Good point about supporting exporting & importing.

> Not quite sure what you meant about the CLI template, though?

You can ask CLI to generate a template json string that you can then manipulate using `jq`: https://bitwarden.com/help/article/cli/#get-template. But the `linkedId` field is not required and would probably only cause confusion if we included it there.

PR #543, opened 2021-11-11
[Key Connector] Add getKeyConnectorAlive to ApiService

eliykat:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Before saving Key Connector config, ping the Key Connector URL to make sure it's alive.

## Code changes

* **common/src/services/api.service.ts** - add the API call to the api service. This uses the existing `/alive` endpoint on Key Connector, which should just return a JSON containing the current date and time.

## Testing requirements

Try saving the Key Connector config with valid and invalid Key Connector URLs. If the URL is invalid (i.e. Key Connector cannot be reached), you should get an error toast.

## Before you submit

### DIFF
--- a/common/src/abstractions/api.service.ts
+++ b/common/src/abstractions/api.service.ts
@@ -455,4 +455,5 @@ export abstract class ApiService {
 
     getUserKeyFromKeyConnector: (keyConnectorUrl: string) => Promise<KeyConnectorUserKeyResponse>;
     postUserKeyToKeyConnector: (keyConnectorUrl: string, request: KeyConnectorUserKeyRequest) => Promise<void>;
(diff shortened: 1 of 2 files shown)


Hinton:
@eliykat We'll need to figure out what to do with the alive endpoint as well. Currently it only verifies that something is reachable at the url.

eliykat:
@Hinton yes you made a good point that we'll want to do a basic health check, which is def out of scope for this release. I'll just let this sit here until we come back to it.

eliykat:
As discussed, I suggest we merge this as an MVP and build a more comprehensive health check later if required.

Hinton (approved):

PR #544, opened 2021-11-11
[Key Connector] QA fixes for CLI and Desktop

eliykat:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix QA feedback on Key Connector:

* Fix no error in desktop when OTP/MP is blank: https://app.asana.com/0/0/1201361837847153/f
* Support CLI changes: https://github.com/bitwarden/cli/pull/410

## Code changes

* **common/src/services/userVerification.service.ts** - 
  * add wrapper for requesting OTP code, it just made sense to have it in this service as the main interface
  * provide CLI support, which I forgot the first time round:
    * break out error handling into its own `protected` method. This lets us override it in CLI, which doesn't use toasts.
    * remove `@Injectable` Angular decorator, which doesn't belong in `jslib-common`. Angular clients will need their `services.module.ts` updated.
  * add proper error handling if the secret is null or an empty string (tell the user, don't just throw an error)
  * remove `logService`, I don't think it's necessary where we're surfacing the error to the user
* **common/src/services/keyConnector.service.ts** - don't ask `environmentService.getKeyConnectorUrl()` for the url as default behaviour. This is only used in one very specific use case (using apikey in CLI), so it didn't make sense to have it as a fallback. I also removed the null check because the try/catch block below will catch a null or empty value.
* **common/src/models/response/identityTokenResponse.ts** - add support for `apiUseKeyConnector` flag, which tells CLI to check for the saved key connector URL. See CLI PR for more context.

## Before you submit

### DIFF
--- a/common/src/abstractions/userVerification.service.ts
+++ b/common/src/abstractions/userVerification.service.ts
@@ -6,4 +6,5 @@ export abstract class UserVerificationService {
     buildRequest: <T extends SecretVerificationRequest> (verification: Verification,
         requestClass?: new () => T, alreadyHashed?: boolean) => Promise<T>;
     verifyUser: (verification: Verification) => Promise<boolean>;
+    requestOTP: () => Promise<void>;
 }
--- a/common/src/services/userVerification.service.ts
+++ b/common/src/services/userVerification.service.ts
@@ -19,7 +19,12 @@ export class UserVerificationService implements UserVerificationServiceAbstracti
     async buildRequest<T extends SecretVerificationRequest>(verification: Verification,
         requestClass?: new () => T, alreadyHashed?: boolean) {
         if (verification?.secret == null || verification.secret === '') {
-            throw new Error('No secret provided for verification.');
+            if (verification.type === VerificationType.OTP) {
+                this.handleError(this.i18nService.t('verificationCodeRequired'));
+            } else {
+                this.handleError(this.i18nService.t('masterPasswordRequired'));
+            }
+            return null;
         }
 
         const request = requestClass != null
@@ -40,10 +45,11 @@ export class UserVerificationService implements UserVerificationServiceAbstracti
     async verifyUser(verification: Verification): Promise<boolean> {
         if (verification?.secret == null || verification.secret === '') {
             if (verification.type === VerificationType.OTP) {
-                this.showError(this.i18nService.t('verificationCodeRequired'));
+                this.handleError(this.i18nService.t('verificationCodeRequired'));
             } else {
-                this.showError(this.i18nService.t('masterPasswordRequired'));
+                this.handleError(this.i18nService.t('masterPasswordRequired'));
             }
+            return false;
         }
 
         if (verification.type === VerificationType.OTP) {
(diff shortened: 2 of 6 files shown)


common/src/abstractions/userVerification.service.ts
```diff
@@ -6,4 +6,5 @@ export abstract class UserVerificationService {
     buildRequest: <T extends SecretVerificationRequest> (verification: Verification,
         requestClass?: new () => T, alreadyHashed?: boolean) => Promise<T>;
     verifyUser: (verification: Verification) => Promise<boolean>;
+    requestOTP: () => Promise<void>;
```
Hinton, line 9: Is this some way to avoid adding another dependency or? Seems to just be a wrapper to ApiService.

common/src/services/userVerification.service.ts
```diff
@@ -1,11 +1,8 @@
-import { Injectable } from '@angular/core';
```
Hinton, line 1: Oops, how did that build

common/src/abstractions/userVerification.service.ts
eliykat: It is just a wrapper, but it made sense to have request OTP and verify OTP in the same service. I don't really like calling `apiService` directly when you've got a more specific service handling that domain.

common/src/services/userVerification.service.ts
eliykat: 🙈

common/src/services/userVerification.service.ts
```diff
@@ -19,7 +19,12 @@ export class UserVerificationService implements UserVerificationServiceAbstracti
     async buildRequest<T extends SecretVerificationRequest>(verification: Verification,
         requestClass?: new () => T, alreadyHashed?: boolean) {
         if (verification?.secret == null || verification.secret === '') {
-            throw new Error('No secret provided for verification.');
+            if (verification.type === VerificationType.OTP) {
+                this.handleError(this.i18nService.t('verificationCodeRequired'));
+            } else {
+                this.handleError(this.i18nService.t('masterPasswordRequired'));
+            }
```
Hinton: We duplicate this code with `verifyUser`, can we extract it to a function?

`this.handleError(this.getError(verification.type)`

common/src/abstractions/userVerification.service.ts
Hinton: No, I agree with encapsulating logic into services. Would be nice to remember to update references though. (In some ways we should stop directly using apiService and have it be used by other services in between but that's a topic for another day.)

common/src/services/userVerification.service.ts
```diff
@@ -15,16 +12,19 @@ import { SecretVerificationRequest } from '../models/request/secretVerificationR
...
     async buildRequest<T extends SecretVerificationRequest>(verification: Verification,
         requestClass?: new () => T, alreadyHashed?: boolean) {
         if (verification?.secret == null || verification.secret === '') {
-            throw new Error('No secret provided for verification.');
+            if (verification.type === VerificationType.OTP) {
+                this.handleError(this.i18nService.t('verificationCodeRequired'));
+            } else {
+                this.handleError(this.i18nService.t('masterPassRequired'));
+            }
+            return null;
```
Hinton, line 27: I'm slightly hesitant to this change, previously we threw an error which got handled by the components, do we plan on updating each component to do a null check?

It feels a bit like a side effect to have the `buildRequest` create a toast.

common/src/abstractions/userVerification.service.ts
eliykat: I'll update any references in the clients when I update jslib

common/src/services/userVerification.service.ts
eliykat: This is done

common/src/services/userVerification.service.ts
eliykat: This block has been extracted to `private verifyInput`

Hinton (approved):

PR #560, opened 2021-11-19
Add support for hiding input on lock screen if using key connector

Hinton:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Adds support for hiding the password input on the lock screen if using key connector.

https://app.asana.com/0/1201363553288890/1201390561003506

## Before you submit

### DIFF
--- a/angular/src/components/lock.component.ts
+++ b/angular/src/components/lock.component.ts
@@ -37,6 +37,7 @@ export class LockComponent implements OnInit {
     supportsBiometric: boolean;
     biometricLock: boolean;
     biometricText: string;
+    hideInput: boolean;
 
     protected successRoute: string = 'vault';
     protected onSuccessfulSubmit: () => void;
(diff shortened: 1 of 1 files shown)


angular/src/components/lock.component.ts
```diff
@@ -60,6 +61,7 @@ export class LockComponent implements OnInit {
             (await this.cryptoService.hasKeyStored('biometric') || !this.platformUtilsService.supportsSecureStorage());
         this.biometricText = await this.storageService.get(ConstantsService.biometricText);
         this.email = await this.userService.getEmail();
+        this.hideInput = await this.keyConnectorService.getUsesKeyConnector() && !this.pinLock;
 
         // Users with key connector and without biometric or pin has no MP to unlock using
         if (await this.keyConnectorService.getUsesKeyConnector() && !(this.biometricLock || this.pinLock)) {
```
djsmith85, line 67: Should we only call `await this.keyConnectorService.getUsesKeyConnector()` once and put the result into a variable. Saves us waiting for another async/await call.

djsmith85 (changes requested):

angular/src/components/lock.component.ts
Hinton: Maybe.

djsmith85 (approved):

Issue #557, opened 2021-11-18
Toggle Visibility of the master password move cursor to start of input


chao813:
On Chrome, when you toggle visibility of the master password, the cursor moves to the start of the input field. Noticed it does this on Chrome web vault and Chrome browser extension. I tried on Safari and can't reproduce this issue. 

There was an issue [(browser#2178)](https://github.com/bitwarden/browser/issues/2178) opened in the browser repo but since the code is in jslib, thought I'd open an issue here. I have a work around but was wondering if this was already tackled before I open a pull request.

eliykat:
Hi @chao813, thanks for creating this issue and the PR. It can be a bit difficult tracking issues across all our different client repos, however let's just use https://github.com/bitwarden/browser/issues/2178 to track this issue so that we're not creating duplicates.

PR #561, opened 2021-11-20
Fix cursor location changing issue on toggle password

chao813:
## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
When you toggle visibility of the master password, the cursor moves to the start of the input field. Noticed it does this on Chrome web vault and Chrome browser extension. I tried on Safari and can't reproduce this issue. 

## Code changes
#557 , also [2178](https://github.com/bitwarden/browser/issues/2178) 
Updated togglePassword() function to work with Chrome browser. Added a setTimeout of 0 ms (to defer the operation until the stack is clear) on focus since Chrome has an odd quirk where the focus event fires before the cursor is moved into the field. 

* **login.component.ts:** Updated togglePassword function in `login.component.ts`

## Testing requirements
Test toggle password visibility on all platforms. 

## Before you submit


chao813:
@eliykat Just saw your comment on the issue I opened, should I move this PR into the browser repo or just leave this here for you to review?

eliykat:
It's fine here - this is the right place for the PR. Because jslib is shared by the clients, it's totally normal to have a PR in jslib that fixes an issue reported in a client.

angular/src/components/login.component.ts
```diff
@@ -132,7 +132,12 @@ export class LoginComponent extends CaptchaProtectedComponent implements OnInit
 
     togglePassword() {
         this.showPassword = !this.showPassword;
-        document.getElementById('masterPassword').focus();
+        const masterPassword: HTMLInputElement = <HTMLInputElement>document.getElementById('masterPassword');
```
eliykat, line 135: Code style: no need to specify the type of the `const`, it'll figure it out from the value being assigned.

chao813:
I figured, thanks!

angular/src/components/login.component.ts
```diff
@@ -132,7 +132,12 @@ export class LoginComponent extends CaptchaProtectedComponent implements OnInit
...
+        const masterPasswordCursorLocation = masterPassword.selectionStart;
+        setTimeout(function(){
```
eliykat, line 137: I assume this is to avoid a `ExpressionChangedAfterItHasBeenCheckedError`? Using `setTimeout` to solve this problem seems hacky, I think we want to avoid this.

How about using the pattern [here](https://github.com/bitwarden/jslib/blob/e1c6e4973a2532b14e3187a79688e8b0a89e9d5c/angular/src/directives/autofocus.directive.ts#L26) instead? It's still more verbose than I'd like, but it's how I solved a similar problem last time, and it seems like better practice.

eliykat (changes requested):
Thanks again for the PR, please see my review comments below.

Once we have the changes settled here, it would be good to replicate them in `lock.component.ts` as well. There are many places in our app that we have a show/hide button like this, it would be good to encapsulate this into a re-usable component, but that's a task for another day. This would at least solve the problem on the 2 most common pages.

angular/src/components/login.component.ts
chao813: Yeah I guess using `setTimeout` does seem a little hacky. Updated the code to use pattern suggested.

eliykat:
Does this change (waiting until `ngZone` is stable) fix the original bug, without having to use `setSelectionRange`? If so, that's great! I didn't actually realise that it would work that way 😅 

Can you please replicate for `lock.component.ts`, then if it all looks good we can merge.

chao813:
@eliykat I believe it did, didn't need to use `setSelectionRange`. 
I just replicated it in `lock.component.ts` as well.

angular/src/components/lock.component.ts
```diff
@@ -185,7 +186,11 @@ export class LockComponent implements OnInit {
 
     togglePassword() {
         this.showPassword = !this.showPassword;
-        document.getElementById(this.pinLock ? 'pin' : 'masterPassword').focus();
+        if (this.ngZone.isStable) {
+            document.getElementById(this.pinLock ? 'pin' : 'masterPassword').focus();
```
eliykat, line 190: Better to extract this this conditional to its own variable, e.g.
`const input = document.getElementById(this.pinLock ? 'pin' : 'masterPassword');`

To avoid repeating it in the `if` and `else` blocks.

eliykat (changes requested):

eliykat (approved):
Nicely done, thanks for the PR!
This will be in the January release.

xSepticSidx:
Many thanks to all for sorting this issue out.

PR #565, opened 2021-11-25
Move DI logic to jslib

Hinton:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Moves most of the Dependency Injection logic from individual clients to jslib. This will streamline development processes since we no longer need to update each client when dependencies are changed.

## Testing requirements

Regular regression testing should catch any potential bugs. (Services instantiated using wrong dependencies)

## Before you submit

### DIFF
--- a/angular/src/services/jslib-services.module.ts
+++ b/angular/src/services/jslib-services.module.ts
@@ -1,0 +1,416 @@
+import {
+    Injector,
+    LOCALE_ID,
+    NgModule,
+} from '@angular/core';
+import { ToasterModule } from 'angular2-toaster';
+
+import { ApiService } from 'jslib-common/services/api.service';
+import { AppIdService } from 'jslib-common/services/appId.service';
+import { AuditService } from 'jslib-common/services/audit.service';
+import { AuthService } from 'jslib-common/services/auth.service';
+import { CipherService } from 'jslib-common/services/cipher.service';
+import { CollectionService } from 'jslib-common/services/collection.service';
+import { ConsoleLogService } from 'jslib-common/services/consoleLog.service';
+import { CryptoService } from 'jslib-common/services/crypto.service';
+import { EnvironmentService } from 'jslib-common/services/environment.service';
+import { EventService as EventLoggingService } from 'jslib-common/services/event.service';
+import { ExportService } from 'jslib-common/services/export.service';
+import { FileUploadService } from 'jslib-common/services/fileUpload.service';
+import { FolderService } from 'jslib-common/services/folder.service';
+import { ImportService } from 'jslib-common/services/import.service';
+import { KeyConnectorService } from 'jslib-common/services/keyConnector.service';
+import { NotificationsService } from 'jslib-common/services/notifications.service';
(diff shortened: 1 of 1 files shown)


angular/src/services/jslib-services.module.ts
```diff
@@ -0,0 +1,416 @@
...
+            provide: SearchServiceAbstraction,
+            useClass: SearchService,
+            deps: [
+                CipherServiceAbstraction,
+                LogService,
+                I18nServiceAbstraction,
+            ],
+        },
+        {
+            provide: ImportServiceAbstraction,
```
eliykat, line 322: You can only import in the web vault, I think this should be omitted from the jslib module and provided for in the client's `services.module.ts` instead.

angular/src/services/jslib-services.module.ts
```diff
@@ -0,0 +1,416 @@
...
+            deps: [
+                CryptoServiceAbstraction,
+                UserServiceAbstraction,
+                ApiServiceAbstraction,
+                StorageServiceAbstraction,
+                I18nServiceAbstraction,
+                CipherServiceAbstraction,
+            ],
+        },
+        { provide: LogService, useFactory: () => new ConsoleLogService(false) },
```
eliykat, line 162: A bit out of scope for this PR, but I've always thought it odd that this bool is hard coded. From memory it sets the log level based on whether you're running in development mode. Would be nice to detect this from the environment (if possible) so that it's more usable for debugging.

eliykat (commented):
This is a great change, I'll leave my general comments on the web PR.

angular/src/services/jslib-services.module.ts
Hinton: Should we also look into moving the code to `web`? To minimize the need to jump between the repos.

angular/src/services/jslib-services.module.ts
Hinton: I agree. We can't simply inject platformUtils though since that will cause a circular dependency. I think we'll need to extract the `isDev` logic since it's currently also the source of another circular dependency.

eliykat: `isDevService` 😆

angular/src/services/jslib-services.module.ts
eliykat: I would be okay with that, I assume that it's in jslib because import & export logically go together, but I think we should be stricter with only including _shared_ code in jslib.

MGibson1: you can import in CLI too

eliykat: Ah good point @MGibson1. Then we have to leave it in jslib. I would still exclude it from the exported module though.

angular/src/services/jslib-services.module.ts
Hinton: That's fair. But moving the `isDev` logic will also solve one of the circular dependencies. So probably worthwhile to still extract. But I think both those changes are out of scope for this PR.

eliykat (approved):

PR #567, opened 2021-11-28
Change ngZone from private to protected

eliykat:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Follow up from https://github.com/bitwarden/jslib/pull/561. `ngZone` was marked as `private` in the base `lock` and `login` components, but it's also used in the subclasses in desktop, so it needs to be `protected` instead.

## Before you submit


djsmith85 (changes requested):
@eliykat looks good. Could you please delete the .DS_Store file which got added with #561

eliykat:
Good catch, done.

djsmith85 (approved):

PR #569, opened 2021-12-03
Fix for issue #1287 in bitwarden/web

linusaarnio:
Format the fieldvalue as a LocaleDateString instead of epoch when importing a date from 1P

This would be better solved by storing it as a date FieldType instead of Text. But since it is unclear when new field types are added, this solution serves as a fix for now and also guides the solution when new fieldtype exists.

I did not want to try and add the FieldType now, since I know it is on your roadmap to add more types but I added a TODO to implement this better when the new type exists.

## Type of change
- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Closes: https://github.com/bitwarden/web/issues/1287
Closes: https://github.com/bitwarden/server/issues/362

## Code changes
If a field imported from 1password is marked as 'date', it will now be stored as a string formatted in the users locale date format instead of as epoch timestamp as it was before.

## Testing requirements
Follow the steps from original bug report and verify that expected result is obtained:
In 1Password, export a vault

    Open the 1Password app (Running on MacOS)
    Select a vault
    Click on 'File' --> 'Export' --> 'All Items...'
    Enter your vaults master password
    Select file format: 1Password Interchange Format (.1pif)
    Click 'Save'

In Bitwarden, import the 1Password vault

    Log into the web vault.
    Click on 'Tools' --> 'Import Data'
    Select the format of the import file: 1Password (1pif)
    Select the import file.
    Click on 'Import Data'
    Click on any newly important item which had a date field from 1Password.

Expected Result

1Password records with a date field show the date as: 2021-11-09

After importing into Bitwarden, the dates should still be shown as 2021-11-09

(I have done this and verified that date is indeed shown correctly)

## Before you submit


linusaarnio:
@Hinton Tagging you just making sure that this PR is noticed, no stress if there just a que of reviews to be made.

Hinton:
@linusaarnio I actually haven't touched any of the importer codes (other than moving the files around). I think we could benefit from updating the `spec` to include this scenario though.

@djsmith85 I believe you have a bit better knowledge of the importers than I do.

djsmith85 (changes requested):
Hi @linusaarnio, 
thank you for your contribution. I agree with @Hinton that adding a test in `onepassword1PifImporter.spec.ts` verifying the behaviour would be great. 

Additionally, I think 1P adds dates (timestamps) in UTC and we should also add them in UTC. This would also aid, if/when we add a date field type. The clients reading that value would do the conversion. Without the timezone information, we wouldn't be able to convert it. 

In case of the birthdate field it doesn't make sense to add the time but as we convert all fields of type 'date'. I'd recommend to use `toUTCString()` (with my comment above in mind) instead of `toLocaleDateString()`

linusaarnio:
@djsmith85 Thank you for the review! 

About the timezone, I understand your thought. I had actually missed that 1P had fields that included time, I only saw fields with only date which is why this fix only included date. 

Will switch to display including time in UTC, and add a test to the spec.

linusaarnio:
@djsmith85 Date fields are now represented as UTC strings and tests for all custom fields, including the birth date as UTC,  in a 1pif file are added to the spec.

djsmith85 (approved):
Thank you @linusaarnio for your contribution. This is looking good. As we just integrated prettier into our workflow, I updated your PR to fix the merge conflicts.

@Hinton Just need a second approval

linusaarnio:
@djsmith85 Thank you for the help and feedback on this issue!

PR #572, opened 2021-12-09
Improve SSO Config validation

eliykat:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [x] Other

## Objective

See https://github.com/bitwarden/web/pull/1332.

## Code changes

Directives:
* `angular/src/directives/a11y-invalid.directive.ts`: automatically apply the `aria-invalid` attribute depending on the state of the linked `formControl`. (Used on all controls to make them accessible.)
* `angular/src/directives/input-strip-spaces.directive.ts`: automatically strip spaces from certain inputs (used on URL fields)

Validators:
* `angular/src/validators/dirty.validator.ts`: design requirements were to validate on blur, but only once the user has actually changed the field value. This is achieved by setting the form to `updateOn: 'blur'` and then running any required validator through this factory, which only runs the validator if the field is dirty.
* `angular/src/validators/requiredIf.validator.ts`: used for conditional validation, usually depending on another field.

Enums:
* `common/src/enums/ssoEnums.ts`: extract the SsoApi enums into their own class and export them so we can use them in the component. Add `None` as an option to `SsoType`.

Models:
* `common/src/models/view/ssoConfigView.ts`: this is a new view model because the form structure no longer reflected the api model in 2 main ways:
  * nesting SAML and OpenId options in their own FormGroups
  * inverting the value of `idpDisableOutboundLogoutRequests`. Design wanted it to read "allow" rather than "deny". Using separate models lets us make this change without having to change the logic all the way up into the server.
* `common/src/models/api/ssoConfigApi.ts`: add helper method to create a new api model from a view model

## Before you submit


common/src/models/api/ssoConfigApi.ts
```diff
@@ -1,40 +1,58 @@
...
+      api.idpSingleLogoutServiceUrl = view.saml.idpSingleLogoutServiceUrl;
+      api.idpArtifactResolutionServiceUrl = view.saml.idpArtifactResolutionServiceUrl;
+      api.idpX509PublicCert = view.saml.idpX509PublicCert;
+      api.idpOutboundSigningAlgorithm = view.saml.idpOutboundSigningAlgorithm;
+      api.idpAllowUnsolicitedAuthnResponse = view.saml.idpAllowUnsolicitedAuthnResponse;
+      api.idpWantAuthnRequestsSigned = view.saml.idpWantAuthnRequestsSigned;
+
+      // Value is inverted in the api model (disable instead of allow)
+      api.idpDisableOutboundLogoutRequests = !view.saml.idpAllowOutboundLogoutRequests;
+    }
```
Hinton, line 52: Would this work? We send defaults from the server which we should receive on the client to show the correct defaults.

Hinton (changes requested):

common/src/models/api/ssoConfigApi.ts
eliykat: Hmm I didn't realise that. I've provided the defaults client-side in the related `web` PR (see [here](https://github.com/bitwarden/web/pull/1332/files#diff-39683ecd9b7c1673e385026e7b610284295d88f5f776e963ba75520d308424abR95)). To me it makes more sense to do this client-side, because they are form defaults and the form belongs to the client. I also think it's required for the new design of the form (now that it only saves the selected configuration). I didn't notice any unexpected behaviour during my testing.

Note that the read-only fields (`callbackPath`, `entityId`, `metadataUrl`, etc) are still populated from the data provided by the server, they're outside the `ssoConfig` object. 

What do you think?

eliykat: Although if we do that then I guess we should remove the defaults server-side and make sure that won't affect anything else.

Hinton: I'm fine with that. Can you open a PR on the server with the removal of the defaults?

eliykat: @Hinton I came back to this and have done the following:
* server PR as discussed above: https://github.com/bitwarden/server/pull/1879
* refactored the web PR to use reusable components rather than copy/pasting throughout the template

angular/src/validators/dirty.validator.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+// Don't use the higher order function because it prevents hasError() from comparing validator functions for equality
+export function dirtyRequired(control: AbstractControl): ValidationErrors | null {
+  return control.dirty ? Validators.required(control) : null;
+}
+
+export function dirtyRequiredIf(predicate: (predicateCtrl: AbstractControl) => boolean) {
+  return dirtyValidator(requiredIf(predicate));
+}
```
Hinton, line 17: Could we add some comments describing what these validators do? I assume they require the component to be "dirty" before they trigger?

angular/src/directives/input-strip-spaces.directive.ts
```diff
@@ -0,0 +1,12 @@
+import { Directive, ElementRef, HostListener } from "@angular/core";
+
+@Directive({
+  selector: "input[appChangeStripSpaces]",
```
Hinton, line 4: The name seems a bit off, why is Change included in the directive? `appInputStripSpaces` seems cleaner, or simply `appStripSpaces`.

angular/src/directives/a11y-invalid.directive.ts
```diff
@@ -0,0 +1,26 @@
...
+      if (status === "INVALID") {
+        this.el.nativeElement.setAttribute("aria-invalid", "true");
+      } else if (status === "VALID") {
+        this.el.nativeElement.setAttribute("aria-invalid", "false");
+      }
+    });
+  }
+
+  ngOnDestroy() {
+    this.sub.unsubscribe();
```
Hinton, line 24:
```suggestion
    this.sub?.unsubscribe();
```

It's possible for the `ngOnDestroy` to be called without `ngOnInit` ever being called. Per https://www.bennadel.com/blog/3356-ngoninit-may-not-get-called-before-ngondestroy-is-called-in-angular-4-4-6.htm.

Hinton (changes requested):
Looks overall good, a few minor things. I think the validators could use a tiny bit of description to avoid confusion down the line.

angular/src/directives/input-strip-spaces.directive.ts
eliykat: Changed to `appInputStripSpaces`, that follows the convention of the other directives.

Hinton (approved):

Hinton:
We should be good code wise on this. Haven't tested the functionality.

PR #574, opened 2021-12-10
Add Prettier configuration

Hinton:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Adds the necessary configurations for configuring Prettier. And setting up a pre-commit hook for automatically running it before commits.

* Removed a couple of tslint rules that conflicts with prettier.
* Temporarily disable tslint in CI. This will be enabled again in the 2nd PR.
* Set `indent_size: 2`, since we decided to stick with prettiers defaults this should ensure our editors don't get confused. (While at it, I also added `trim_trailing_whitespace: true`
* Added huskier with a pre-commit hook relying on lint-staged. Which will ensure each commit is formatted. Note, this "should" never require manual intervention and will just make the commits easier to read.
* Installed prettier, with the only config change to set line-width to 100.

## Before you submit


.github/workflows/build.yml
```diff
@@ -51,8 +51,8 @@ jobs:
       - name: Install Node dependencies
         run: npm install
 
-      - name: Run linter
-        run: npm run lint
+      # - name: Run linter
+      #   run: npm run lint
```
djsmith85, line 55: This is temporary as we'll be adding this step again once we've migrated to ESLint?

djsmith85 (commented):
Looking good. The git-blame-ignore-revs will follow once jslib is done?

.github/workflows/build.yml
Hinton: We're adding it back in the prettier PR.

Hinton:
Yes.

djsmith85 (approved):
Looking good

PR #576, opened 2021-12-10
Move tests to correct sub-package

Hinton:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

We previously split up the jslib repo into multiple modules but the tests remained in the root project. With npm v7 having support for workspaces, I sent some time cleaning it up and moving the tests to the correct packages.

This should also allow us to properly build each package independently and use the built files vs depending on the typescript source. Which I believe will help alleviate many of the issues we've encountered.

## Before you submit


common/karma.conf.js
```diff
@@ -68,31 +78,6 @@ module.exports = (config) => {
                 reasons: true,
                 errorDetails: true,
             },
-            devtool: 'inline-source-map',
-        },
-
-        detectBrowsers: {
```
Hinton, line 74: TODO: Should bring this back

common/spec/importers/firefoxCsvImporter.spec.ts
```diff
@@ -1,8 +1,8 @@
-import { FirefoxCsvImporter as Importer } from 'jslib-common/importers/firefoxCsvImporter';
+import { FirefoxCsvImporter as Importer } from '../../src/importers/firefoxCsvImporter';
```
djsmith85, line 1: Do we have to do this? I'd prefer it staying the way it is. The clients using jslib also work with referrering to jslib-common package

Hinton: Added `src` as an alias.

MGibson1:
> 

Do you happen to have an issue/commit to link to?

MGibson1:
This was done in #744

PR #584, opened 2021-12-16
[feat(Account Switching)] Allow for extending application state

addisonbeck:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Code changes

* **file.ext:** Description of what was changed and why

## Before you submit


common/src/models/domain/globalState.ts
```diff
@@ -21,4 +21,7 @@ export class GlobalState {
   noAutoPromptBiometrics?: boolean;
   noAutoPromptBiometricsText?: string;
   stateVersion: number;
+  environmentUrls?: any = {
+    server: "bitwarden.com",
+  };
```
Hinton, line 26: Does this mean we can't have different environments per account?

addisonbeck: Temporarily. Right now setting environment urls is just plain not working, moving them out of account scope is a temporary fix until I'm back next week to resolve the underlying issue with having them be scoped to an account, which is going to require a bit of thought.

Hinton: Fair enough, just wanted to verify the purpose.

Hinton (approved):
Do we have an actual need of extending `Account`? Wondering if we could implement the same by adding another getter/setter with client specific data.

Or more generally, what additional data do we need to store on the account object?

differsthecat:
> 

We had originally implemented it that way for browser specific state items, but it seemed better to have the clients themselves handle their own state specific items. Browser and Directory Connector are the clients that have to do this.
For example, in directory connector it's items such as directory configurations that don't apply outside of that app.

PR #587, opened 2021-12-20
Use MP policies when registering a new user through SSO

jlf0dev:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Asana Task: https://app.asana.com/0/1169444489336079/1201499725597872/f
Dependency: https://github.com/bitwarden/server/pull/1779
Users currently don't have their master password checked against the organization policy while registering via SSO. This fix checks if the user is registering a new account, and if so pulls org policies from the endpoint created in the server dependency.

**Will be cherry picked to rc**

## Code changes

- **angular/src/components/change-password.component.ts:** Fill policies if user is not registering for first time
- **angular/src/components/set-password.component.ts:** If identifier is set (registering), fill policies using userId and orgId
- **common/src/services/policy.service.ts:** Add method for finding policies if userId is an invited user to organization
- **common/src/services/api.service.ts:** Call to get policies for invited users

## Testing requirements

- Test all registration methods for MP policies
    - sso
        - invitation
        - no invitation
    - non-sso
        - invitation

**Note: Existing accounts using an invitation will still not work and out of scope for this task**
See: https://app.asana.com/0/1169444489336079/1200003629652775/f

## Before you submit

### DIFF
--- a/angular/src/components/change-password.component.ts
+++ b/angular/src/components/change-password.component.ts
@@ -40,7 +40,7 @@ export class ChangePasswordComponent implements OnInit {
 
   async ngOnInit() {
     this.email = await this.stateService.getEmail();
-    this.enforcedPolicyOptions = await this.policyService.getMasterPasswordPolicyOptions();
+    this.enforcedPolicyOptions ??= await this.policyService.getMasterPasswordPolicyOptions();
   }
 
   async submit() {
--- a/angular/src/components/set-password.component.ts
+++ b/angular/src/components/set-password.component.ts
@@ -79,6 +79,7 @@ export class SetPasswordComponent extends BaseChangePasswordComponent {
         const response = await this.apiService.getOrganizationAutoEnrollStatus(this.identifier);
         this.orgId = response.id;
         this.resetPasswordAutoEnroll = response.resetPasswordEnabled;
+        this.enforcedPolicyOptions =  await this.policyService.getMasterPasswordPoliciesForInvitedUsers(this.orgId);
       } catch {
         this.platformUtilsService.showToast("error", null, this.i18nService.t("errorOccurred"));
       }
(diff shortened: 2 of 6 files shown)


angular/src/components/change-password.component.ts
```diff
@@ -40,7 +40,7 @@ export class ChangePasswordComponent implements OnInit {
 
   async ngOnInit() {
     this.email = await this.stateService.getEmail();
-    this.enforcedPolicyOptions = await this.policyService.getMasterPasswordPolicyOptions();
+    this.enforcedPolicyOptions ??= await this.policyService.getMasterPasswordPolicyOptions();
```
differsthecat, line 43: If I'm understanding this correctly, this assigns the result of `this.policyService.getMasterPasswordPolicyOptions()` to `enforcedPolicyOptions` if it is falsy, right? Wouldn't it always be falsy on `ngOnInit` if it is not passed as an input?

jlf0dev: Correct, inside of the `ngOnInit` for `set-password.component.ts` we detect if we're registering a new user into an organization and if so fill `enforcedPolicyOptions`:
```ts
// Automatic Enrollment Detection
    if (this.identifier != null) {
      try {
        const response = await this.apiService.getOrganizationAutoEnrollStatus(this.identifier);
        this.orgId = response.id;
        this.resetPasswordAutoEnroll = response.resetPasswordEnabled;
        this.enforcedPolicyOptions =
          await this.policyService.getMasterPasswordPoliciesForInvitedUsers(this.orgId);
      } catch {
        this.platformUtilsService.showToast("error", null, this.i18nService.t("errorOccurred"));
      }
    }
```

The reason I check if its falsy is to make sure it hasn't been previously loaded. Unless I'm misunderstanding the order in which those are run?

differsthecat: Ah that makes sense! I just looked and it looks like `super.ngOnInit()` is in fact called the logic in `set-password.component.ts` 👍

differsthecat (approved):

PR #589, opened 2021-12-22
Fix "no amr found" error in bwdc cli

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Logging in with the bwdc cli was throwing an error: "No amr found".

## Code changes

* **common/src/services/token.service.ts**: `getIsExternal` checks whether the token has the `external` amr claim. It appears that when logging in with org api credentials, there is no amr claim, so we should do an early return in that case without throwing the error.

## Testing requirements

* confirm user can log in and perform basic sync with bwdc cli.
* basic regression testing on an existing user being required to convert to key connector (in one of the other Angular clients, e.g. web or desktop)

## Before you submit

### DIFF
--- a/common/src/services/token.service.ts
+++ b/common/src/services/token.service.ts
@@ -209,7 +209,7 @@ export class TokenService implements TokenServiceAbstraction {
 
   async getIsExternal(): Promise<boolean> {
     const decoded = await this.decodeToken();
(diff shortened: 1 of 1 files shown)


Hinton (commented):
I wonder if we could just return false if amr is null?

eliykat:
Good point, done

Hinton (approved):

PR #593, opened 2021-12-29
[bug] Add several state value defaults

addisonbeck:
## Related Work
- [Asana Task](https://app.asana.com/0/inbox/1183359552741416/1201592404236978/1201607436480545)
- https://github.com/bitwarden/web/pull/1365

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
The storage engine refactor created a bug where dropdown setting values are not defaulting appropriately, most noticeable on web where a vault timeout of "never" should only be allowed for dev, but instead was being applied as a default for all environments. This is addressed in two parts: a jslib PR that sets defaults for locale, theme, and timeout action (as those are the same on each client) and a web PR that extends the account model and sets an appropriate default for vault timeout duration based on environment. 

I fixed one other thing while in the area: simply removing a call to set InMemory state while loading state from disk. This simply isn't needed and was added as a misunderstanding - the OnDiskInMemory options are a web specific location for secure storage, and likely needs to be removed completely. 

## Code changes
* Added defaults for theme, locale, timeout action, and state version
* Adjusted the migrator to use an initialized GlobalState instead of a hardcoded partial 
* Removed the call to save disk loaded state to the HtmlStorageServices's memory store

## Before you submit

### DIFF
--- a/common/src/models/domain/globalState.ts
+++ b/common/src/models/domain/globalState.ts
@@ -2,14 +2,14 @@ export class GlobalState {
   enableAlwaysOnTop?: boolean;
   installedVersion?: string;
   lastActive?: number;
-  locale?: string;
+  locale?: string = "en";
   openAtLogin?: boolean;
   organizationInvitation?: any;
   ssoCodeVerifier?: string;
   ssoOrganizationIdentifier?: string;
   ssoState?: string;
   rememberedEmail?: string;
-  theme?: string;
+  theme?: string = "light";
   window?: Map<string, any> = new Map<string, any>();
   twoFactorToken?: string;
   disableFavicon?: boolean;
@@ -23,7 +23,7 @@ export class GlobalState {
   biometricText?: string;
   noAutoPromptBiometrics?: boolean;
   noAutoPromptBiometricsText?: string;
-  stateVersion: number;
+  stateVersion: number = 2;
   environmentUrls?: any = {
     server: "bitwarden.com",
   };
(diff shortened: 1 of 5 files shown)


common/src/models/domain/globalState.ts
```diff
@@ -23,7 +23,7 @@ export class GlobalState {
   biometricText?: string;
   noAutoPromptBiometrics?: boolean;
   noAutoPromptBiometricsText?: string;
-  stateVersion: number;
+  stateVersion: number = 2;
```
MGibson1, line 26: Any way we can default this to latest?

MGibson1:
Do we not need settings like kdfIterations because its set on Server?

These defaults should be all set in the same location, right?

common/src/services/stateMigration.service.ts
```diff
@@ -124,12 +126,13 @@ export class StateMigrationService {
...
   }
 
   async migrate(): Promise<void> {
     let currentStateVersion =
-      (await this.storageService.get<State<Account>>("state"))?.globals?.stateVersion ?? 1;
+      (await this.storageService.get<State<Account>>("state"))?.globals?.stateVersion ??
+      StateVersion.One;
     while (currentStateVersion < this.latestVersion) {
       switch (currentStateVersion) {
         case 1:
```
MGibson1, line 138: Looks like this to line 143 needs to be rethought with the enum

djsmith85 (approved):
Looks good to me. Would only ask to address Matt's question.

addisonbeck:
Matt and I discussed this privately, it isn't in scope for this bug to modify behavior of other settings.

PR #597, opened 2022-01-05
Master password policy is not checked when accepting invite from an existing account

jlf0dev:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Asana: https://app.asana.com/0/1169444489336079/1200003629652775/f

We currently don't check master passwords against the org policy when accepting an invite from an existing account. We can't perform this server side as the password is encrypted, so this PR introduces the validation in the web client.

When a user logs in to accept an invite, we will check the password against the organization policy. If it fails, we ask the user to change their password to accept the invitation. Once the user changes it, we log out and require the user to log in once more with the changed password to accept the invitation. If the user cancels, we clear the org invite from storage and navigate them to the vault. 

## Code changes

- **angular/src/components/update-password.component.ts:** Create a new update-password component. We can't reuse the update-temp-password component because temporary password changes don't require the current password in the API call. We also perform validation to make sure the current password they enter is correct.
- Web: https://github.com/bitwarden/web/pull/1371

## Testing requirements

- Entire log in workflow should be tested to make sure the new update-password component doesn't show erroneously.
- Check the log in (and password update) workflow when accepting invite.
- Check the log in (and password update) workflow when accepting an invite for additional orgs.
- These changes shouldn't impact SSO or new user registration workflow.

## Before you submit

### DIFF
--- a/angular/src/components/update-password.component.ts
+++ b/angular/src/components/update-password.component.ts
@@ -1,0 +1,134 @@
+import { Directive } from "@angular/core";
+import { Router } from "@angular/router";
+
+import { ApiService } from "jslib-common/abstractions/api.service";
+import { CryptoService } from "jslib-common/abstractions/crypto.service";
+import { I18nService } from "jslib-common/abstractions/i18n.service";
+import { LogService } from "jslib-common/abstractions/log.service";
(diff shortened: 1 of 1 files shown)


justindbaur (approved):
LGTM

PR #600, opened 2022-01-07
Misc Account Switching Fixes & Refactors

addisonbeck:
https://app.asana.com/0/1200804338582616/1201630086148170

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
This PR addresses a couple of small issues and may be best read by going through each commit. The main things being addressed here are:

1. EnvironmentUrls need to be account specific, and currently they are global
2. Performance concerns with the state service
3. A bug in the logout flow that could leave applications in a semi-authenticated state

## Code changes
* See commits and descriptions

## Before you submit

### DIFF
--- a/common/src/models/domain/environmentUrls.ts
+++ b/common/src/models/domain/environmentUrls.ts
@@ -2,5 +2,9 @@ export class EnvironmentUrls {
   base: string;
   api: string;
   identity: string;
+  icons: string;
+  notifications: string;
   events: string;
+  webVault: string;
+  keyConnector: string;
 }
--- a/common/src/services/state.service.ts
+++ b/common/src/services/state.service.ts
@@ -101,6 +101,7 @@ export class StateService<TAccount extends Account = Account>
 
   async clean(options?: StorageOptions): Promise<void> {
     // Find and set the next active user if any exists
+    await this.setAccessToken(null, { userId: options?.userId });
     if (options?.userId == null || options.userId === (await this.getUserId())) {
       for (const userId in this.state.accounts) {
         if (userId == null) {
--- a/common/src/services/state.service.ts
+++ b/common/src/services/state.service.ts
@@ -2184,7 +2184,6 @@ export class StateService<TAccount extends Account = Account>
     state.accounts[account.profile.userId] = account;
 
     await storageLocation.save("state", state, options);
-    await this.pushAccounts();
   }
 
   protected async saveAccountToMemory(account: TAccount): Promise<void> {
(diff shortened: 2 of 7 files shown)


common/src/services/state.service.ts
```diff
@@ -1359,24 +1362,16 @@ export class StateService<TAccount extends Account = Account>
...
-        keyConnector: null,
-        // TODO: this is a bug and we should use base instead for the server detail in the account switcher, otherwise self hosted urls will not show correctly
-        server: "bitwarden.com",
-      }
-    );
+    options = this.reconcileOptions(options, await this.defaultOnDiskOptions());
+    if (this.state.activeUserId == null) {
+      return (await this.getGlobals(options)).environmentUrls ?? new EnvironmentUrls();
+    }
+    return (await this.getAccount(options))?.settings?.environmentUrls ?? new EnvironmentUrls();
```
differsthecat, line 1369: Cool, this looks like it handles the case where directory connector has you set environment urls before logging in.

common/src/services/state.service.ts
```diff
@@ -2283,7 +2277,7 @@ export class StateService<TAccount extends Account = Account>
     return {
       storageLocation: StorageLocation.Disk,
       htmlStorageLocation: HtmlStorageLocation.Session,
-      userId: await this.getActiveUserIdFromStorage(),
+      userId: this.state.activeUserId ?? (await this.getActiveUserIdFromStorage()),
```
differsthecat, line 2280: This should cut down on the unnecessary calls to get the userId from storage every time a setting is loaded. Nice!

common/src/models/domain/environmentUrls.ts
```diff
@@ -2,5 +2,9 @@ export class EnvironmentUrls {
   base: string;
   api: string;
   identity: string;
+  icons: string;
+  notifications: string;
   events: string;
+  webVault: string;
+  keyConnector: string;
```
differsthecat, line 9: I'm guessing these are used individually by the clients such as desktop?

differsthecat (approved):

common/src/models/domain/environmentUrls.ts
addisonbeck: Most of these are used in all clients, we have just been using an `any` type for handling urls instead of this model. I tried to refactor all that to use a model, and discovered that we already had one that wasn't being used and just brought it up to speed.

common/src/services/state.service.ts
addisonbeck: Yep! Desktop works that way too, so you shouldn't need any special stuff to make this work.

common/src/models/domain/environmentUrls.ts
differsthecat: Got it, that makes sense! I'm definitely a fan a making those more strongly typed

PR #603, opened 2022-01-11
[Account Switching] Fix options dropdowns being empty

differsthecat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix for some dropdowns defaulting to empty.
Fixes: https://app.asana.com/0/1153292148278596/1201632897395074/f

## Code changes

The reason this is an issue is because the values are coming back as undefined in state service. Angular is doing a strict check to determine the selected option, so the value of null isn't mapping as expected. 

We could also approach this by changing the the default options in the components to be selected when they are null or undefined, but I observed this as an issue in Browser as well and doing it this way fixes it there too without having to manually update the select options in the components there as well. I waffled on which was the better approach, so I went with this way as it fixes it from the deepest level and will fix it for Browser when jslib is updated :) 

- **common/src/services/state.service.ts:** Return null explicitly for theme, clearClipboard, and locale if the values are undefined.

## Testing requirements

Ensure that the dropdowns for new accounts have options for clearClipboard, theme, and local selected as defaults.

## Before you submit

### DIFF
--- a/common/src/services/state.service.ts
+++ b/common/src/services/state.service.ts
@@ -355,8 +355,12 @@ export class StateService<TAccount extends Account = Account>
 
   async getClearClipboard(options?: StorageOptions): Promise<number> {
     return (
-      await this.getAccount(this.reconcileOptions(options, await this.defaultOnDiskLocalOptions()))
-    )?.settings?.clearClipboard ?? null;
+      (
+        await this.getAccount(
+          this.reconcileOptions(options, await this.defaultOnDiskLocalOptions())
+        )
+      )?.settings?.clearClipboard ?? null
+    );
   }
 
   async setClearClipboard(value: number, options?: StorageOptions): Promise<void> {
@@ -1603,8 +1607,12 @@ export class StateService<TAccount extends Account = Account>
 
   async getLocale(options?: StorageOptions): Promise<string> {
     return (
-      await this.getGlobals(this.reconcileOptions(options, await this.defaultOnDiskLocalOptions()))
-    )?.locale ?? null;
+      (
+        await this.getGlobals(
+          this.reconcileOptions(options, await this.defaultOnDiskLocalOptions())
+        )
+      )?.locale ?? null
+    );
   }
 
   async setLocale(value: string, options?: StorageOptions): Promise<void> {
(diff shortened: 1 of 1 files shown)


common/src/services/state.service.ts
```diff
@@ -1603,8 +1607,12 @@ export class StateService<TAccount extends Account = Account>
 
   async getLocale(options?: StorageOptions): Promise<string> {
     return (
-      await this.getGlobals(this.reconcileOptions(options, await this.defaultOnDiskLocalOptions()))
-    )?.locale;
+      (
+        await this.getGlobals(
+          this.reconcileOptions(options, await this.defaultOnDiskLocalOptions())
+        )
+      )?.locale ?? null
```
djsmith85, line 1614: Based on this https://github.com/bitwarden/jslib/blob/2341b1907ac601f70842e6a405cd641bf2706c06/common/src/models/domain/globalState.ts#L8 I'd expect it to default to `en`

common/src/services/state.service.ts
```diff
@@ -1945,8 +1953,12 @@ export class StateService<TAccount extends Account = Account>
...
   async getTheme(options?: StorageOptions): Promise<string> {
     return (
-      await this.getGlobals(this.reconcileOptions(options, await this.defaultOnDiskLocalOptions()))
-    )?.theme;
+      (
+        await this.getGlobals(
+          this.reconcileOptions(options, await this.defaultOnDiskLocalOptions())
+        )
+      )?.theme ?? null
+    );
```
djsmith85, line 1961: Same here: https://github.com/bitwarden/jslib/blob/2341b1907ac601f70842e6a405cd641bf2706c06/common/src/models/domain/globalState.ts#L15

djsmith85 (changes requested):
Just two things and I'm wondering why it doesn't return a new GlobalState object with those defaults set. Maybe it's another issue that the initial global state wasn't saved properly

common/src/services/state.service.ts
differsthecat: That's a good point, I'll look into locale and theme. This might just be because Desktop is behind on jslib at the moment and it may have been fixed already.

common/src/services/state.service.ts
differsthecat: Same comment as above

common/src/services/state.service.ts
differsthecat: I just checked and you are right, the global state does have defaults for those two now. The jslib version that desktop has now doesn't have that change, but Addison has a PR out now that brings that in
https://github.com/bitwarden/desktop/pull/1223

I'll remove the changes for the theme and locale, but we're still gonna need the change for clearClipboard

differsthecat: @djsmith85 I removed those and this should be good to go now, thank you!

djsmith85 (approved):
Awesome, good catch and ready to merge

PR #605, opened 2022-01-12
Make lowdbStorageService wait until initialised

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Clients that use `lowdbStorageService` (CLI and bwdc CLI) need that service initialized before it can be used.

#600 added a subscription to the `environmentService` constructor, which indirectly calls `storageService` before it's initialized, resulting in the following error:

```
 /snapshot/directory-connector/build-cli/bwdc.js:1513
            this.db.read();
                    ^

TypeError: Cannot read properties of undefined (reading 'read')
```

I encountered this when trying to bump jslib in https://github.com/bitwarden/directory-connector/pull/204.

No other clients need to initialize their storage services, so this is unique to `lowdbStorageService`.

## Code changes

One solution would be to move the subscription out of the `environmentService` constructor and into its own `init` method. However, that would require every client to call `init()`, even though this is only really required for the CLI clients. Really this is a quirk of `lowdbStorageService` and should ideally be handled within that implementation.

Therefore, the solution is:
* each method of `lowdbStorageService` checks whether the service has been initialized before running. If the service is not initialized, it'll initialize itself
* add an early check on `init` so it can only be run once
* add the `sequentialize()` decorator to the `init` function. This decorator prevents multiple simultaneous calls to an async method (in case it's called a second time before it's set the `ready` flag).

## Testing requirements

None required, beyond basic/usual regression testing in CLI and bwdc CLI.

## Before you submit

### DIFF
--- a/node/src/services/lowdbStorage.service.ts
+++ b/node/src/services/lowdbStorage.service.ts
@@ -25,7 +25,7 @@ export class LowdbStorageService implements StorageService {
     this.defaults = defaults;
   }
 
-  @sequentialize(() => 'lowdbStorageInit')
+  @sequentialize(() => "lowdbStorageInit")
   async init() {
     if (this.ready) {
       return;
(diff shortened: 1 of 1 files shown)


differsthecat (approved):
Good catch on this, looks good!

PR #608, opened 2022-01-14
BEEEP: Add importer for Keeper in json format

djsmith85:
## Type of change

- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective
To aid new users in migrating to Bitwarden we offer a wide variety of importers. This PR adds importing `.json` exports from Keeper, providing our new users a really smooth transition into Bitwarden.

The importer supports creating Login types with the following fields:
- Title
- Username
- Password
- URIs
- TOTP
- Notes
- Folders/Collections (Only creates collections when an import is done from within an organization)
- and creates custom fields for any fields in the custom_fields that could not be mapped

**The permissions present in the shared_folders are not imported. These will have to be set once the data has been imported.**

For more information on exporting from Keeper: https://docs.keeper.io/user-guides/export-and-reports/vault-export
and for information on the json export: https://docs.keeper.io/user-guides/import-records-1/import-json#creating-a-json-file-from-keeper-commander

Asana task: https://app.asana.com/0/1200804338582616/1201664207082273/f
## Code changes
- **spec/common/importers/testData/keeperJson/testData.ts:** Added testdata which I pulled from keeper's website linked above.
- **common/src/importers/keeperImporters/types/keeperJsonTypes.ts:** Added types to describe the exported json file
- **common/src/importers/keeperImporters/keeperJsonImporter.ts:** Created the actual importer to handle json exports from Keeper
- **spec/common/importers/keeperJsonImporter.spec.ts:** Several unit tests to test the behavior of the importer
- **common/src/importers/keeperCsvImporter.ts:** Moved to `common/src/importers/keeperImporters/keeperCsvImporter.ts`
- **common/src/services/import.service.ts:** Registered new KeeperJsonImporter and updated import for moved KeeperCsvImporter

## Testing requirements
The pre-existing importer for csv-files from Keeper should still work
All of the above mentioned mappings should work when importing a json export from Keeper into Bitwarden.

## Before you submit

- [X] I have added **unit tests** where it makes sense to do so (encouraged but not required)

### DIFF
--- a/common/src/importers/keeperImporters/keeperJsonImporter.ts
+++ b/common/src/importers/keeperImporters/keeperJsonImporter.ts
@@ -1,0 +1,68 @@
+import { BaseImporter } from ".././baseImporter";
+import { Importer } from ".././importer";
+
+import { ImportResult } from "../../models/domain/importResult";
+
+import { KeeperJsonExport, RecordsEntity } from "./types/keeperJsonTypes";
+
+export class KeeperJsonImporter extends BaseImporter implements Importer {
+  parse(data: string): Promise<ImportResult> {
+    const result = new ImportResult();
+    const keeperExport: KeeperJsonExport = JSON.parse(data);
+    if (keeperExport == null || keeperExport.records == null || keeperExport.records.length === 0) {
+      result.success = false;
+      return Promise.resolve(result);
+    }
+
+    keeperExport.records.forEach((record) => {
+      this.parseFolders(result, record);
+
+      const cipher = this.initLoginCipher();
+      cipher.name = record.title;
+      cipher.login.username = record.login;
+      cipher.login.password = record.password;
+
+      cipher.login.uris = this.makeUriArray(record.login_url);
+      cipher.notes = record.notes;
+
+      let customfieldKeys = Object.keys(record.custom_fields);
+      if (customfieldKeys.includes("TFC:Keeper") && record.custom_fields["TFC:Keeper"] != null) {
+        customfieldKeys = customfieldKeys.filter((item) => item !== "TFC:Keeper");
+        cipher.login.totp = record.custom_fields["TFC:Keeper"];
+      }
+
+      customfieldKeys.forEach((key) => {
+        this.processKvp(cipher, key, record.custom_fields[key]);
+      });
+
+      this.convertToNoteIfNeeded(cipher);
+      this.cleanupCipher(cipher);
+      result.ciphers.push(cipher);
+    });
+
+    if (this.organization) {
+      this.moveFoldersToCollections(result);
+    }
+
+    result.success = true;
+    return Promise.resolve(result);
+  }
+
+  private parseFolders(result: ImportResult, record: RecordsEntity) {
+    if (record.folders == null || record.folders.length === 0) {
+      return;
+    }
+
+    record.folders.forEach((item) => {
+      if (item.folder != null) {
+        this.processFolder(result, item.folder);
+        return;
+      }
+
+      if (item.shared_folder != null) {
+        this.processFolder(result, item.shared_folder);
+        return;
+      }
+    });
+  }
+}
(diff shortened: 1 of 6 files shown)


common/src/importers/keeperImporters/keeperJsonImporter.ts
```diff
@@ -0,0 +1,68 @@
+import { BaseImporter } from ".././baseImporter";
```
eliykat, line 1: What's with the `/./`, shouldn't it just be `../baseImporter`?

common/src/importers/keeperImporters/keeperJsonImporter.ts
```diff
@@ -0,0 +1,68 @@
...
+      const cipher = this.initLoginCipher();
+      cipher.name = record.title;
+      cipher.login.username = record.login;
+      cipher.login.password = record.password;
+
+      cipher.login.uris = this.makeUriArray(record.login_url);
+      cipher.notes = record.notes;
+
+      let customfieldKeys = Object.keys(record.custom_fields);
+      if (customfieldKeys.includes("TFC:Keeper") && record.custom_fields["TFC:Keeper"] != null) {
```
eliykat, line 29: The first part of this condition seems unnecessary, if there's no key then `record.custom_fields["TFC:Keeper"]` will always return `null` anyway.

spec/common/importers/keeperJsonImporter.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import { KeeperJsonImporter as Importer } from "jslib-common/importers/keeperImporters/keeperJsonImporter";
+
+import { testData as TestData } from "./testData/keeperJson/testData";
+
+describe("Keeper Json Importer", () => {
+  const testDataJson = JSON.stringify(TestData);
+
+  it("should parse login data", async () => {
+    const importer = new Importer();
```
eliykat, line 11: These 3 lines of setup are used in each test. You should extract them to a `beforeEach` function: https://jasmine.github.io/api/3.9/global.html#beforeEach

Or you may have to use a (non-Jasmine) setup function that takes the orgId as an argument, to accommodate the last test.

eliykat (changes requested):
Nice work!

common/src/importers/keeperImporters/keeperJsonImporter.ts
djsmith85: Fixed

common/src/importers/keeperImporters/keeperJsonImporter.ts
djsmith85: That is true, removed the unnecessary check for the key

spec/common/importers/keeperJsonImporter.spec.ts
djsmith85: The testDataJson is only stringified once. I've moved the instantiation of the importer into a beforeEach but kept the call to `parse` in the tests as that is not technically setup but the Act of AAA. Also solves the issue with the OrgId on the last test.

eliykat: Personally I think it's okay to have that basic assertion in a common function, it's almost part of the setup before continuing with the 'real' test. But this does solve the orgId problem, so that's fine.

eliykat (changes requested):
Just one more potential issue that I spotted

common/src/importers/keeperImporters/keeperJsonImporter.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      const cipher = this.initLoginCipher();
+      cipher.name = record.title;
+      cipher.login.username = record.login;
+      cipher.login.password = record.password;
+
+      cipher.login.uris = this.makeUriArray(record.login_url);
+      cipher.notes = record.notes;
+
+      let customfieldKeys = Object.keys(record.custom_fields);
```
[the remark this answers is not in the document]
djsmith85: Per the docs I can't tell if the property will be left off when empty. I'll try and get a new export without custom fields and see how Keeper exports it. If it can I'll also adjust the types. Good catch!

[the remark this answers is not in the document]
djsmith85: I got a new export fresh from Keeper and you were right, some fields were optional. Adjusted the types, added a test case and the fix for missing custom fields.

eliykat (approved):

PR #610, opened 2022-01-14
Add eslint

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Initial work at replacing tslint with eslint. I went with the default eslint rules and added some of the old rules as well. I didn't add `member-ordering`, I'm open to adding it, but we disable it in a few places already since js needs to be organized a specific way, so unsure how much value it would provide.

I mainly preferred to ignore eslint errors vs fixing them. We should remove any comments when we modify nearby code but it seemed a waste of time to test everything now.

Biggest change to before is that imports are now automatically sorted and grouped using eslint. This results in a slightly changed grouping than we are used to. I think the benefits of automating it is an improvement.

## Before you submit


Hinton:
Added hold to make sure we don't merge this before `rc` is cut.

common/src/models/request/tokenRequest.ts
```diff
@@ -0,0 +1,90 @@
+import { TwoFactorProviderType } from "../../enums/twoFactorProviderType";
+import { Utils } from "../../misc/utils";
+
+import { CaptchaProtectedRequest } from "./captchaProtectedRequest";
+import { DeviceRequest } from "./deviceRequest";
+
+export class TokenRequest implements CaptchaProtectedRequest {
```
jlf0dev, line 7: Curious about this added file?

Hinton: Probably a file that was removed in a PR and got accidently re-added.

jlf0dev (approved):

jlf0dev:
I found one missing reference to `tslint` in `bitwardenpasswordprotectedimporter.ts` in your latest merge from master.

Other than that looks good!

jlf0dev (approved):

PR #611, opened 2022-01-14
[Bug] [Account Switching] Improve State Management Performance

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Large vaults see a clear degrade in performance using the state service, especially when multiple vaults are authed and unlocked at the same time.

## Code changes
Some changes made to address this:
1. Clearing in memory decrypted data for non active accounts. This really should have been something we were doing anyway, but letting go of that memory burden has a noticeable performance boost.
2. Not loading a bunch of unnecessary data from disk accounts into memory on application startup. This was being done to initialize in memory accounts, but brought a lot of extra baggage with it like storing encrypted data in memory, even though it is never referenced that way.
3. Breaking the on disk state object up into separate keys for accounts instead of storing everything together under a "state" key. This ensures there is less information fetched from disk each time we call for an account.

There were some restructuring changes needed to facilitate these items:
1. We need to be able to construct an account in the StateService, but typescript doesn't allow for new() constraints on generics so a factory needs to be created and passed into the StateService for this to work.
2. Since we can't reference an all-knowing "accounts" object for on disk state anymore we have to maintain a list of authenticated accounts, and this has been added.
3. The StateMigration service needed to be updated to break up the accounts object, so current dev and QA state will be broken and need to be reset.

Some other general refactorings that were helpful getting this working:
1. Added a constant for keys to the StateService and StateMigrationService.
2. Bundling everything needed to deauthenticate a user into a dedicated method.
3. Bundling all the disk storage clear methods (that should be refactored later into client specific state services) into one helper method.
4. Bundling everything needed to dynamically select a new active user into a dedicated method.

## Testing requirements
This is going to break local storage for QA and anything currently running the previous state changes where everything was saved under a "state" key will need to be cleared.

## Before you submit

### DIFF
--- a/angular/src/services/jslib-services.module.ts
+++ b/angular/src/services/jslib-services.module.ts
@@ -76,6 +76,8 @@ import { PasswordRepromptService } from "./passwordReprompt.service";
 import { UnauthGuardService } from "./unauth-guard.service";
 import { ValidationService } from "./validation.service";
 
+import { Account, AccountFactory } from "jslib-common/models/domain/account";
+
 @NgModule({
   declarations: [],
   providers: [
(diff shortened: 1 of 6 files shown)


djsmith85 (approved):
Code looks good, and already tested these changes combined with changes done on https://github.com/bitwarden/desktop/pull/1237

PR #612, opened 2022-01-14
Feature/password protected export

MGibson1:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Add a new Bitwarden Import/Export that is encrypted with a user-supplied password. This should enable secure storage of Bitwarden vaults without the worry that Bitwarden instance data-loss or key rotations will impact your ability to decrypt!

CLI PR is incoming, but we're still working on exactly how we want to present this in Web.

## Code changes
Largely this is just a new importer and a wrapper function for exports, but further breakdown below.

* **export.service**: `getPasswordProtectedExport` calls an inner exporter depending on format and organizationId, then encrypts the result with the given password.
* **import.service**: I've extracted the ImportTypes to constants so that I can generate union type strings for each importer's Id. That way we can't give invalid Ids to the import service. Types are good 🥳 
  * Type safety changes: There will be small changes in `web` and `cli` due to the change in type of the import options to `readonly`. Basically we can't `sort` in place or pass in `string` formats anymore.
* **bitwardenPasswordProtectedImporter**: Handles decryption of a password protected export file and passes the decrypted inner data to an inner importer. Uses the same mac validation scheme we use for `encrypted_json` files, where we write a short string with a mac and validate we were able to decrypt it.

### Test coverage
* password protected importer 
  * We validate that the imported json doc has all the fields we need in the correct types.
  * We validate we're grabbing the right inner importers
* import service
  * Validates we get the right kind of importer for the appropriate importer type request
  * validates we're passing on organizationId and password appropriately
* Export service
  * Validates we're password protecting the export
  * Validates we're adding all the required fields property to the export json

We'll need integration tests that validate round-tripped password protected exports.

## Before you submit

- [x] I have added **unit tests** where it makes sense to do so (encouraged but not required)
- [x] This change requires a **documentation update** (notify the documentation team)

### DIFF
--- a/common/src/importers/bitwardenPasswordProtectedImporter.ts
+++ b/common/src/importers/bitwardenPasswordProtectedImporter.ts
@@ -1,0 +1,90 @@
+import { BaseImporter } from './baseImporter';
+import { Importer } from './importer';
+
+import { EncString } from '../models/domain/encString';
+import { ImportResult } from '../models/domain/importResult';
+
+import { CipherWithIds } from '../models/export/cipherWithIds';
+import { CollectionWithId } from '../models/export/collectionWithId';
+import { FolderWithId } from '../models/export/folderWithId';
+
+import { CryptoService } from '../abstractions/crypto.service';
+import { I18nService } from '../abstractions/i18n.service';
+import { CryptoFunctionService } from '../abstractions/cryptoFunction.service';
+import { BitwardenCsvImporter } from './bitwardenCsvImporter';
+import { BitwardenJsonImporter } from './bitwardenJsonImporter';
+import { KdfType } from '../enums/kdfType';
+import { SymmetricCryptoKey } from '../models/domain/symmetricCryptoKey';
+
+class BitwardenPasswordProtectedFileFormat {
+    encrypted: boolean;
+    passwordProtected: boolean;
+    format: 'json' | 'csv' | 'encrypted_json';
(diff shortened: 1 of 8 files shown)


MGibson1:
FYI @fschillingeriv

common/src/importers/bitwardenPasswordProtectedImporter.ts
```diff
@@ -0,0 +1,98 @@
...
+import { CryptoService } from "../abstractions/crypto.service";
+import { I18nService } from "../abstractions/i18n.service";
+import { ImportService } from "../abstractions/import.service";
+import { KdfType } from "../enums/kdfType";
+import { SymmetricCryptoKey } from "../models/domain/symmetricCryptoKey";
+
+class BitwardenPasswordProtectedFileFormat {
+  encrypted: boolean;
+  passwordProtected: boolean;
+  format: "json" | "csv" | "encrypted_json";
```
kspearrin, line 16: Might want to add KDF type here too.

MGibson1: Good call, done

differsthecat (commented):
This is awesome! Looks like there's a conflict; once that is resolved, let me know I can look over again and ✅  if all is well

differsthecat (approved):
Looks great!

PR #614, opened 2022-01-18
Always migrate envUrls to new storage format

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix the following defect in CLI: 
1. run a pre-account switching version of CLI
2. set your environment urls using `bw config server`
3. log out
4. upgrade CLI and try to log in

Expected result: saved urls are migrated
Actual result: saved urls are lost and it defaults to production servers

## Code changes

* `stateMigrationService.ts` - always migrate the environment urls to `global` settings, regardless of whether the user is logged in (which I believe is what the `userId` storage key indicates). This seems like a reasonable course of action for all clients.

Incidentally, it appears that the environment urls are copied to the user's settings when the migration runs, but after that point they're only read from and written to the global settings. This seems like it could cause problems down the line because we'll have unused values sitting in user's settings, but we won't know whether they're valid or not. However, I haven't been doing account switching work, so I haven't changed anything here yet without some feedback from the team.

## Testing requirements

Will be covered in general account switching regression testing. But note that the migration path should be tested specifically.

## Before you submit

### DIFF
--- a/common/src/services/stateMigration.service.ts
+++ b/common/src/services/stateMigration.service.ts
@@ -177,9 +177,7 @@ export class StateMigrationService {
                 v1Keys.enableBiometric,
                 options
               ),
-              environmentUrls:
-                (await this.storageService.get<EnvironmentUrls>(v1Keys.environmentUrls, options)) ??
-                new EnvironmentUrls(),
+              environmentUrls: null,
               installedVersion: await this.storageService.get<string>(
                 v1Keys.installedVersion,
                 options
@@ -488,6 +486,10 @@ export class StateMigrationService {
               }),
             },
           };
+
+    initialState.globals.environmentUrls =
+      (await this.storageService.get<EnvironmentUrls>(v1Keys.environmentUrls, options)) ??
+      new EnvironmentUrls();
 
     await this.storageService.save("state", initialState, options);
 


common/src/services/stateMigration.service.ts
```diff
@@ -177,9 +177,7 @@ export class StateMigrationService {
                 v1Keys.enableBiometric,
                 options
               ),
-              environmentUrls:
-                (await this.storageService.get<EnvironmentUrls>(v1Keys.environmentUrls, options)) ??
-                new EnvironmentUrls(),
+              environmentUrls: null,
```
differsthecat, line 180: I'm not sure we want to remove the environment urls from the account level during these migrations. They might be used by desktop, web, and browser from here after the initial migration.

The global environmentUrls were created when we discovered that the directory connector allows environmentUrls to be set before login. We added a step in the stateMigrationService in the directory connector to migrate the environmentUrls globally as you did below, but we failed to consider that the cli may have the same issue.

eliykat: This change is in the `globals`, not the account level. I just removed it to avoid duplication given that we set it below. But I can't remove the key entirely due to type checking.

differsthecat: Ah got it! I misread where it was!

differsthecat (approved):
Looks good!

PR #615, opened 2022-01-19
[bug] Migrate state even if there is not a user logged in

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Currently the StateMigrationService depends on a userId key for running migrations, but if there is not an authenticated user saved to storage that userId is not present and settings are not migrated.

## Code changes
These changes allow for migrating state data even without an active user. For account specific settings like clearClipboard we now temporarily store those values together in disk state until an account is authed that they can be added to. Temp account state is then cleared.

Some notes:
* In order for this to work we need GlobalState.stateVersion to have a default value of StateVersion.One instead of StateVersion.Latest. Defaulting it to latest was causing migrations to not run on some clients (like desktop) that try to access storage before migrations have been run but save a version as if migrations did run.
* I also noticed we aren't clearing old state items from before migrating, and added a case for this to the migrator.
* I extracted a few bits of reused code into private methods in the stateMigration service. Things like get/set from storage, default options, etc.

## Testing requirements
Testing this would involve testing the migrator and ensuring all data is appropriately moved and cleared without breaks.

## Before you submit

### DIFF
--- a/common/src/services/stateMigration.service.ts
+++ b/common/src/services/stateMigration.service.ts
@@ -1,9 +1,7 @@
 import { StorageService } from "../abstractions/storage.service";
 
-import { Account } from "../models/domain/account";
 import { GeneratedPasswordHistory } from "../models/domain/generatedPasswordHistory";
 import { GlobalState } from "../models/domain/globalState";
-import { State } from "../models/domain/state";
 import { StorageOptions } from "../models/domain/storageOptions";
 
 import { CipherData } from "../models/data/cipherData";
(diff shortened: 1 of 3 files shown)


common/src/services/stateMigration.service.ts
```diff
@@ -132,17 +132,12 @@ export class StateMigrationService {
   ) {}
 
   async needsMigration(): Promise<boolean> {
-    const currentStateVersion = (
-      await this.storageService.get<GlobalState>(keys.global, {
-        htmlStorageLocation: HtmlStorageLocation.Local,
-      })
-    )?.stateVersion;
+    const currentStateVersion = await this.getCurrentStateVersion();
```
djsmith85, line 135: This call no longer passes the StorageOptions to retrieve GlobalState. Can this be omitted?

addisonbeck: This ends up back in a this.get() call that handles passing in correct options.

djsmith85: Ah I see, missed that 😄

djsmith85 (approved):

PR #616, opened 2022-01-20
Fix migration to Key Connector in cli commands

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

See https://github.com/bitwarden/cli/pull/452.
## Code changes

- **common/src/services/sync.service.ts** - set the KC conversion flag in sync service. This is done by all clients anyway, but needs to be done here for cli because it doesn't have a messaging service. I'll go through and remove the duplicate logic in each client after merging.
- **node/src/cli/commands/login.command.ts** - remove KC conversion check and syncing.

Other minor fixes:
- **common/src/services/keyConnector.service.ts**
  - add missing await (whoops!)
  - move the request model constructor call out of the try/catch block, it was giving me misleading error messages otherwise

**directory-connector will need a jslib update after this is merged.**

## Testing requirements

See linked cli ticket.

## Before you submit

### DIFF
--- a/common/src/services/sync.service.ts
+++ b/common/src/services/sync.service.ts
@@ -334,6 +334,7 @@ export class SyncService implements SyncServiceAbstraction {
     ]);
 
     if (await this.keyConnectorService.userNeedsMigration()) {
+      await this.keyConnectorService.setConvertAccountRequired(true);
       this.messagingService.send("convertAccountToKeyConnector");
     } else {
       this.keyConnectorService.removeConvertAccountRequired();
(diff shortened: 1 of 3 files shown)


common/src/services/sync.service.ts
```diff
@@ -334,6 +334,7 @@ export class SyncService implements SyncServiceAbstraction {
     ]);
 
     if (await this.keyConnectorService.userNeedsMigration()) {
+      await this.keyConnectorService.setConvertAccountRequired(true);
```
Hinton, line 337: Does this work correctly with the other clients? Do we need to change anything there? I think they had the check in the handler for the message.

Hinton (approved):
Please make sure QA regression test the convert flows on each client :).

common/src/services/sync.service.ts
eliykat: We already discussed, but for posterity:

All the other clients call setConvertAccountRequired from the message handler. Shouldn’t make any difference setting that flag before we send the message. After these changes, setConvertAccountRequired will be called from syncService, and the message handler will only handle the redirect.

PR #625, opened 2022-01-21
Component Library scaffolding

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Adds the scaffolding for the Component Library. It's packaged as an Angular cli application to provide support for storybook. In the long term we should look into how we can run storybook outside jslib since it's currently limited to components within jslib, which is a bit limited.

Tailwind is configured to use the prefix `tw-` to avoid clashing with bootstrap.

This project is setup slightly different to most of our projects in that it has `index.ts` which exposes a public api. This is intentional as a way to define a public api.

Resolves:
- CL-8
- CL-9
- CL-12

## Before you submit


components/src/styles.css
```diff
@@ -0,0 +1,3 @@
+/* You can add global styles to this file, and also import other style files */
+
+@import "../../../src/scss/tailwind.css";
```
Hinton, line 3: This means that this package will only work in web for now. I think that's an acceptable limitation vs having to maintain two styles + tailwind configs.

danielleflinn (approved):
Visual reviewed via Chromatic and Storybook; looks good to me.

vincentsalucci (approved):

vincentsalucci (approved):

PR #638, opened 2022-01-27
[bug] Properly define stored window state

addisonbeck:
https://app.asana.com/0/1201648796371593/1201724590570903

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Bug
Electron windows do not spawn on second launch of Bitwarden. They open fine the first time, because we haven't migrated window information yet, but a bug in the window management logic causes second launches to not spawn a window.

## Code changes
1. Properly define what we store for persistent window management in the form of WindowState(), and make that the type used for getters/setters for Window in StateService()
2. Stop searching for a configKey in Window when we get it in window.main. This is where the break happens: because the key we are looking for doesn't exist in the Window object - it IS the window object. Simply using this value without checking for a key corrects window spawning behavior. 

## Testing requirements
I tested this by building mac universal apps locally, we will need to check Testflight once a build is available there. 

## Before you submit

### DIFF
--- a/common/src/models/domain/windowState.ts
+++ b/common/src/models/domain/windowState.ts
@@ -1,0 +1,10 @@
+export class WindowState {
+  width?: number;
+  height?: number;
+  isMaximized?: boolean;
+  // TODO: displayBounds is an Electron.Rectangle.
+  // We need to establish some kind of client-specific global state, similiar to the way we already extend a base Account.
+  displayBounds: any;
+  x?: number;
+  y?: number;
+}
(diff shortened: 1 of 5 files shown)


common/src/models/domain/windowState.ts
```diff
@@ -0,0 +1,10 @@
+export class WindowState {
+  width?: number;
+  height?: number;
+  isMaximized?: boolean;
+  // TODO: displayBounds is an Electron.Rectangle.
+  // We need to establish some kind of client-specific global state, similiar to the way we already extend a base Account.
```
addisonbeck, line 6: Tech debt item for this: https://app.asana.com/0/1201461753012676/1201724590570905

djsmith85 (approved):

PR #641, opened 2022-01-27
[bug] Always save environmentUrls passed to setUrls

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
With the change to our state model to have a concept of an "account" we started saving global environmentUrls to a temporary location in storage to be fetched and added to any newly authed accounts

This isn't compatible with the option to not save envUrls passed to setUrls to storage, since we need that storage value to handle urls for authed accounts. This is only applicable to web.

There was some discussion around maybe extending the environmentService in web to skip the temporary storage route, since web isn't planned to ever offer account switching, but that would also require a distinction in web specifically that environmentUrls aren't an account level data point, which isn't really true. Ultimately I think just removing this option is the clearest way to make this work.

## Code changes
* Remove the flag from setUrls that allowed for skipping saving urls to storage

## Before you submit

### DIFF
--- a/common/src/abstractions/environment.service.ts
+++ b/common/src/abstractions/environment.service.ts
@@ -29,6 +29,6 @@ export abstract class EnvironmentService {
   getEventsUrl: () => string;
   getKeyConnectorUrl: () => string;
   setUrlsFromStorage: () => Promise<void>;
-  setUrls: (urls: any, saveSettings?: boolean) => Promise<Urls>;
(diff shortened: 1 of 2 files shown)


Hinton (approved):

djsmith85 (changes requested):
The method signature change is missing on the abstraction `common\src\abstractions\environment.service.ts`

djsmith85 (approved):

PR #642, opened 2022-01-27
[refactor] Use ThemeType enum instead of string

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
A good refactor point was made that we can use the ThemeType enum instead of strings when managing theme in state. 

## Code changes
Convert type used for theme in the state model, service, and migration

## Before you submit


eliykat (approved):

djsmith85 (approved):
Verified with Desktop that the migration is run correctly. (Prod client, set theme to Nord and then start up current master. Theme is set correctly to Nord)

PR #646, opened 2022-01-31
[bug] Allow for GlobalState to be extended and modified in clients

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Some clients have unique global setting defaults (and unique global settings)
For example: the web vault defaults to light theme, but most clients with theme support default to system theme.

The current way we handle `GlobalState` is buried in jslib and not easily extendible in clients.

To fix this, we need to treat `GlobalState` as a generic in the `StateService` and `StateMigration` service and allow for its extension in those methods and anywhere `GlobalState` is inited. We already use this pattern for extending the `Account` object into clients.

## Code changes
* Set the default `ThemeType` to `System`, since that is the standard among clients.
* Move `AccountFactory` into a folder dedicated to factories
* Add a `GlobalStateFactory`, and a `StateFactory` that combines the combines the two for one easy parameter to `StateService`.
* Make `GlobalState` on the `State` model generic, and pass in default values through a constructor since we can't new up generics in typescript.
* Make `GlobalState` on the `StateService` generic, and replace any initializers with calls to the `StateFactory`
* Add `GlobalState` as a type argument to the `StateMigrationService` so when we init the object we use the correct one. 

## Before you submit

### DIFF
--- a/angular/src/services/jslib-services.module.ts
+++ b/angular/src/services/jslib-services.module.ts
@@ -76,7 +76,11 @@ import { PasswordRepromptService } from "./passwordReprompt.service";
 import { UnauthGuardService } from "./unauth-guard.service";
 import { ValidationService } from "./validation.service";
 
-import { Account, AccountFactory } from "jslib-common/models/domain/account";
+import { Account } from "jslib-common/models/domain/account";
+import { GlobalState } from "jslib-common/models/domain/globalState";
+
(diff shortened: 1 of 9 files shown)


djsmith85 (approved):
Great work on this!

PR #653, opened 2022-02-03
[bug] Assign client specific account settings during migration

addisonbeck:
This will need to be cherry picked to RC and applied to all clients

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
On https://github.com/bitwarden/jslib/pull/626 I added a line of code to the login process for new accounts that assigned account settings to their defaults on login. This was to make sure client set defaults for settings like vaultTimeout were considered in the stateService, to prevent situations like web vaults with never lock. 

This was erroneous in a couple of ways:
1. It was meant to override default settings with stored settings, but does the opposite. This creates bugs like [this one](https://app.asana.com/0/inbox/1183359552741416/1201760543238398/1201761782438505) where logging out and back in clears any settings that clients have overridden. 
2. It was added as a protection against a migration issue, but having to consider migrations in this way outside of the migration service smells bad.

## Code changes
To resolve this, I removed the problematic line of code and instead account for defaults in the migration service itself. This was done by:

1. Giving the migration service an AccountFactory to create accounts from
2. Setting account settings during migration the same way we set globals now: checking for a value, and using the default if we don't find one.
3. (a refactor) setting account settings in one place and sharing them between unauthenticated and authenticated state migration flows instead of duplicating the initializer in both places.

## Testing requirements
In addition to testing that this resolves the bug in the attached Asana ticket, we are going to need to double check that settings carry over appropriately during the upgrade path. 

## Before you submit

### DIFF
--- a/angular/src/services/jslib-services.module.ts
+++ b/angular/src/services/jslib-services.module.ts
@@ -360,7 +360,7 @@ import { StateFactory } from "jslib-common/factories/stateFactory";
         new StateMigrationService(
           storageService,
           secureStorageService,
-          new GlobalStateFactory(GlobalState)
+          new StateFactory(GlobalState, Account)
         ),
       deps: [StorageServiceAbstraction, "SECURE_STORAGE"],
     },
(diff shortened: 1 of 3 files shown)


addisonbeck:
@djsmith85 and I spoke about this and one additional change was made: making the order type arguments for State consistent. Sometimes we had <GlobalState, Account>, and others <Account, GlobalState> - I added a commit to make them all <GlobalState, Account>.

djsmith85 (approved):

PR #656, opened 2022-02-07
[bug] Make activity a top level storage key

addisonbeck:
https://app.asana.com/0/1201648796371593/1201771028067731

This will need to be cherry-picked to all clients.

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
During the storage data model refactor we have frequently seen issues where the frequency at which we record activity sometimes creates a race condition with other operations that create unexpected behavior. 

This happens because activity tracking runs outside of normal client processes, and runs a lot (on every click, for example), but access the same Account storage key as other operations (like saving vault timeout settings). This creates a race condition where:

1. The vaultTimeout.valueChange methods starts modifying state to save the updated setting
2. At the same time (from the same click) activity tracking starts its process to record activity
3. Both operations, running at the same time, access the storage key used for the Account they are modifying
4. Explosions begin as the operations save over each other with bad data

## Code changes
To stop this from happening I've moved account activity to its own top level storage key, outside of the Account model that represents settings, vaults, etc. This way these processes access different keys and don't step on each others toes.

## Before you submit

### DIFF
--- a/common/src/models/domain/account.ts
+++ b/common/src/models/domain/account.ts
@@ -94,7 +94,6 @@ export class AccountProfile {
   everBeenUnlocked?: boolean;
   forcePasswordReset?: boolean;
   hasPremiumPersonally?: boolean;
-  lastActive?: number;
   lastSync?: string;
   userId?: string;
   usesKeyConnector?: boolean;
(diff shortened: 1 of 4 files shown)


differsthecat (approved):
Good fix 👍

djsmith85 (approved):

PR #660, opened 2022-02-09
[bug] Fix logout timeout action for inactive accounts

addisonbeck:
https://app.asana.com/0/inbox/1183359552741416/1201787020823690/1201782081866250

This will need to be cherry picked to rc

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
When an account has the Vault Timeout Action set to Logout, the other authenticated accounts incorrectly get logged out

## Code changes
* Pass userId in to the logout callback parameter to the vaultTimeoutService. The message handle in desktop already expects this.
* Set lastActive on account login, and null it on account deauthentication. This prevents an issue where newly logged in accounts immediately time out due to inactivity.

## Before you submit

### DIFF
--- a/angular/src/services/jslib-services.module.ts
+++ b/angular/src/services/jslib-services.module.ts
@@ -313,7 +313,8 @@ import { StateFactory } from "jslib-common/factories/stateFactory";
           keyConnectorService,
           stateService,
           null,
-          async () => messagingService.send("logout", { expired: false })
+          async (userId?: string) =>
+            messagingService.send("logout", { expired: false, userId: userId })
         ),
       deps: [
         CipherServiceAbstraction,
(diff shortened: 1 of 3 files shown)


djsmith85 (approved):

addisonbeck:
@MGibson1 I'm not totally sure what you mean, but this userId is optional and is only used in desktop. We only really pass userIds around when dealing with inactive accounts (so just on desktop). Other clients use the current active user Id and run as expected.

MGibson1:
Addison, that's exactly my concern. The only place we use the method `executeTimeoutAction` is https://github.com/bitwarden/jslib/blob/master/common/src/services/vaultTimeout.service.ts#L71-L75. Will we still log out if we don't pass userId?

PR #663, opened 2022-02-10
Individually await saving of orgs and providers

differsthecat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Now that organizations and providers are saved to the same state object, we need to save them individually instead of at the same time because they were causing shenanigans and overwriting each other. 

## Code changes

- **common/src/services/sync.service.ts:** Await saving of organizations and providers individually

## Testing requirements

Ensure both organizations and providers show up as expected on the vault page.

## Before you submit

### DIFF
--- a/common/src/services/sync.service.ts
+++ b/common/src/services/sync.service.ts
@@ -328,10 +328,8 @@ export class SyncService implements SyncServiceAbstraction {
       }
     });
 
-    await Promise.all([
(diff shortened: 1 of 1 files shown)


cscharf (approved):
Awesome, and great use of the term, `shenanigans`

PR #666, opened 2022-02-11
Add StateVersion.Three to fix premium migration

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix bug where user will not have premium after upgrading, until they log out and log back in.

Note: will be cherry-picked to `rc`

## Code changes

- `common/src/services/stateMigration.service.ts` - add a new state migration to read the premium status from the accessToken and then save it to that account
- `common/src/services/token.service.ts` - extract token decoding logic into a static method to avoid circular dependencies (ty @MGibson1)

## Before you submit

### DIFF
--- a/common/src/services/stateMigration.service.ts
+++ b/common/src/services/stateMigration.service.ts
@@ -472,7 +472,6 @@ export class StateMigrationService<
     const globals = await this.getGlobals();
     globals.stateVersion = StateVersion.Three;
     await this.set(keys.global, globals);
-
   }
 
   protected get options(): StorageOptions {
--- a/common/src/services/stateMigration.service.ts
+++ b/common/src/services/stateMigration.service.ts
@@ -155,6 +155,9 @@ export class StateMigrationService<
         case StateVersion.One:
           await this.migrateStateFrom1To2();
           break;
+        case StateVersion.Two:
+          await this.migrateStateFrom2To3();
+          break;
       }
 
       currentStateVersion += 1;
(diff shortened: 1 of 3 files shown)


common/src/services/stateMigration.service.ts
```diff
@@ -448,6 +450,23 @@ export class StateMigrationService<
     }
   }
 
+  protected async migrateStateFrom2To3(): Promise<void> {
+    const authenticatedUserIds = await this.get<string[]>(keys.authenticatedAccounts);
+    await Promise.all(
+      authenticatedUserIds.map(async (userId) => {
+        const account = await this.get<Account>(userId);
```
addisonbeck, line 457: This needs to use the generic TAccount

addisonbeck (commented):
Looks good, one small thing

common/src/services/stateMigration.service.ts
```diff
@@ -448,6 +450,23 @@ export class StateMigrationService<
     }
   }
 
+  protected async migrateStateFrom2To3(): Promise<void> {
```
MGibson1, line 453: I believe you're missing a call to this method from `migrate`

MGibson1 (changes requested):
Looks good, but we need to use the migration 🤣

common/src/services/stateMigration.service.ts
addisonbeck: Good catch! StateVersion.Latest also needs to be bumped up to three

MGibson1:
Tested with v2=>3 migration and it works

MGibson1:
Tested with v1 => 2 => 3 and retained my premium status

common/src/services/stateMigration.service.ts
eliykat: Yikes! I definitely had this in another branch, obviously lost it along the way when squashing/rebasing

eliykat: Should be fixed now, called from `migrate` and updated the `stateVersion` at the end of the migration method itself.

MGibson1 (approved):

addisonbeck (approved):

PR #672, opened 2022-02-11
Clean up the cached account

djsmith85:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
With https://github.com/bitwarden/jslib/pull/668 we introduced an account cache which results in good performance but had some issues when logging out and back in again.

## Code changes
- **common/src/services/state.service.ts:** Make sure the account cache is cleared when logging out

## Testing requirements
Testing will be done on the clients with logging out of accounts and back in again. Also with timeoutAction == logout

## Before you submit

### DIFF
--- a/common/src/services/state.service.ts
+++ b/common/src/services/state.service.ts
@@ -2425,6 +2425,7 @@ export class StateService<
 
   protected removeAccountFromMemory(userId: string = this.state.activeUserId): void {
     delete this.state.accounts[userId];
+    this.accountDiskCache.delete(userId);
   }
 
   protected async pruneInMemoryAccounts() {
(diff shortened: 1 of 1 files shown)


common/src/services/state.service.ts
```diff
@@ -2425,13 +2425,15 @@ export class StateService<
...
   protected removeAccountFromMemory(userId: string = this.state.activeUserId): void {
     delete this.state.accounts[userId];
+    this.accountDiskCache.delete(userId);
   }
 
   protected async pruneInMemoryAccounts() {
     // We preserve settings for logged out accounts, but we don't want to consider them when thinking about active account state
     for (const userId in this.state.accounts) {
       if (!(await this.getIsAuthenticated({ userId: userId }))) {
         delete this.state.accounts[userId];
```
Hinton, line 2435: We could just call `removeAccountFromMemory` here and avoid the duplicated code 😁

Hinton (approved):

PR #673, opened 2022-02-11
[bug] Also check for a never lock timeout when determining where to put tokens

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
https://github.com/bitwarden/jslib/pull/661 resolved a bug causing "On Restart - Logout" timeouts to not work, but introduced a bug where "Never - Logout" timeouts do not work. This is because we never store the access token to disk for logout actions, but we should for never logout.

## Code changes
* Created a helper for timeout based options
* Added a check for `timeout != null` to the timeout based options conditional, and to the conditional that manages this behavior during login scaffolding. 
* Applied the updated timeout based options to any getters/setters that were using it before.

## Testing requirements
I tested never lock, never logout, on restart lock, and on restart logout - all were behaving as expected.

## Before you submit

### DIFF
--- a/common/src/services/state.service.ts
+++ b/common/src/services/state.service.ts
@@ -151,20 +151,12 @@ export class StateService<
   }
 
   async getAccessToken(options?: StorageOptions): Promise<string> {
-    const defaultOptions =
-      (await this.getVaultTimeoutAction({ userId: options?.userId })) === "logOut"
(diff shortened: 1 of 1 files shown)


differsthecat (approved):

MGibson1 (commented):
lgtm

PR #674, opened 2022-02-11
[bug] Allow for toggling the account cache

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
We can't depend on the account cache in web because we split storage between session and local storage. This creates an inconsistent cache that at any given time may be caching session storage and may be caching local storage.

## Code changes
The performance issues the account cache addresses aren't really relevant to web, since we already store vault data in memory there. This PR just creates an option to the StateService to disable the cache.

## Before you submit

### DIFF
--- a/common/src/services/state.service.ts
+++ b/common/src/services/state.service.ts
@@ -76,7 +76,8 @@ export class StateService<
     protected secureStorageService: StorageService,
     protected logService: LogService,
     protected stateMigrationService: StateMigrationService,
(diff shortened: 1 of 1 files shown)


kspearrin:
@djsmith85 Didnt you make some changes in other methods? Does that need to be added here?

addisonbeck:
Good catch, I did find one other reference that could use a check.

Hinton (approved):

djsmith85 (approved):

PR #677, opened 2022-02-11
Clear active user if state has no users

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Logging out in some clients is crashing because the active user is being set back to the user that was just logged out.

I'm not entirely sure when/how this started, but a simple length check on state and setting the active user to null if we don't have any options does the job of logging out successfully. 

## Code changes
* Set active userId to null if state doesn't have any users
* Use getAccount/saveAccount for account teardown functions. I don't think this has any effect on the bug, but we should be using those methods for all account access, especially with the cache addition. 

## Before you submit

### DIFF
--- a/common/src/services/state.service.ts
+++ b/common/src/services/state.service.ts
@@ -2381,10 +2381,7 @@ export class StateService<
     );
     await this.saveAccount(
       this.resetAccount(storedAccount),
-      this.reconcileOptions(
-        { userId: userId },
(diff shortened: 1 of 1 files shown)


addisonbeck:
something else is up here

djsmith85 (approved):

PR #680, opened 2022-02-13
[Icons] Fix sass deprecation warnings

vincentsalucci:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [X] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Other

## Objective
Cleanup the new `bwicons -> styles.scss` file that has deprecation warnings revolving around the use of `/` for division.  Use the `calc()` function since the `math.div` function is not available in our current version of `sass`

## Code changes
- **styles.scss:** Surrounded all division math with the `calc()` function. Removed unused `export` call

## Testing requirements
- Make sure icons using either `bwi-fw, bwi-lg, bwi-ul, bwi-li, or bwi-li & bwi-lg` still appear the same throughout the client applications (once the updated `jslib` has been applied)

## Before you submit

### DIFF
--- a/angular/src/scss/bwicons/styles/style.scss
+++ b/angular/src/scss/bwicons/styles/style.scss
@@ -31,7 +31,7 @@ $icomoon-font-path: "~@bitwarden/jslib-angular/src/scss/bwicons/fonts/" !default
 
 // Fixed Width Icons
 .bwi-fw {
-  width: (18em / 14);
+  width: math.div(18em, 14);
   text-align: center;
 }
 
@@ -41,8 +41,8 @@ $icomoon-font-path: "~@bitwarden/jslib-angular/src/scss/bwicons/fonts/" !default
 }
 
 .bwi-lg {
-  font-size: (4em / 3);
-  line-height: (3em / 4);
+  font-size: math.div(4em, 3);
+  line-height: math.div(3em, 4);
   vertical-align: -15%;
 }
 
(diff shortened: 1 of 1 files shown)


angular/src/scss/bwicons/styles/style.scss
```diff
@@ -84,12 +84,12 @@ $icomoon-font-path: "~@bitwarden/jslib-angular/src/scss/bwicons/fonts/" !default
 
 .bwi-li {
   position: absolute;
-  left: -(30em / 14);
-  width: (30em / 14);
-  top: (2em / 14);
+  left: calc(30em / 14) * -1;
```
cscharf, line 87: Wouldn't the `* -1` go inside of the `calc()` function? Order or operations for deriving a negative result shouldn't matter in these cases anyway.

angular/src/scss/bwicons/styles/style.scss
```diff
@@ -84,12 +84,12 @@ $icomoon-font-path: "~@bitwarden/jslib-angular/src/scss/bwicons/fonts/" !default
...
+  width: calc(30em / 14);
+  top: calc(2em / 14);
   text-align: center;
   &.bwi-lg {
-    left: -(30em / 14) + (4em / 14);
+    left: (calc(30em / 14) * -1) + calc(4em / 14);
```
cscharf, line 92: Same

angular/src/scss/bwicons/styles/style.scss
Hinton: Do we need the `* -1` trick or does calc support that natively using `calc(-30em/14)`?

vincentsalucci: After doing some digging it appears that as long as we supply a unit of measurement the negative value will be calculated properly. I will update the PR to remove  `* -1` trick

Hinton (approved):

PR #687, opened 2022-02-16
fix infinite recursion on getEncKey

kspearrin:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Allow new `bw serve` command to work without getting stuck in an infinite recursion loop.

It's still not clear to be with this only happens when running the new `bw serve` CLI command. Perhaps because all other clients are preceded by an unlock function of some sort.

## Code changes

Moved `getEncKey()` functionality out to a private helper so that the `sequentialize()` decorator does not try to parallelize the call when coming from other internal functions such as `validateKey()`.

## Testing requirements

No behavior to test here other than ensure we do not lose the performance boost of  `getEncKey()` being sequentialized in other areas of the code, such as vault decryption.

## Before you submit

### DIFF
--- a/common/src/services/crypto.service.ts
+++ b/common/src/services/crypto.service.ts
@@ -158,41 +158,8 @@ export class CryptoService implements CryptoServiceAbstraction {
   }
 
   @sequentialize(() => "getEncKey")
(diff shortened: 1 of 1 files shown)


Hinton (approved):
I hate this, but I can't think of an alternative quick fix.

PR #688, opened 2022-02-17
[feature] Implement scope warning for exports

addisonbeck:
https://app.asana.com/0/1183359552741420/1200074829339233

## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
Exports are often confusing to users in organizations because there is no indication on screen that a personal export won't include organization items and vice versa. We should add an information panel to this screen noting what exactly is being exported.

## Code changes
- **organization.service:** Added a `hasOrganizations()` method.
- **export-scope-callout.component:** Created this component to be used in client `export.component` templates.

## Before you submit

### DIFF
--- a/angular/src/components/export-scope-callout.component.ts
+++ b/angular/src/components/export-scope-callout.component.ts
@@ -26,7 +26,6 @@ export class ExportScopeCalloutComponent implements OnInit {
     if (!(await this.organizationService.hasOrganizations())) {
       return;
     }
-    this.show = true;
     this.scopeConfig =
       this.organizationId != null
         ? {
@@ -39,5 +38,6 @@ export class ExportScopeCalloutComponent implements OnInit {
             description: "exportingPersonalVaultDescription",
             scopeIdentifier: await this.stateService.getEmail(),
           };
+    this.show = true;
   }
 }
--- a/angular/src/components/export-scope-callout.component.ts
+++ b/angular/src/components/export-scope-callout.component.ts
@@ -1,0 +1,43 @@
+import { Component, Input, OnInit } from "@angular/core";
+
+import { OrganizationService } from "jslib-common/abstractions/organization.service";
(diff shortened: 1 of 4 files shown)


angular/src/components/export-scope-callout.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor(
+    protected organizationService: OrganizationService,
+    protected stateService: StateService
+  ) {}
+
+  async ngOnInit(): Promise<void> {
+    if (!(await this.organizationService.hasOrganizations())) {
+      return;
+    }
+    this.show = true;
```
eliykat, line 29: Suggestion: move this line to the end of this method after all the async calls, so that text (e.g. org name or user's email) isn't popping in.

eliykat (changes requested):

angular/src/components/export-scope-callout.component.ts
addisonbeck: I moved this as requested, but I don't _think_ the situation described here is possible? ngOnInit is [supposed](https://angular.io/api/core/OnInit#ngoninit) to run completely before any content is loaded.

Regardless, it's moved, so just trying to see if you know something I don't about this behavior.

differsthecat (approved):
Looks great!

angular/src/components/export-scope-callout.component.ts
eliykat: @differsthecat The providers tab was exactly what I'm thinking of. But @addisonbeck I think the simplest answer is that Angular calls the lifecycle hooks at the appropriate times, but the DOM rendering does not wait for them to return/resolve. This is why we have spinners or `*ngIf` directives. Here's a simple-as-possible reproduction of this behaviour: https://stackblitz.com/edit/angular-4ariyn

eliykat (approved):

PR #689, opened 2022-02-19
Feature/password protected export

MGibson1:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
Implements recommendations from bitwarden/cli#446.
* Password protected json format now simply uses the json-formatted export, rather than allowing any contained data.
  * This also simplifies the format definition -- `format` property is removed

## Code changes
There are numerous simplifications and refactors here. The functional changes are described above
- **export.service abstraction**: Make a type for export formats
- **export.service implementation**: Biggest change is to add optional orgId to `getExport` This simplifies calls since we're always testing `orgId ? getExport : getOrganizationExport`. We can easily handle this internally.
- **BitwardenJsonImporter**: Error for password protected jsons. This will be used in clients to immediately retry import with a password prompt.
- **BitwardenPasswordProtectedImporter**: subclass json importer. This is subclassed because of the extra password requirement.
- **import.service**: I don't want to display password protected as an import option anywhere, but it should be available internally to decrypt password protected imports.

### Tests
* **BitwardenJsonImporter**: Tests that we are returning the correct error message for a password protected file
* **BitwardenPasswordProtectedImporter**: Removes references to format and updates tests for subclassing the json importer
* **export.service spec**: Remove reference to format in password protected file

## Testing requirements

Will be tested in clients.

## Before you submit


common/src/services/import.service.ts
```diff
@@ -209,7 +208,7 @@ export class ImportService implements ImportServiceAbstraction {
   }
 
   getImporter(
-    format: ImportType,
+    format: ImportType | "bitwardenpasswordprotected",
```
eliykat, line 211: do we still need this format? My understanding is that we can just import with `bitwardenjson` and it will figure it out.

Also applies to line 223.

I think I also saw it in the cli PR, maybe just do a global search (assuming that my understanding is correct).

spec/common/importers/bitwardenJsonImporter.spec.ts
```diff
@@ -0,0 +1,31 @@
+import Substitute, { SubstituteOf } from "@fluffy-spoon/substitute";
```
eliykat, line 1: Curly brackets seem misplaced

eliykat (changes requested):

common/src/services/export.service.ts
```diff
@@ -39,22 +42,22 @@ export class ExportService implements ExportServiceAbstraction {
     private cryptoFunctionService: CryptoFunctionService
   ) {}
 
-  async getExport(format: "csv" | "json" | "encrypted_json" = "csv"): Promise<string> {
+  async getExport(format: ExportFormat = "csv", organizationId?: string): Promise<string> {
+    if (organizationId) {
```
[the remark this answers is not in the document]
MGibson1: Why? are we not using falsiness anymore?

common/src/services/import.service.ts
MGibson1: There are a couple of points here that as I write them sounds like excuses, but here they are:

The `ImportType` type is used to display the known import types with clients. Since `bitwardenpasswordprotected` is meant to be a transparent, internal alternative to `bitwardenjson`, we don't want to include it in that display. Since there's only one, I felt the best course right now is to just tack it on here (the only place it's needed) rather than try and figure out a good permanent home for these internal-only importers. I really can't see the use for another one atm.

We can get rid of the importer class entirely if we want to roll all the password protected. However, I think that the way to do that properly is to add a `parse_with_password` method for specifically this importer. It just feels weird since the way we currently alter importer behavior is with the importer type. It also feels weird to expand the `bitwardenjson` importer to include another json variant.

common/src/services/export.service.ts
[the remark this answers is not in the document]
eliykat: I thought it was our code style, and is definitely my preference because it's clearer what you're testing (false? null/undefined? empty string? etc). But if I'm mistaken here that's fine.

common/src/services/import.service.ts
eliykat: Good explanation thank you, I agree that it's a little awkward being tacked on but it seems OK given that it's the only internal-only importer.

common/src/abstractions/import.service.ts
```diff
@@ -9,6 +10,14 @@ export abstract class ImportService {
...
   getImportOptions: () => ImportOption[];
-  import: (importer: Importer, fileContents: string, organizationId?: string) => Promise<Error>;
-  getImporter: (format: ImportType, organizationId: string, password?: string) => Importer;
+  import: (
+    importer: Importer,
+    fileContents: string,
+    organizationId?: string
+  ) => Promise<ImportError>;
+  getImporter: (
+    format: ImportType | "bitwardenpasswordprotected",
```
eliykat, line 19: Just noticed that the abstract class is importing the `ImportType` from the implementation, which seems backwards. i.e. the implementation is defining the abstraction. Do you think it's worth lifting the `featuredImportOptions` and `regularImportOptions` to their own file, or (probably a better idea) lifting their `id`s out to a real `importType` enum? It's not quite as elegant as your current `ImportType` solution but feels like it's separating things out a bit more.

I've seen a few cases lately where we've stuck enums or types at the top of their related services instead of in their own file and I'm not sure about it. I can see why it was done here but once it starts growing it can get messy.

MGibson1: The reason it's here and not in the abstract file or anywhere else is because the only place we use the split version of the ImportType (`featuredImportOptions` and `regularImportOptions`) is directly in this service.

If we move them out we'll have to export them rather than have them be private. I didn't like them being exported from the abstraction, it didn't make sense to me.

I think I'm OK with a new file. I don't see a reason to extract to an enum since I'm able to generate the `ImportType` type from our import option arrays.

common/src/importers/bitwardenJsonImporter.ts
```diff
@@ -15,14 +15,20 @@ export class BitwardenJsonImporter extends BaseImporter implements Importer {
...
   }
 
   async parse(data: string): Promise<ImportResult> {
     this.result = new ImportResult();
     this.results = JSON.parse(data);
-    if (this.results == null || this.results.items == null || this.results.items.length === 0) {
+    if (this.results == null || this.results.items == null) {
+      if (this.results?.passwordProtected) {
+        this.result.success = false;
+        this.result.errorMessage = this.i18nService.t("importPasswordRequired");
```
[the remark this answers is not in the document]
eliykat: nice solution to this 👍

eliykat (approved):

PR #690, opened 2022-02-20
[AuthService refactor] Don't clear state if 2FA is invalid

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix error where you've used an incorrect 2FA code and then try again: `this.logInStrategy is null`.

## Code changes

When 2FA is required, `authService` saves state so that it can resubmit all login information with the 2FA token.

However, we were clearing state as soon as the 2FA request was submitted, without seeing whether it was successful or not. This meant that subsequent attempts to provide the 2FA token failed.

Now, only clear state if the 2FA request was successful.

Note: clients will need to be updated with jslib and an i18n string.

## Testing requirements

* Confirm bug is fixed
* After 2 minutes on the 2FA screen, any attempt to provide 2FA credentials should fail with a session timeout error. You should be able to go back and start the login process again.

## Before you submit


vincentsalucci (approved):
This seems fine to me! Wouldn't mind a second review though.

common/src/services/auth.service.ts
```diff
@@ -115,11 +115,13 @@ export class AuthService implements AuthServiceAbstraction {
...
-    try {
-      return await this.logInStrategy.logInTwoFactor(twoFactor);
-    } finally {
+    const result = await this.logInStrategy.logInTwoFactor(twoFactor);
+
+    // Only clear state if 2FA token has been accepted, otherwise we need to be able to try again
+    if (!result.requiresTwoFactor) {
       this.clearState();
     }
+    return result;
```
MGibson1, line 124: We need to clear this state more aggressively than this.

Certainly we want a try/catch around this for unexpected errors, but I think we also want to somehow detect moving off of the 2fa page at all.

In some sense, I think a single-shot is the lesser of two evils.

MGibson1 (changes requested):

common/src/services/auth.service.ts
eliykat: As discussed, have added:
* error handling in `logInTwoFactor` (not required in `logIn` because state has not yet been saved).
* a timeout that will clear state after 2 minutes, plus an error message that gets returned to the client

MGibson1 (approved):
Man, I'm not sure if you're feeling it right now, but the ease you were able to add this new feature speaks volumes to your refactor -- nice work 😄

PR #697, opened 2022-03-01
Fix autoFillOnPageLoadDefault default value

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix https://github.com/bitwarden/browser/issues/2400.

`autoFillOnPageLoadDefault` should default to true if it's not explicitly set. Note that this is the default per-item autofill setting - it still won't do anything unless `enableAutoFillOnPageLoad` is true (to enable the feature in general).

The intention was always to replicate the previous behaviour, where once Autofill On Page Load was enabled, all items would autofill by default unless configured otherwise.

Luckily this does not appear to be a migration issue as such, any saved value should still have been migrated, we just need to handle the `null` case.

## Code changes

Change default value if null.

## Testing requirements

* use browser client on a fresh install (no saved data)
* log in
* confirm that Settings > Options are:
  * Enable Auto-Fill On Page Load: `false`
  * Default autofill setting for login items: `Auto-fill on page load`
* brief regression test to make sure the actual autofill behaviour follows these settings

## Before you submit

### DIFF
--- a/common/src/models/domain/account.ts
+++ b/common/src/models/domain/account.ts
@@ -100,7 +100,7 @@ export class AccountProfile {
 
 export class AccountSettings {
   autoConfirmFingerPrints?: boolean;
-  autoFillOnPageLoadDefault?: boolean;
+  autoFillOnPageLoadDefault?: boolean = true;
   biometricLocked?: boolean;
   biometricUnlock?: boolean;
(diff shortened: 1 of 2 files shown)


addisonbeck (approved):
Thanks @eliykat

PR #699, opened 2022-03-01
Roll back organizationId and collectionId's if ciphers fail to be moved

differsthecat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
While working on [this](https://github.com/bitwarden/server/pull/1890), I noticed that if a bulk share action fails, the items in the cipher view are updated to show that they transferred to the organization even though they were not. 

The items in the vault that were attempted to be shared are then in this state until a sync is performed or the vault is refreshed.

## Code changes

- **common/src/services/cipher.service.ts:** Add try catch that reverts setting of items `organizationId` and `collectionIds` properties if the api call fails.

## Testing requirements

Make sure that if an item fails to share to an organization (e.g. has attachments and is sent to a free org)

## Before you submit

### DIFF
--- a/common/src/services/cipher.service.ts
+++ b/common/src/services/cipher.service.ts
@@ -657,7 +657,15 @@ export class CipherService implements CipherServiceAbstraction {
     }
     await Promise.all(promises);
     const request = new CipherBulkShareRequest(encCiphers, collectionIds);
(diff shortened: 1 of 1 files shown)


eliykat (approved):
Good spot!

PR #702, opened 2022-03-02
add additional familySponsorship properties to profile org data

jlf0dev:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Initial properties for models in clients. Subject to change in the future but will allow us to work on clients
## Code changes

- **file.ext:** Description of what was changed and why

## Before you submit

### DIFF
--- a/common/src/enums/sponsorshipStatus.ts
+++ b/common/src/enums/sponsorshipStatus.ts
@@ -1,0 +1,8 @@
+export enum SponsorshipStatus {
+  None = 0,
+  Pending = 1,
+  Sent = 2,
+  Active = 3,
+  PendingRevoke = 4,
+  Revoked = 5,
+}
--- a/common/src/enums/sponsorshipStatus.ts
+++ b/common/src/enums/sponsorshipStatus.ts
@@ -1,8 +1,6 @@
 export enum SponsorshipStatus {
-  None = 0,
-  Pending = 1,
-  Sent = 2,
-  Active = 3,
-  PendingRevoke = 4,
-  Revoked = 5,
+  Requested = 0, // self-hosted only
+  Sent = 1,
+  Active = 2,
+  ToRevoke = 3,
 }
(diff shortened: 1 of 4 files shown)


common/src/enums/sponsorshipStatus.ts
```diff
@@ -0,0 +1,8 @@
+export enum SponsorshipStatus {
+  None = 0,
+  Pending = 1,
+  Sent = 2,
+  Active = 3,
+  PendingRevoke = 4,
+  Revoked = 5,
```
MGibson1, line 7: I think these can be combined into `toDelete`, since we don't really delineate between revoked and removed for sponsorship lifetime purposes

common/src/enums/sponsorshipStatus.ts
```diff
@@ -0,0 +1,8 @@
+export enum SponsorshipStatus {
+  None = 0,
+  Pending = 1,
+  Sent = 2,
```
MGibson1, line 4: I would prefer these to be just something like `offered`

I'm not sure when `none` would be used since the row just wouldn't exist, `pending` would be a transient state after create and before sending the offer, and `Sent` would require updating the model after emailing, but I'd prefer not to have that be part of a transaction with the DB. It's why we have resend commands

I think you might be trying to expand these states a little more than I imagined to accommodate self-hosted states where things have "happened" but haven't been synced. I think it's a good enough experience to just have the state lag behind by 1 sync, right?

common/src/enums/sponsorshipStatus.ts
MGibson1: ofc, we could make it `toRemove` and `toRevoke` to record who caused the deletion...

MGibson1 (changes requested):

common/src/enums/sponsorshipStatus.ts
```diff
@@ -0,0 +1,6 @@
+export enum SponsorshipStatus {
+  Requested = 0, // self-hosted only
+  Sent = 1,
+  Active = 2,
+  ToRevoke = 3,
```
jlf0dev, line 5: I left in `Requested` here because we don't include a Sync Date to the client that we could use to communicate that it is on self hosted and hasn't synced yet like the figma shows.

PR #710, opened 2022-03-07
Add premium and unless-premium directives

Hinton:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Adds two directives which will allow us to do the following.

```html
<span *appPremium>Premium</span>
<span *appNotPremium>No Premium<span>
```

## Before you submit

### DIFF
--- a/angular/src/directives/unless-premium.directive.ts
+++ b/angular/src/directives/unless-premium.directive.ts
@@ -1,0 +1,27 @@
+import { Directive, OnInit, TemplateRef, ViewContainerRef } from "@angular/core";
+
+import { StateService } from "jslib-common/abstractions/state.service";
+
+/**
+ * Hides the element if the user has premium.
+ */
+@Directive({
+  selector: "[appUnlessPremium]",
+})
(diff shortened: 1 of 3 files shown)


angular/src/directives/unless-premium.directive.ts
```diff
@@ -0,0 +1,27 @@
+import { Directive, OnInit, TemplateRef, ViewContainerRef } from "@angular/core";
+
+import { StateService } from "jslib-common/abstractions/state.service";
+
+/**
+ * Hides the element if the user has premium.
+ */
+@Directive({
+  selector: "[appUnlessPremium]",
```
cscharf, line 9: Could we use more clear language such as `appNotPremium`, vs. "unless"? I had to wrack my brain a few times trying to do the reverse Boolean logic in my head with that word to figure out whether the logic was correct or not or to validate it... that may be because I've only had 1 cup of coffee so far, but I think using *not* is more clear.

cscharf (approved):
Looks good, thank you Oscar!

PR #716, opened 2022-03-10
Avoid duplicate fullSync api calls

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix https://github.com/bitwarden/web/issues/1324.

I came across this code in the web navbar component which is likely causing duplicate fullSyncs: https://github.com/bitwarden/web/blob/master/src/app/layouts/navbar.component.ts#L38

That code is not unreasonable though, sometimes we need to ensure a full sync has been done. Better to handle duplicate calls in syncService so that callers don't need to worry about it.

## Code changes

Add sequentialize decorator to `syncService.fullSync`.

## Before you submit


Hinton (approved):
We need to be careful with sequentialize to ensure we don't get into potential lock scenarios.

PR #719, opened 2022-03-11
Add Chromatic

Hinton:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Adds Chromatic for Component Library PR reviews. Also added a11y addon to stprybook, made storybook runable without web and added colors story.

## Before you submit

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -38,3 +38,6 @@ jobs:
           workingDir: ./components
           exitOnceUploaded: true
           onlyChanged: true
+          externals:
+            - "*.scss"
+            - "tailwind.config*.js"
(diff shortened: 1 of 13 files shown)


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,39 @@
+---
+name: Chromatic
+
+on: push
+
```
joseph-flinn, line 5: Does this need to run on every change to jslib? If it just needs to run on changes to `./components`, can we add a `paths` block?

Hinton: We should be safe to only run it when components change.

joseph-flinn:
@Hinton Looks good from the DevOps point of view

vincentsalucci (approved):
I don't really speak github workflow - so I'll leave that to @joseph-flinn

PR #723, opened 2022-03-14
Remove Internet Explorer logic

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Remove Internet Explorer logic since we no longer supports that browser. While at it, I also extracted default KDF settings to constants.

EC-109

## Testing requirements

Regression test actions related to KDF.

- Register
- Login
- Change KDF
- Create, modify, access send with password

## Before you submit

### DIFF
--- a/common/src/services/send.service.ts
+++ b/common/src/services/send.service.ts
@@ -5,6 +5,7 @@ import { FileUploadService } from "../abstractions/fileUpload.service";
 import { I18nService } from "../abstractions/i18n.service";
 import { SendService as SendServiceAbstraction } from "../abstractions/send.service";
 import { StateService } from "../abstractions/state.service";
+import { SEND_KDF_ITERATIONS } from "../enums/kdfType";
 import { SendType } from "../enums/sendType";
 import { Utils } from "../misc/utils";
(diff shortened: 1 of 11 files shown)


common/src/services/send.service.ts
```diff
@@ -19,6 +19,8 @@ import { ErrorResponse } from "../models/response/errorResponse";
 import { SendResponse } from "../models/response/sendResponse";
 import { SendView } from "../models/view/sendView";
 
+export const SEND_KDF_ITERATIONS = 100000;
```
eliykat, line 22: This isn't a great place to declare this. I think it's better to keep these default settings together, e.g. in the `KdfType` enum file with the others, or move all of them to a dedicated file.

eliykat (changes requested):

eliykat:
Can you add a reference to the Jira ticket?

common/src/services/send.service.ts
Hinton: Fair enough, I was a bit split since I didn't like adding it to the KDF enum either as it's more of a feature isolated constant.

eliykat (approved):

PR #727, opened 2022-03-16
[End User Vault Refresh] Refactor route permission checking

eliykat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

This enables the latest changes in https://github.com/bitwarden/web/pull/1524.

In https://github.com/bitwarden/jslib/pull/713 I moved access permissions into the jslib organization model. This is because we need to access it from various places in web and I wanted it to live in a single place. But really we should not hard-bake client permissions into a jslib model, and I hit this issue pretty quickly when I needed to change these permissions.

This replaces the previous changes with a more flexible method that will take an array of `Permission`s and evaluate it for the organization. This is moved from web's `organization-type-guard.service.ts` so it can be used more generally.

## Code changes

As above

## Before you submit

### DIFF
--- a/common/src/models/domain/organization.ts
+++ b/common/src/models/domain/organization.ts
@@ -184,7 +184,8 @@ export class Organization {
   }
 
   hasAnyPermission(permissions: Permissions[]) {
-    return (permissions.includes(Permissions.AccessEventLogs) && this.canAccessEventLogs) ||
+    const specifiedPermissions =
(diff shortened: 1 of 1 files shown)


Hinton (commented):
Prettier seems to be sad.

eliykat:
> Forgot to run Prettier?

My least favourite message :(

Hinton (approved):

PR #728, opened 2022-03-16
Add ellipsis pipe

differsthecat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

I noticed that we don't have a nice way to truncate large text (such as organization names) while working on the end user vault refresh work for browser, so I added this ellipsis pipe to do just that 🙂  

The default limit is 25 characters, but you can pass the amount you want to add the ellipsis on. You can also pass a flag to break on complete words and to also change what the ellipsis is.

## Code changes

- **angular/src/pipes/ellipsis.pipe.ts:** Add ellipsis pipe

## Testing requirements

None, until it is used in one of our clients.

## Before you submit

### DIFF
--- a/angular/src/pipes/ellipsis.pipe.ts
+++ b/angular/src/pipes/ellipsis.pipe.ts
@@ -1,0 +1,13 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'ellipsis'
+})
+export class EllipsisPipe implements PipeTransform {
+  transform(value: string, limit = 25, completeWords = false, ellipsis = '...') {
+    if (completeWords) {
+      limit = value.substring(0, limit).lastIndexOf(' ');
+    }
+    return value.length > limit ? value.substring(0, limit) + ellipsis : value;
+  }
+}
--- a/angular/src/pipes/ellipsis.pipe.ts
+++ b/angular/src/pipes/ellipsis.pipe.ts
@@ -5,9 +5,10 @@ import { Pipe, PipeTransform } from "@angular/core";
 })
 export class EllipsisPipe implements PipeTransform {
   transform(value: string, limit = 25, completeWords = false, ellipsis = "...") {
+    limit -= ellipsis.length;
     if (completeWords) {
       limit = value.substring(0, limit).lastIndexOf(" ");
     }
-    return value.length > limit ? value.substring(0, limit - ellipsis.length) + ellipsis : value;
+    return value.length > limit ? value.substring(0, limit) + ellipsis : value;
   }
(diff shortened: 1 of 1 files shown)


angular/src/pipes/ellipsis.pipe.ts
```diff
@@ -0,0 +1,13 @@
+import { Pipe, PipeTransform } from "@angular/core";
+
+@Pipe({
+  name: "ellipsis",
+})
+export class EllipsisPipe implements PipeTransform {
+  transform(value: string, limit = 25, completeWords = false, ellipsis = "...") {
```
MGibson1, line 7: What do you think about limit being the character limit vs the elipses-start-point? I think it's more intuitive to say "I have room for 25 characters" than to say "I want to start elipses at 22". Of course, this all depends on font. Any way we can calculate that instead?

MGibson1 (commented):
This is freaking awesome. We super need this.

angular/src/pipes/ellipsis.pipe.ts
differsthecat: Ah yes, that totally makes sense! I'll change it to that way instead!

differsthecat: I did a bunch of experimenting with this to see if it could be a bit smarter and return the correct string length based on the width of the containing element, but the solution I came up with felt dubious, so I'm sticking with doing it this more straightforward way 🙂 

I did update the original calculations to take into account the length of the ellipsis so the total length of the string returned accounts for it.

MGibson1 (approved):

angular/src/pipes/ellipsis.pipe.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (value.length <= limit) {
+      return value;
+    }
+    limit -= ellipsis.length;
+    if (completeWords && value.length > limit) {
+      if (value.indexOf(" ") > 0) {
```
MGibson1, line 13: Please combine if statements to reduce nesting

MGibson1 (changes requested):
I'm really sorry for the back and forth on this, I know it feels like a small item that should just go in, but I'm hoping this will be used _a lot_ so I want it to be really robust.

differsthecat:
> I'm really sorry for the back and forth on this, I know it feels like a small item that should just go in, but I'm hoping this will be used _a lot_ so I want it to be really robust.

@MGibson1 No worries it all, I'd rather it be done right! 
It should be good to go now!

MGibson1 (approved):

PR #735, opened 2022-03-24
PS-91 make isMacAppStore return true/false

jlf0dev:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

`isMacAppStore()` was returning undefined if on mac and not a MAS build. This fix forces the function to return true/false

## Code changes

- **file.ext:** Description of what was changed and why

## Testing requirements

`isMacAppStore` is called during the update function (to make sure the build is not a MAS build), while building the toolbar menus (some options like "Check for Update" is not available on MAS build) and while closing the window (to keep application alive in background until user quits it with Cmd-Q). These will need to be checked.

## Before you submit

### DIFF
--- a/electron/src/utils.ts
+++ b/electron/src/utils.ts
@@ -34,7 +34,7 @@ export function isMac() {
 }
 
 export function isMacAppStore() {
-  return isMac() && (process.mas ?? false);
+  return isMac() && process.mas === true;
 }
 
 export function isWindowsStore() {
--- a/electron/src/utils.ts
+++ b/electron/src/utils.ts
@@ -34,7 +34,7 @@ export function isMac() {
 }
 
 export function isMacAppStore() {
-  return isMac() && process.mas && process.mas === true;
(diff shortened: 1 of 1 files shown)


electron/src/utils.ts
```diff
@@ -34,7 +34,7 @@ export function isMac() {
 }
 
 export function isMacAppStore() {
-  return isMac() && process.mas && process.mas === true;
+  return isMac() && (process.mas ?? false);
```
cscharf, line 37: Not sure I understand this change, `process.mas` from electron's docs state's it's going to be `true` or `undefined`, therefore the `=== true` seems like the more correct test vs. using _truthy_.

Perhaps instead use:
```suggestion
  return isMac() && process.mas === true;
```

cscharf: Here's a better example of why it's returning `undefined`

image

cscharf (approved):

PR #737, opened 2022-03-24
[end user vault refresh] Base Changes For Vault Filters

addisonbeck:
https://bitwarden.atlassian.net/browse/SG-95?atlOrigin=eyJpIjoiYzdlY2U4Y2ZjYWVjNGE3ODllMWFlMzljM2Y0M2JmNTAiLCJwIjoiaiJ9

## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
- Support organization filters & collapsable top level filters
- Refactor the groupings component into a module with several smaller and more independent components

An implementation example can be found at https://github.com/bitwarden/desktop/pull/1397

## Code changes
See commits and descriptions

## Before you submit


Hinton:
Without reviewing this in depth, any specific reason to not use https://material.angular.io/cdk/tree/examples? The tree logic feels like it would be quite nice, and would let us get rid of our own `TreeNode` datatypes.

addisonbeck:
No, I just didn't go any further than I had to with refactoring past the component breakup. I appreciate the excuse to though 😁 I'll take a look at the tree component in the cdk.

addisonbeck:
After taking a look at the tree component in the CDK, I'm not sure it offers any real benefit to us compared to the way the tree works now. Any example past a basic tree implementation, a threshold we definitely pass,  still define their own types for nodes, so we wouldn't lose that. 

There also seems to be a lot of library junk baked in to the cdk tree component that we aren't having to deal with right now. Some examples: needing to connect an arbitrary `DataSource` object to the tree element, vs. the typed objects we are using now, and using simple angular syntax for displaying elements (`*ngIf` vs. `[cdkTreeNodeToggleRecursive]` 🤢)

The CDK also doesn't immediately play well with our cross-session collapse logic. We could get this to work, of course, but I'm struggling to see the benefit.

@Hinton let me know if you would like to discuss this some more, but I'd like to not implement the cdk tree component with this work and end any major refactoring where we are now.

angular/src/modules/vault-filter/components/collection-filter.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  isCollapsed(node: ITreeNodeObject) {
+    return this.collapsedFilterNodes.has(node.id);
+  }
+
+  applyFilter(collection: CollectionView) {
+    this.activeFilter.resetFilter();
+    this.activeFilter.selectedCollectionId = collection.id;
+    this.onFilterChange.emit(this.activeFilter);
```
Hinton, line 45: I don't like that we're calling methods on inputs. What are your thoughts on just having this component emit an event that we selected something and let the parent component handle figuring out how the filter should be built?

Hinton: Which is in line with how the other things behave.

addisonbeck: I think that's a good idea. I'll rework this so the `applyFilter` events pass up a `VaultFilter()` instance that they claim should be applied, and let the `VaultFilterComponent` or `VaultComponent` determine if that is valid or not.

addisonbeck: Unfortunately, after taking a closer look at this I think we need to keep the pattern used to manage `activeFilter` the same.

I think it's safe to update  `Input` instances, because the whole thing acts as a closed loop with the initial `activeFilter` value at the top of the component hierarchy being the source of truth for all `Inputs` (even `VaultFiltersComponent` has `activeFilter` as an `Input`, allowing a default value to be set on init). `activeFilter` is passed as an input in each component in the hierarchy, and is emitted at each stage until it bubbles all the way back up to the top, so it is always in sync. 

Leaving the pattern as-is the individual filter components (`Folders`, `Types`, etc.) update their instance of the filter that they use to manage display in their views, and then emits the value back up to all parents to update their filter instances that are used for more of the actual logic processing and filtering. Change detection then picks up, and ensures the instances are all synced up (even though they should be in sync already because of the bottom-up update approach).

Flipping this around and setting up a "we can't modify inputs" rule makes all this logic no matter its context (display vs. filter processing vs. reloading data, etc). the responsibility of the top level component. In this case, that's `VaultComponent`, which means we would be pushing even more logic into a component that is already rather overworked. It doesn't make sense to read and doesn't have any clear pattern.

I'd be happy to talk about this a bit more if you want to go over it some more, @Hinton

Hinton: My main gripe is that we've moved from having the filter logic in a single component to having the same business logic being split across 5. Which makes it very difficult to get an understanding of how everything works.

Why does the folder filter need to know that it should call `resetFilter`? That doesn't have anything to do with selecting a folder.

The `VaultFilterComponent` is a  "smart component" which knows how to fetch the data it requires. It would therefore be logical to have the sub components be strictly "presentational components".

With that said what is your thoughts on performing the following refactor?

---

Combine the status and type components into a single component. That expects a list of items, of the following interface:
```ts
interface Item {
  name: string;
  icon: string;
  id: number; // needed?
}
```

It would have the following bindings. (Following the angular naming convention for [two-way bindings](https://angular.io/guide/two-way-binding))

```ts
@Input() items: Item[];
@Input() selectedItem: Item;

@Output() selectedItemChange = new EventEmitter<Item>();
```

We could do a similar change for the folders and possibly collections?

The logic for how to handle the actions would then end up in the `VaultFilterComponent`, which I think is fine. It will be the centralized place for any filter change to occur. (possibly with inputs from it's parent component).

differsthecat (approved):
I think this looks good, I have been working off of this with web and it's been great!
Thanks @addisonbeck!

PR #745, opened 2022-03-28
[Angular DI] EC-131 Use Injectable services

eliykat:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Refactor our use of Angular DI to automatically manage service dependencies.

## Code changes

* add `@angular/core` to `jslib-common/package.json`. CLI clients will use a webpack plugin to swap this out for `angular-core.noop.ts` at compile time.
* add the `@Injectable()` decorator to all service classes. This generates the necessary metadata for Angular to autodetect their constructor parameters.
* use `InjectionToken`s to provide other dependencies/arguments without having to use factories
* refactor `services.module.ts` to remove factories and `deps` arrays as much as possible. Now we can just specify `useClass` and Angular will figure out the rest.

Client PRs - to show the impact on Angular and non-Angular clients:
* web: https://github.com/bitwarden/web/pull/1568
* cli: https://github.com/bitwarden/cli/pull/516

Other client PRs will follow once I've gotten feedback on this general approach.

## Testing requirements

General regression testing across all clients (except mobile).

## Before you submit


Hinton:
I don't really like adding `@angular/core` as a dependency to common and then swapping it out. Would it be possible to do the reverse? I.E. add a dummy injectable which we swap out in Angular apps to use Angular?

eliykat:
I actually have that already sitting on a branch if you want, but it still requires angular/core to be installed by default. Otherwise webpack does the substitution and can't find the module. It otherwise works as you describe, noop by default.

Hinton:
I was hoping we could remove `@angular/core` from `commons` as keeping it would cause the cli to have a build dependency on Angular.

eliykat:
Not sure if there's any great way around this. Angular clients need `@angular/core` installed within `jslib-common`, CLI doesn't, so I guess we need a way to conditionally include/exclude that package at the `jslib-common` level. The best I can think of is to use the `preinstall` hook in the client to pass an environment variable indicating this, e.g. `BW_DI_ANGULAR`. `jslib-common` then has a `preinstall` hook that checks for this and runs an additional command to install `@angular/core` if it's present.

I also thought about having empty subclasses with the decorator in `jslib-angular`. I think that would work, it would just be an extra layer of indirection to have to remember/maintain.

What do you think?

eliykat:
Thinking through other options: we could have empty subclasses in `jslib-angular` for each service, which just attach the decorator. Angular clients use them as the base class.

The downside is that any services that use `InjectionToken`s would need their constructors repeated in that subclass, which leads us back to the problem of updating dependencies in multiple places. Plus just all the boilerplate this would be.

Hinton:
Could we also change the imports to the noop and to the replace in the webpacks for angular so we don't have imports in common.

eliykat:
I wasn't able to use `noop` by default, I kept getting Angular compiler errors because it thought it was missing the `@Injectable()` decorator. After spending some time experimenting with this, I'm not sure it's the way to go. I'll create a new PR with some more modest improvements.

PR #749, opened 2022-03-30
EC-134 Fix api token refresh

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix https://github.com/bitwarden/cli/issues/490.

### Steps to reproduce
1. log in with the CLI using the `--apikey` option
2. wait for your accessToken to time out, about 5 minutes
3. perform an operation with the server, such as adding or deleting an item
4. run `bw list items` and observe that no items are returned

### Explanation

`nodeApiService` calls the `authService` apikey login method (via a callback) in order to refresh the `accessToken`. This seems like a bit of a hack because logging in does a lot more than just get a new `accessToken`, but it did work before account switching.

After account switching, login methods bootstrap the local environment and local data as well. This means that whenever the access token is refreshed, it will:
* add a duplicate userId to the `authenticatedAccounts` array
* lose existing account data in `data.json` (most notably the encrypted ciphers)

The login logic is no longer suitable for refreshing the `accessToken`, so I've created a separate method to handle this.

## Code changes

- **common/src/services/api.service.ts** - fill out the `doApiTokenRefresh()` method with actual logic to do this. Now that it doesn't rely on `authService`, we can get rid of the callback. Even though this previously would've set all the initial login information (email, kdf, key, premium, etc), I don't think that's necessary. The non-apikey logic only sets the token, so that's all we're doing here.
- **common/src/services/state.service.ts** - tweak the deauthentication logic to remove duplicate userIds from the `authenticatedAccounts` array. While this is transitional logic, it's also a valid way to do this operation in general, so this can just be a permanent change.
- refactor: replace the `TokenRequestTwoFactor` interface with a class, mostly so that we can easy set default values. (this is what most of the diff is)

## Testing requirements

* confirm the defect is fixed
* quick regression test on our various login flows in all typescript clients, especially any involving 2FA

If you don't want to wait 5 minutes between tests for your token to timeout, you can just change the `tokenService.tokenNeedsRefresh` method to return `true`. PM me if you need help.

## Before you submit


differsthecat (approved):
Looks good, this may have some side effects on the directory connector next time we update jslib for it.

PR #759, opened 2022-04-08
Add banner component

Hinton:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Adds the banner component which we'll want for future work. Sets the alert tag by default.

## Before you submit

### DIFF
--- a/components/src/banner/banner.component.html
+++ b/components/src/banner/banner.component.html
@@ -1,0 +1,14 @@
+<div class="tw-py-2.5 tw-px-4 tw-text-contrast tw-flex tw-gap-2" [ngClass]="bannerClass">
+  <div class="tw-block">
+    <i class="bwi {{ icon }} tw-align-middle" *ngIf="icon" aria-hidden="true"></i>
+  </div>
+  <h3 class="tw-inline tw-m-0 tw-text-base tw-shrink-0" *ngIf="title">
+    {{ title }}
+  </h3>
+  <span class="tw-font-semibold tw-text-base tw-grow">
+    <ng-content></ng-content>
+  </span>
+  <button class="tw-border-0 tw-bg-transparent tw-text-contrast tw-p-0" (click)="onClose.emit()">
+    <i class="bwi bwi-close tw-text-sm" *ngIf="icon" aria-hidden="true"></i>
+  </button>
+</div>
--- a/components/src/banner/banner.component.spec.ts
+++ b/components/src/banner/banner.component.spec.ts
@@ -1,0 +1,24 @@
+import { ComponentFixture, TestBed } from "@angular/core/testing";
+
+import { BannerComponent } from "./banner.component";
+
+describe("BannerComponent", () => {
+  let component: BannerComponent;
+  let fixture: ComponentFixture<BannerComponent>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      declarations: [BannerComponent],
+    }).compileComponents();
+  });
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(BannerComponent);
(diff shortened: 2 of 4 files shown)


components/src/banner/banner.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="tw-py-2.5 tw-px-4 tw-text-contrast tw-flex tw-gap-2" [ngClass]="bannerClass">
+  <div class="tw-block">
+    <i class="bwi {{ icon }} tw-align-middle" *ngIf="icon" aria-hidden="true"></i>
```
eliykat, line 3: I don't think we should use interpolation in `class` like this, this is what `ngClass` is for.

components/src/banner/banner.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="tw-py-2.5 tw-px-4 tw-text-contrast tw-flex tw-gap-2" [ngClass]="bannerClass">
```
eliykat, line 1: I didn't know you could use decimals in Tailwind like this 😮  seems like too much choice to me! (your code is fine though)

components/src/banner/banner.component.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  let component: BannerComponent;
+  let fixture: ComponentFixture<BannerComponent>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      declarations: [BannerComponent],
+    }).compileComponents();
+  });
+
+  beforeEach(() => {
```
eliykat, line 15: Why have 2 `beforeEach`s?

components/src/banner/banner.component.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    fixture = TestBed.createComponent(BannerComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it("should create", () => {
+    expect(component).toBeTruthy();
```
eliykat, line 22: Open question, is this test worthwhile? What would let the build compile but break instantiation of the component in TestBed?

components/src/banner/banner.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  warning: "warning",
+  danger: "error",
+};
+
+@Component({
+  selector: "bit-banner",
+  templateUrl: "./banner.component.html",
+})
+export class BannerComponent implements OnInit {
+  @Input() type: BannerTypes = "info";
```
eliykat, line 24: I think we should avoid having inputs that duplicate/shadow HTML attributes (same with title).

components/src/banner/banner.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  @Input() icon: string;
+  @Input() title: string;
+  @Input() useAlertRole = true;
```
eliykat, line 27: This is not referred to anywhere. Relatedly - you're missing the accessibility attributes referred to in the Figma.

eliykat (changes requested):

components/src/banner/banner.component.spec.ts
Hinton: I added two proper tests, reminds me to check my PRs before requesting reviews 🥲.

components/src/banner/banner.component.ts
Hinton: I think that's a fair argument.

Hinton:
Note for self. We should remove title and only provide the inner content.

eliykat (approved):
Code looks good, not sure why you've removed the title, but design can review that change.

PR #771, opened 2022-04-19
Remove userId from data models

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Remove unused `userId` fields from data models. It seems these got populated during sync for some reason. Since they are never used I've removed them.

## Before you submit

### DIFF
--- a/common/src/models/domain/cipher.ts
+++ b/common/src/models/domain/cipher.ts
@@ -49,13 +49,12 @@ export class Cipher extends Domain {
       obj,
       {
         id: null,
-        userId: null,
         organizationId: null,
         folderId: null,
         name: null,
         notes: null,
       },
-      ["id", "userId", "organizationId", "folderId"]
+      ["id", "organizationId", "folderId"]
     );
 
     this.type = obj.type;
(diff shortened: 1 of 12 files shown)


common/src/models/domain/cipher.ts
```diff
@@ -186,12 +185,11 @@ export class Cipher extends Domain {
     return model;
   }
 
-  toCipherData(userId: string): CipherData {
+  toCipherData(): CipherData {
     const c = new CipherData();
     c.id = this.id;
     c.organizationId = this.organizationId;
     c.folderId = this.folderId;
-    c.userId = this.organizationId != null ? userId : null;
```
cscharf, line 194: Are we sure this isn't used? Seems that we use `userId` vs. `organizationId` to determine the ownership of the vault item.

Hinton: It's only available in `CipherData`. Which makes it quite difficult for us to actually use it in our applications 😀.

cscharf (approved):

PR #773, opened 2022-04-19
Refactor Folder Domain

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Suggestion on how we can refactor our domain models to avoid weird chains (`Response -> Data -> Domain -> View`). This puts the `Domain` in the center with everything converting to/from it.

The new data model is represented by the following objects:

* Domain: Our domain model
* Data: Our long term local storage model
* Decrypted: Decrypted Domain
* Request/Response: Servers representation

Something I haven't gotten around to investigating is how we can actually store `Domain` in memory instead of converting back/from Data all the time.

I also deprecated `EncString.decrypt` and replaced it with `EncString.decryptWithCryptoService` which uses a local crypto service instead of the global container.

## Before you submit

### DIFF
--- a/common/src/abstractions/folder.service.ts
+++ b/common/src/abstractions/folder.service.ts
@@ -1,15 +1,15 @@
 import { FolderData } from "../models/data/folderData";
 import { Folder } from "../models/domain/folder";
 import { TreeNode } from "../models/domain/treeNode";
-import { FolderView } from "../models/view/folderView";
+import { FolderDecrypted } from "../models/view/folderDecrypted";
 
 export abstract class FolderService {
   clearCache: (userId?: string) => Promise<void>;
   get: (id: string) => Promise<Folder>;
   getAll: () => Promise<Folder[]>;
-  getAllDecrypted: () => Promise<FolderView[]>;
-  getAllNested: () => Promise<TreeNode<FolderView>[]>;
-  getNested: (id: string) => Promise<TreeNode<FolderView>>;
+  getAllDecrypted: () => Promise<FolderDecrypted[]>;
+  getAllNested: () => Promise<TreeNode<FolderDecrypted>[]>;
+  getNested: (id: string) => Promise<TreeNode<FolderDecrypted>>;
   saveWithServer: (folder: Folder) => Promise<any>;
   upsert: (folder: FolderData) => Promise<any>;
   replace: (folders: { [id: string]: FolderData }) => Promise<any>;
--- a/common/src/models/export/folderExport.ts
+++ b/common/src/models/export/folderExport.ts
@@ -1,6 +1,6 @@
 import { EncString } from "../domain/encString";
 import { Folder as FolderDomain } from "../domain/folder";
-import { FolderView } from "../view/folderView";
+import { FolderDecrypted } from "../view/folderDecrypted";
 
 export class FolderExport {
   static template(): FolderExport {
@@ -9,7 +9,7 @@ export class FolderExport {
     return req;
   }
 
-  static toView(req: FolderExport, view = new FolderView()) {
+  static toView(req: FolderExport, view = new FolderDecrypted()) {
     view.name = req.name;
     return view;
   }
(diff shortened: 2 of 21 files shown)


Hinton:
I'm still a bit unhappy about `FolderView.fromFolder`, but I prefer that over having `Domain` convert itself to `View` since that would make a circular dependency, `view -> domain -> view`.

common/src/abstractions/folder.service.ts
```diff
@@ -1,21 +1,19 @@
...
   clearCache: (userId?: string) => Promise<void>;
-  encrypt: (model: FolderView, key?: SymmetricCryptoKey) => Promise<Folder>;
   get: (id: string) => Promise<Folder>;
   getAll: () => Promise<Folder[]>;
   getAllDecrypted: () => Promise<FolderView[]>;
   getAllNested: () => Promise<TreeNode<FolderView>[]>;
   getNested: (id: string) => Promise<TreeNode<FolderView>>;
   saveWithServer: (folder: Folder) => Promise<any>;
-  upsert: (folder: FolderData | FolderData[]) => Promise<any>;
+  upsert: (folder: FolderData) => Promise<any>;
```
MGibson1, line 14: Are you planning on an `upsertMany` if a bulk operation is supported?

common/src/models/domain/encString.ts
```diff
@@ -96,11 +96,10 @@ export class EncString {
...
   }
 
+  /**
+   * @deprecated As of 2022-04-20, use {@link decryptWithCryptoService} instead
+   */
   async decrypt(orgId: string, key: SymmetricCryptoKey = null): Promise<string> {
-    if (this.decryptedValue != null) {
-      return this.decryptedValue;
-    }
-
```
MGibson1, line 103: I don't think it's a good idea to remove this guard.

Are you thinking that the object hierarchy makes multiple decrypts less likely?

common/src/models/domain/folder.ts
```diff
@@ -1,40 +1,7 @@
-import { FolderData } from "../data/folderData";
-import { FolderView } from "../view/folderView";
-
-import Domain from "./domainBase";
 import { EncString } from "./encString";
 
-export class Folder extends Domain {
+export interface Folder {
```
MGibson1, line 3: 🥳

common/src/models/export/folderExport.ts
```diff
@@ -14,7 +14,7 @@ export class FolderExport {
     return view;
   }
 
-  static toDomain(req: FolderExport, domain = new FolderDomain()) {
+  static toDomain(req: FolderExport, domain: FolderDomain = {} as FolderDomain) {
```
MGibson1, line 17: Is this due to the interface?

common/src/services/folder.service.ts
```diff
@@ -30,33 +30,25 @@ export class FolderService implements FolderServiceAbstraction {
     await this.stateService.setDecryptedFolders(null, { userId: userId });
   }
 
-  async encrypt(model: FolderView, key?: SymmetricCryptoKey): Promise<Folder> {
-    const folder = new Folder();
-    folder.id = model.id;
-    folder.name = await this.cryptoService.encrypt(model.name, key);
-    return folder;
+  async decrypt(model: Folder, key?: SymmetricCryptoKey): Promise<FolderView> {
```
MGibson1, line 33: I don't think this is used externally

MGibson1 (commented):
I have a few questions, but the primary concern is I don't yet understand what this will be cleaning up.

I **do** want a better definition of all of the models we have and a definitive hierarchy sounds great, but it didn't result in the code cleanup I expected.

common/src/models/domain/encString.ts
Hinton: It's called in `decryptWithCryptoService`. So the guard is still there just moved below the container service initialization.

common/src/models/export/folderExport.ts
Hinton: Yes

common/src/services/folder.service.ts
Hinton: I meant to remove that 😢

common/src/abstractions/folder.service.ts
Hinton: Yes, I think that's a cleaner pattern than allowing two different data models.

`upsert(folder) => upsertMany([folder])`

PR #774, opened 2022-04-19
Add in notAllowedValueAsync.validator.ts

justindbaur:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Code changes

- **angular/src/validators/notAllowedValueAsync.validator.ts:** A validator that will get a value from an async method and compare it to the control value and return an error if they are equal.

## Before you submit

### DIFF
--- a/angular/src/validators/notAllowedValueAsync.validator.ts
+++ b/angular/src/validators/notAllowedValueAsync.validator.ts
@@ -1,6 +1,6 @@
 import { AbstractControl, AsyncValidatorFn, ValidationErrors } from "@angular/forms";
 
-export function notAllowedValueAsync(valueGetter: () => Promise<string>, caseInsensitive: boolean = false): AsyncValidatorFn {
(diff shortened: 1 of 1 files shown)


angular/src/validators/notAllowedValueAsync.validator.ts
```diff
@@ -0,0 +1,21 @@
+import { AbstractControl, AsyncValidatorFn, ValidationErrors } from "@angular/forms";
+
+export function notAllowedValueAsync(
+  valueGetter: () => Promise<string>,
```
MGibson1, line 4: Do you think it's worth the type acrobatics to allow `string | Promise<string>`?

MGibson1 (approved):

angular/src/validators/notAllowedValueAsync.validator.ts
justindbaur: No, in that case you'd need to change the return value of `AsyncValidatorFn` to `AsyncValidatorFn | ValidatorFn` to and have to branch on which function to return based on the type. I think if this is needed in the future having a sync and async version of this validator would be for the best.

PR #775, opened 2022-04-22
[EC-159] [BEEEP] Remove factory providers in Angular DI

eliykat:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

We overuse factory providers in our services modules. (i.e. the `useFactory` pattern in `jslib-services.module.ts`)

I think this is undesirable because:
* there are more places to update (or forget to update) dependencies in the `providers` array - you have to update the `deps` array, the factory function definition, and the call to the class constructor. In general it's just more complicated to follow.
* you have to re-declare the entire provider with all its dependencies just to override a single arg in the class constructor
* you repeat certain definitions, like the `StateFactory` call, when you're using it in multiple places

I've gone through and refactored our provider definitions to use `useClass` wherever possible, use `InjectionToken`s more widely, and made some small related refactors in the process.

## Code changes

- **angular/src/services/jslib-services.module.ts**
  - extract common constructor args out to be injectionTokens, declare them as providers
  - change `useFactory` providers to `useClass`
  - replace existing string tokens with `InjectionToken`s
- **common/src/services/notifications.service.ts** and **common/src/services/vaultTimeout.service.ts** - make the callback interfaces consistent so that we can use the same token for them
- **electron/src/services/electronPlatformUtils.service.ts** - pass in the `clientType` directly instead of `isDesktop`, this just seems more straightforward

Client PRs will reference this one, see the list of referenced PRs below.

## Testing requirements

General regression testing on all Angular clients.

## Before you submit


electron/src/services/electronPlatformUtils.service.ts
```diff
@@ -11,17 +11,14 @@ import { ThemeType } from "jslib-common/enums/themeType";
...
 
 export class ElectronPlatformUtilsService implements PlatformUtilsService {
-  private clientType: ClientType;
   private deviceCache: DeviceType = null;
 
   constructor(
     protected i18nService: I18nService,
     private messagingService: MessagingService,
-    private isDesktopApp: boolean,
+    private clientType: ClientType.Desktop | ClientType.DirectoryConnector,
```
Hinton, line 19: 👍

angular/src/services/jslib-services.module.ts
```diff
@@ -82,20 +82,49 @@ import { PasswordRepromptService } from "./passwordReprompt.service";
...
+      provide: LOCALE_ID,
+      useFactory: (i18nService: I18nServiceAbstraction) => i18nService.translationLocale,
+      deps: [I18nServiceAbstraction],
+    },
+    {
+      provide: STATE_FACTORY,
+      useValue: new StateFactory(GlobalState, Account),
+    },
+    {
+      provide: USE_DISK_CACHE,
```
Hinton, line 114: Should we rename this? `Use_disk_cache` doesn't really convey that it is only used in the state service.

Hinton (approved):

Hinton (approved):

PR #786, opened 2022-04-29
[CL-23] Forms

Hinton:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Adds Form components.

## Before you submit


danielleflinn (dismissed):
Storybook looks good.

components/src/form-field/form-field.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    this.suffixChildren.forEach((suffix) => {
+      suffix.last = suffix == this.suffixChildren.last;
+    });
+
+    if (this.error) {
+      this.input.ariaDescribedBy = this.error.id;
+    } else {
+      this.input.ariaDescribedBy = undefined;
```
eliykat, line 44: It should be described by its Hint, if there is one.

I'm not sure how you reconcile this with also using `describedBy` for the error message. I suggest you put both the error and the hint in a generic `<div>` which is used for this purpose, so you capture both.

components/src/form-field/form-field.module.ts
```diff
@@ -0,0 +1,54 @@
...
+import { BitErrorSummary } from "./error-summary.component";
+import { BitError } from "./error.component";
+import { BitFormFieldComponent } from "./form-field.component";
+import { BitHintComponent } from "./hint.component";
+import { BitLabel } from "./label.directive";
+import { BitPrefix } from "./prefix.directive";
+import { BitSuffix } from "./suffix.directive";
+
+/**
+ * Temporarily duplicate this pipe
```
eliykat, line 18: What's going on here? Can't we just import the i18n pipe from jslib?

components/src/form-field/error.component.ts
```diff
@@ -0,0 +1,39 @@
...
+import { I18nService } from "jslib-common/abstractions/i18n.service";
+
+// Increments for each instance of the input component
+let nextId = 0;
+
+@Component({
+  selector: "bit-error",
+  template: `<i class="bwi bwi-error"></i> {{ displayError }}`,
+  host: {
+    "[attr.id]": "id",
```
eliykat, line 12: The Angular style guide recommends using `@HostBinding` over this: https://angular.io/guide/styleguide#hostlistenerhostbinding-decorators-versus-host-metadata

>Why?
>The property associated with @[HostBinding](https://angular.io/api/core/HostBinding) or the method associated with @[HostListener](https://angular.io/api/core/HostListener) can be modified only in a single place —in the directive's class. If you use the host metadata property, you must modify both the property/method declaration in the directive's class and the metadata in the decorator associated with the directive.

We've used `@HostBinding` in the other components, so I think we should be consistent in that.

components/src/form-field/error-summary.component.ts
```diff
@@ -0,0 +1,27 @@
...
+    class: "tw-block tw-text-danger tw-mt-2",
+    "aria-live": "assertive",
+  },
+})
+export class BitErrorSummary {
+  @Input()
+  formGroup: FormGroup;
+
+  get errorCount(): number {
+    const errors = Object.values(this.formGroup.controls).filter((e) => e.touched && !e.valid);
```
eliykat, line 19: What about nested formGroups? I think you need to recurse, like this: https://github.com/bitwarden/web/blob/master/bitwarden_license/src/app/organizations/manage/sso.component.ts#L245 (although your implementation is easier to understand)

components/src/form-field/error.component.ts
```diff
@@ -0,0 +1,39 @@
+import { Component, Input } from "@angular/core";
+
+import { I18nService } from "jslib-common/abstractions/i18n.service";
+
+// Increments for each instance of the input component
+let nextId = 0;
```
eliykat, line 6: I think this is fine, but the comment is misleading. It increments with each instance of this component, not the input component. Same for the pattern in other components.

components/src/form-field/error-summary.component.ts
```diff
@@ -0,0 +1,27 @@
+import { Component, Input } from "@angular/core";
```
eliykat, line 1: This count isn't working in the Storybook, is that just because we're mocking the i18nService?
Screen Shot 2022-05-27 at 2 43 43 pm

eliykat: Also, we need singular wording if only 1 field requires attention (so we don't have "1 fields")

components/src/form-field/form-field.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.prefixChildren.forEach((prefix) => {
+      prefix.first = prefix == this.prefixChildren.first;
+    });
+
+    this.suffixChildren.forEach((suffix) => {
+      suffix.last = suffix == this.suffixChildren.last;
+    });
+
+    if (this.error) {
+      this.input.ariaDescribedBy = this.error.id;
```
eliykat, line 42: We should also set `aria-invalid` on the input if there are errors.

components/src/form-field/error.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  get displayError() {
+    switch (this.error[0]) {
+      case "required":
+        return this.i18nService.t("inputRequired");
+      case "email":
+        return this.i18nService.t("inputEmail");
+      default:
+        if (this.error[1]?.message) {
+          return this.error[1]?.message;
```
eliykat, line 33: I don't quite understand what's going on here, is this for a custom error message? Can you provide a story that uses it?

components/src/form-field/suffix.directive.ts
```diff
@@ -0,0 +1,18 @@
+import { Directive, HostBinding, Input } from "@angular/core";
+
+import { PrefixClasses } from "./prefix.directive";
+
+@Directive({
+  selector: "[bitSuffix]",
```
eliykat, line 6: What do you think about making these selectors more specific, like `button[bitSuffix]`? That way they indicate their intended usage (i.e. "use this with a button element")

eliykat (changes requested):
Nice work! Can't wait to start using it :D

components/src/form-field/form-field.module.ts
Hinton: That would create a cyclic dependency. `jslib/angular -> components -> jslib/angular`. The proper solution would be to split up `jslib/angular` but I won't do that until we have migrated to the mono repository.

components/src/form-field/form-field.component.ts
Hinton: We only show a single value at the time. So I believe we can use `describedBy = hint` if there are no error.

components/src/form-field/error-summary.component.ts
Hinton: Yes, the mock doesn't handle substitutions. (Which I think is fine)

components/src/form-field/suffix.directive.ts
Hinton: We should probably also support non-buttons for text values.

components/src/form-field/error-summary.component.ts
Hinton: Good catch. I copied your code.

eliykat (approved):

PR #787, opened 2022-05-02
[CL-38] Follow Angular styleguide naming convention, simplify hostbinding logic

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Code changes

- **badge:**
  - Rename directive to `bitBadge`.
  - Rename file to `badge.directive`. Simplify hostbinding.
  - Update story to use same variable name for badge type to preserve select input in storybook.
- **button:**
  - Rename directive to `bitButton`.
  - Rename file to `button.directive`. Simplify hostbinding.

## Before you submit


components/src/button/button.directive.ts
```diff
@@ -74,8 +58,14 @@ export class ButtonComponent implements OnInit, OnChanges {
       "focus:tw-ring",
       "focus:tw-ring-offset-2",
       "focus:tw-ring-primary-700",
-      this.block ? "tw-w-full tw-block" : "tw-inline-block",
-      buttonStyles[this.buttonType ?? "secondary"],
-    ];
+    ]
+      .concat(this.block ? ["tw-w-full", "tw-block"] : ["tw-inline-block"])
+      .concat(buttonStyles[this.buttonType ?? "secondary"]);
```
eliykat, line 63: We're doubling up on setting the default value here - `buttonType` already has a default value, I don't think you need the null coalescing in this line.

eliykat (changes requested):

components/src/button/button.directive.ts
```diff
@@ -60,7 +60,7 @@ export class ButtonDirective {
       "focus:tw-ring-primary-700",
     ]
       .concat(this.block ? ["tw-w-full", "tw-block"] : ["tw-inline-block"])
-      .concat(buttonStyles[this.buttonType ?? "secondary"]);
+      .concat(buttonStyles[this.buttonType] ?? []);
```
eliykat, line 63: But will `this.buttonType` ever be null if it has a default value? Can't we just do

```suggestion
      .concat(buttonStyles[this.buttonType]);
```

?

If I'm missing something and we do need to account for `this.buttonType` being null, then your original code would be better - my query was whether we need the `??` at all.

Hinton: I think so. `[].concat(undefined) => [undefined]`. The new code is logically better since it ensures we don't keep `undefined` in the array.

eliykat (dismissed):

gbubemismith (approved):
looks good.

components/src/badge/badge.directive.ts
```diff
@@ -21,28 +21,10 @@ const hoverStyles: Record<BadgeTypes, string[]> = {
...
-    this.classList = this.classes.join(" ");
-  }
-
-  ngOnChanges() {
-    this.ngOnInit();
-  }
-
-  get classes() {
+export class BadgeDirective {
+  @HostBinding("class") get classList() {
```
coroiu, line 27: praise: nice, that's a pretty significant simplification!

coroiu (approved):
Looks like some nice simplification and then just a bunch of renaming, lgtm
Haven't worked with storybook though so that part I wasn't able to review
Also I haven't been able to run the code to test because I don't have the dev env up yet :D

PR #791, opened 2022-05-05
[SG-225] [EUVR] Remove BaseGuard

eliykat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

When doing the refresh work, we had competing refactors:
* in the feature branch, I created a `baseGuard` class to simplify the redirect pattern we use in our routing guards (calling a redirect and then returning `false`)
* on master, Oscar refactored routing guards to also simplify this pattern (returning `this.router.createUrlTree` directly)

We should adopt the changes on master and drop `baseGuard`, which is no longer needed.

## Code changes

* remove `baseGuard`
* make sure all guards go back to using `this.router.createUrlTree` instead of `super.redirect()`

## Before you submit


Hinton (approved):
This looks to match the changes I made to the guards in another PR!

PR #802, opened 2022-05-12
[PS-250] Prevent timeouts less than 1 minute

gbubemismith:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Web Vault should not allow the user to select a value less than 1 minute in the custom timeout inputs.

## Code changes
   - **angular/src/components/settings/vault-timeout-input.component.ts:** Added change detection on the custom minutes field to know when the inout is valid or not and then update the validMinutes variable appropraitely.
 

https://user-images.githubusercontent.com/13024008/170218090-a5d5784d-e47f-460c-8220-bc671d0c5b17.mov

## Before you submit


eliykat:
I'm not sure this will have the intended effect - I left some comments in the Jira ticket for discussion.

angular/src/components/settings/vault-timeout-input.component.ts
```diff
@@ -19,6 +19,8 @@ export class VaultTimeoutInputComponent implements ControlValueAccessor, Validat
   }
 
   static CUSTOM_VALUE = -100;
+  static MIN_CUSTOM_MINUTES = 1;
+  validMinutes = true;
```
Hinton, line 23: This is a custom input component which implements `ControlValueAccessor`. For validation purposes you can use the `validate` function below.

angular/src/components/settings/vault-timeout-input.component.ts
```diff
@@ -19,6 +19,8 @@ export class VaultTimeoutInputComponent implements ControlValueAccessor, Validat
   }
 
   static CUSTOM_VALUE = -100;
+  static MIN_CUSTOM_MINUTES = 1;
```
Hinton, line 22: Should the value be 0 on some clients or do we default to not handling `00:00` in the custom timeout option?

Hinton (changes requested):
This looks like it requires client changes to function correctly, any PRs for that yet (to graph the full picture)

gbubemismith:
> This looks like it requires client changes to function correctly, any PRs for that yet (to graph the full picture)

yeah it has. The PR is in drafts but I'm opening it now.

angular/src/components/settings/vault-timeout-input.component.ts
gbubemismith: On web the expectation is to not allow 00:00. The minimum timeout should be a minute.

angular/src/components/settings/vault-timeout-input.component.ts
gbubemismith: okay I would effect the change

gbubemismith:
> This looks like it requires client changes to function correctly, any PRs for that yet (to graph the full picture)

@Hinton  https://github.com/bitwarden/web/pull/1675

angular/src/components/settings/vault-timeout-input.component.ts
```diff
@@ -132,6 +134,10 @@ export class VaultTimeoutInputComponent implements ControlValueAccessor, Validat
       return { policyError: true };
     }
 
+    //Check minimum allowed custom minutes
+    this.validMinutes =
+      control.value < VaultTimeoutInputComponent.MIN_CUSTOM_MINUTES ? false : true;
```
Hinton, line 139: We don't need to assign this to a variable, simply return an error and handle the error in the form validation logic in the clients.

angular/src/components/settings/vault-timeout-input.component.ts
```diff
@@ -132,6 +133,10 @@ export class VaultTimeoutInputComponent implements ControlValueAccessor, Validat
       return { policyError: true };
     }
 
+    if (control.value < VaultTimeoutInputComponent.MIN_CUSTOM_MINUTES) {
```
eliykat, line 136: You also want to check that the user has selected a custom value. e.g. The "on refresh" option is represented by `-1`: https://github.com/bitwarden/web/blob/master/src/app/settings/preferences.component.ts#L46. Desktop also has some nonstandard options that are negative values. They are special cases and we still want to allow them.

A custom value is indicated by `this.form.vaultTimeout` having the `CUSTOM_VALUE` value.

eliykat (changes requested):

eliykat:
Making this change in jslib will make it apply to all clients. If we limit it to custom inputs only, I think that's fine, because then custom inputs will behave consistently, but options in the dropdown list (which differ between clients) will always be OK. However, this means you'll need to submit a PR for the `clients` repo as well, to update error messages across all clients.

angular/src/components/settings/vault-timeout-input.component.ts
eliykat: Actually, this `validate` method applies at the form level and is only exposed to the parent component. You need access to this validation state within this component so that you can show the inline error message. I suggest you extract this logic to a new method, that you can call within the component (to decide whether to show the in-line validation error) and also from this method (to decide whether the form as a whole is valid).

angular/src/components/settings/vault-timeout-input.component.ts
```diff
@@ -19,6 +19,7 @@ export class VaultTimeoutInputComponent implements ControlValueAccessor, Validat
   }
 
   static CUSTOM_VALUE = -100;
+  static MIN_CUSTOM_MINUTES = 1;
```
eliykat, line 22: The default custom timeout value for me is 0 hours and 0 minutes, which makes the error message appear immediately. I think it should default to the minimum allowed time.

gbubemismith: with the new implementation for the web vault the minimum allowed time should be `1 minute`

angular/src/components/settings/vault-timeout-input.component.ts
```diff
@@ -132,10 +136,18 @@ export class VaultTimeoutInputComponent implements ControlValueAccessor, Validat
...
+
     return null;
   }
 
   registerOnValidatorChange(fn: () => void): void {
     this.validatorChange = fn;
   }
+
+  getValidMinutesInput(): boolean {
+    return this.showCustom && this.customTimeInMinutes > 0;
```
eliykat, line 151: There's an error in this boolean logic - it'll return `false` (indicating an error) if a custom time out is not selected. 

You could go
```typescript
return !this.showCustom || this.customTimeInMinutes > 0
```

But I think breaking it out into a ternary if condition might be a clearer way of expressing it:

```suggestion
    return this.showCustom
      ? this.customTimeInMinutes > 0
      : true;
```

angular/src/components/settings/vault-timeout-input.component.ts
```diff
@@ -132,10 +136,18 @@ export class VaultTimeoutInputComponent implements ControlValueAccessor, Validat
...
+    }
+
     return null;
   }
 
   registerOnValidatorChange(fn: () => void): void {
     this.validatorChange = fn;
   }
+
+  getValidMinutesInput(): boolean {
```
eliykat, line 150: I don't think this name is very clear, maybe use a proper getter and then call it something like:

```suggestion
  get exceedsMinimumTimeout(): boolean {
```

angular/src/components/settings/vault-timeout-input.component.ts
eliykat: I also liked that your previous implementation didn't have a magic number here. I recommend declaring a const outside the class to hold the minimum timeout value.

eliykat (changes requested):

angular/src/components/settings/vault-timeout-input.component.ts
gbubemismith: I think the condition is clear

eliykat (dismissed):

angular/src/components/settings/vault-timeout-input.component.ts
```diff
@@ -18,7 +18,18 @@ export class VaultTimeoutInputComponent implements ControlValueAccessor, Validat
...
 
+  get customTimeInMinutes() {
+    return this.form.get("custom.hours")?.value * 60 + this.form.get("custom.minutes")?.value;
+  }
+
+  get exceedsMinimumTimout(): boolean {
+    return (
+      !this.showCustom || this.customTimeInMinutes > VaultTimeoutInputComponent.MIN_CUSTOM_MINUTES
+    );
+  }
```
Hinton, line 29: Should we make these private? They are only used internally right?

gbubemismith: `exceedsMinimumTimout` is used in a view but I would make `customTimeInMinutes` private

Hinton (approved):

PR #804, opened 2022-05-16
[ps-136] Ignore accented characters in vault search

gbubemismith:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Ignore accented characters in vault searches. 

## Code changes
- **common/src/services/search.service.ts:** Customised lunr text processing by adding a pipeline function `normalizeAccentsPipelineFunction` to strip off the accented characters during search. The pipeline function is registered and the added to the pipeline.

## Before you submit


jlf0dev (dismissed):

common/src/services/search.service.ts
```diff
@@ -51,7 +51,10 @@ export class SearchService implements SearchServiceAbstraction {
     const builder = new lunr.Builder();
     builder.ref("id");
     builder.field("shortid", { boost: 100, extractor: (c: CipherView) => c.id.substr(0, 8) });
-    builder.field("name", { boost: 10 });
+    builder.field("name", {
+      boost: 10,
+      extractor: (c: CipherView) => c.name.normalize("NFD").replace(/[\u0300-\u036f]/g, ""),
```
MGibson1, line 56: Shouldn't we add the same extractor for most user-input fields?
* subtitle
* username
* notes
?

Hinton: Do we also need to also strip out the characters from the search input as well? With this change searching for `náme` would no longer match `nàme` since it would be normalized as `name`.

gbubemismith: @MGibson1 yeah I would add them to those fields. @Hinton the accented characters are currently being striped off from the input field.

Hinton:
We could look into extending Lunr to be unicode safe per https://github.com/olivernn/lunr.js/issues/269. The last gist is compatible with 2 but not something I've used.

gbubemismith:
> We could look into extending Lunr to be unicode safe per [olivernn/lunr.js#269](https://github.com/olivernn/lunr.js/issues/269). The last gist is compatible with 2 but not something I've used.

I would look into this right away

gbubemismith:
> We could look into extending Lunr to be unicode safe per [olivernn/lunr.js#269](https://github.com/olivernn/lunr.js/issues/269). The last gist is compatible with 2 but not something I've used.

@Hinton I was able to customise lunr and added a pipeline function to handle accented characters. This is a cleaner approach as we follow DRY principle.

PR #806, opened 2022-05-17
[SG-295] Use admin collections endpoint for org vault

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix SG-295: after creating a new collection, you have to refresh the page before it will appear in the vault filters on the organization page.

This is because, pre-End User Vault Refresh, users with `canEditAnyCollection` permissions got their collections directly from the server, rather than synced data: https://github.com/bitwarden/web/blob/v2.28.1/src/app/organizations/vault/groupings.component.ts#L39. Best I can tell, this unintentionally got dropped during EUVR, so it's still pulling from outdated sync data.

Aside from that, we also just need to make sure that admin users are getting the right data.

@addisonbeck and @differsthecat can you please look at this as you are most familiar with filters.

**This ~will~ _may_ be picked to rc**

## Code changes

* **angular/src/modules/vault-filter/models/vault-filter.model.ts** - add a `useAdminCollections` flag to `VaultFilter`. This is because:
  * conceptually, this model is a record of what ciphers/collections/folders to show the user, so it fits
  * practically, this model is the main throughline of communication from the Vault component to the `VaultFilterService` that fetches the collections (which is several layers deep)
  * we can't just pass the whole `organization` object, because if the user is applying an org filter in their personal vault, you don't want to show all collections, even if they technically have permissions to access them. We should only fetch all permissions on the org vault page.
* **angular/src/modules/vault-filter/vault-filter.service.ts** - replicate the old code which hits the `collections` endpoint, and switch based on the `useAdminCollections` flag

* **angular/src/modules/vault-filter/vault-filter.component.ts** - I encountered an issue where there were duplicate calls to `buildCollections`, which is more significant now that it involves async network calls. 
  * Turns out, each vault component already calls this, so we can remove it from `VaultFilterComponent.ngOnInit()`.
    * organizationVaultComponent calls it here: https://github.com/bitwarden/web/blob/fix/filters-new-collection/src/app/modules/vault/modules/organization-vault/organization-vault.component.ts#L113
    * individualVaultComponent calls it here: https://github.com/bitwarden/web/blob/fix/filters-new-collection/src/app/modules/vault/modules/individual-vault/individual-vault.component.ts#L109
    * desktop still needs this call, but that can be pushed down into its subclass
    * browser does not subclass this at all and is unaffected by this change
  * now, `initCollections` is not called by anything, so it can be removed.

## Before you submit

### DIFF
--- a/angular/src/modules/vault-filter/models/vault-filter.model.ts
+++ b/angular/src/modules/vault-filter/models/vault-filter.model.ts
@@ -11,6 +11,7 @@ export class VaultFilter {
   selectedOrganizationId?: string;
   myVaultOnly = false;
   refreshCollectionsAndFolders = false;
+  useAdminCollections = false;
 
   constructor(init?: Partial<VaultFilter>) {
     Object.assign(this, init);
--- a/angular/src/modules/vault-filter/vault-filter.component.ts
+++ b/angular/src/modules/vault-filter/vault-filter.component.ts
@@ -72,7 +72,7 @@ export class VaultFilterComponent implements OnInit {
     this.folders = await this.vaultFilterService.buildFolders(filter.selectedOrganizationId);
     this.collections = filter.myVaultOnly
       ? null
-      : await this.vaultFilterService.buildCollections(filter.selectedOrganizationId);
+      : await this.vaultFilterService.buildCollections(filter);
   }
 
   async reloadOrganizations() {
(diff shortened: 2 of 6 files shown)


differsthecat (approved):

eliykat:
Updated description per additional changes

eliykat:
Closing in favour of @addisonbeck's work here: https://github.com/bitwarden/web/pull/1700

PR #811, opened 2022-05-20
[PS-526] Allow loading captcha cookie on desktop

jlf0dev:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

hCaptcha allows users to load an accessibility cookie into their browser in order to bypass the captcha. This currently works for our web, browser, and mobile clients, but we haven't had a way to load cookies into desktop in the past.

This PR adds a component that allows users to paste in their accessibility cookie url from hCaptcha. This url is then opened in a new window and cookies are added to the process. When the user retries their hCaptcha, it is then able to see the accessibility cookie.

## Code changes

- **angular/src/scss/bwicons/\*:** Add new `universal-access` icon to our icon set
- **electron/src/services/electronMainMessaging.service.ts:** Add a message handler on the main process that checks for the specified cookies
- **electron/src/utils.ts:** Add a utility function that allows the renderer process to check for specified cookies. Returns a list of matching cookies.

## Before you submit


eliykat (dismissed):

eliykat:
Linter is sad

Hinton (approved):

djsmith85:
Merging this now, to be included in the mono-repo move. I don't see any breaking changes/side effects with clients.

PR #812, opened 2022-05-20
[PS-136] stripped off accented characters from search field for browser client

gbubemismith:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
- Replicate same functionality of being able to search for texts with accented characters using normal texts. 

## Code changes
- **angular/src/components/ciphers.component.ts:** Added `normalize("NFD").replace(/[\u0300-\u036f]/g, "")` to the `searchText` variable to strip off accent from a character. 

- This PR is related to https://github.com/bitwarden/web/pull/1690 and https://github.com/bitwarden/jslib/pull/804

## Before you submit

### DIFF
--- a/angular/src/components/ciphers.component.ts
+++ b/angular/src/components/ciphers.component.ts
@@ -77,14 +77,19 @@ export class CiphersComponent {
   }
 
   isSearching() {
-    return !this.searchPending && this.searchService.isSearchable(this.searchText);
+    return (
+      !this.searchPending &&
+      this.searchService.isSearchable(
+        this.searchText?.normalize("NFD").replace(/[\u0300-\u036f]/g, "")
+      )
+    );
   }
 
   protected deletedFilter: (cipher: CipherView) => boolean = (c) => c.isDeleted === this.deleted;
 
   protected async doSearch(indexedCiphers?: CipherView[]) {
     this.ciphers = await this.searchService.searchCiphers(
-      this.searchText,
+      this.searchText?.normalize("NFD").replace(/[\u0300-\u036f]/g, ""),
       [this.filter, this.deletedFilter],
       indexedCiphers
     );
--- a/angular/src/components/ciphers.component.ts
+++ b/angular/src/components/ciphers.component.ts
@@ -77,19 +77,14 @@ export class CiphersComponent {
   }
 
   isSearching() {
-    return (
-      !this.searchPending &&
-      this.searchService.isSearchable(
-        this.searchText?.normalize("NFD").replace(/[\u0300-\u036f]/g, "")
-      )
-    );
+    return !this.searchPending && this.searchService.isSearchable(this.searchText);
   }
 
   protected deletedFilter: (cipher: CipherView) => boolean = (c) => c.isDeleted === this.deleted;
 
   protected async doSearch(indexedCiphers?: CipherView[]) {
     this.ciphers = await this.searchService.searchCiphers(
-      this.searchText?.normalize("NFD").replace(/[\u0300-\u036f]/g, ""),
+      this.searchText,
       [this.filter, this.deletedFilter],
       indexedCiphers
     );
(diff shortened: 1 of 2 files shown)


eliykat (commented):
Is there any reason why we're not normalizing the search string in `SearchService` itself? Ideally this should be handled by the service, so that you don't have to repeat the normalization logic everywhere that it's called from.

Also, don't forget to handle searching Sends as well. If you do it in `SearchService`, then you could have a single private method, `normalizeSearchQuery`, which is called from both `SearchCiphers` and `SearchSends`.

gbubemismith:
> Is there any reason why we're not normalizing the search string in `SearchService` itself? Ideally this should be handled by the service, so that you don't have to repeat the normalization logic everywhere that it's called from.
> 
> Also, don't forget to handle searching Sends as well. If you do it in `SearchService`, then you could have a single private method, `normalizeSearchQuery`, which is called from both `SearchCiphers` and `SearchSends`.

This has been added.

angular/src/components/ciphers.component.ts
```diff
@@ -77,14 +77,19 @@ export class CiphersComponent {
...
+      )
+    );
   }
 
   protected deletedFilter: (cipher: CipherView) => boolean = (c) => c.isDeleted === this.deleted;
 
   protected async doSearch(indexedCiphers?: CipherView[]) {
     this.ciphers = await this.searchService.searchCiphers(
-      this.searchText,
+      this.searchText?.normalize("NFD").replace(/[\u0300-\u036f]/g, ""),
```
Hinton, line 92: This is no longer needed since it's done in the service instead.

angular/src/components/ciphers.component.ts
```diff
@@ -77,14 +77,19 @@ export class CiphersComponent {
   }
 
   isSearching() {
-    return !this.searchPending && this.searchService.isSearchable(this.searchText);
+    return (
+      !this.searchPending &&
+      this.searchService.isSearchable(
+        this.searchText?.normalize("NFD").replace(/[\u0300-\u036f]/g, "")
```
Hinton, line 83: This is no longer needed since it's done in the service instead.

eliykat (approved):
Looks good to me, thanks. Please also submit a PR to revert https://github.com/bitwarden/web/pull/1690 now that it's no longer needed.

common/src/services/search.service.ts
```diff
@@ -294,12 +295,13 @@ export class SearchService implements SearchServiceAbstraction {
...
-        .normalize("NFD")
-        .replace(/[\u0300-\u036f]/g, "");
+      return this.normalizeSearchQuery(token.toString());
     }
 
     return token;
   }
+
+  private normalizeSearchQuery(query: string): string {
+    return query?.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
```
eliykat, line 305: On second thoughts - please add a comment here explaining what this does, regexes with unicode ranges are not very self-explanatory. (e.g. "Normalize unicode string to its decomposed form and then remove all accents. This unicode range is equivalent to the Diacritic unicode property escape, but is backwards-compatible with older browsers.")

common/src/services/search.service.ts
```diff
@@ -294,12 +295,14 @@ export class SearchService implements SearchServiceAbstraction {
...
-        .toString()
-        .normalize("NFD")
-        .replace(/[\u0300-\u036f]/g, "");
+      return this.normalizeSearchQuery(token.toString());
     }
 
     return token;
   }
+
+  //Normalize diacritics characters from text
```
eliykat, line 304:
```suggestion
  // Remove accents/diacritics characters from text. This regex is equivalent to the Diacritic unicode property escape, i.e. it will match all diacritic characters.
```

eliykat: I wanted to explain the regex in particular

eliykat (approved):

PR #813, opened 2022-05-23
[PS-153] upgrade premium on refresh

gbubemismith:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Grant access to premium features immediately after upgrade without having to log out.

## Code changes
- **common/src/abstractions/state.service.ts:** abstract method `setCanAccessPremium` added.
- **common/src/services/state.service.ts:** implementation of the abstract method `setCanAccessPremium` with sets the current state of `hasPremiumPersonally` property when premium has been activated.
- **common/src/services/sync.service.ts:** calls the `setCanAccessPremium` method which syncs and sets the current state of `hasPremiumPersonally` property.

https://user-images.githubusercontent.com/13024008/169881648-b03aa6c3-baab-4f58-a036-0973668e1fe0.mov

## Before you submit

### DIFF
--- a/common/src/abstractions/state.service.ts
+++ b/common/src/abstractions/state.service.ts
@@ -58,6 +58,7 @@ export abstract class StateService<T extends Account = Account> {
   getBiometricUnlock: (options?: StorageOptions) => Promise<boolean>;
   setBiometricUnlock: (value: boolean, options?: StorageOptions) => Promise<void>;
   getCanAccessPremium: (options?: StorageOptions) => Promise<boolean>;
+  setCanAccessPremium: (value: boolean, options?: StorageOptions) => Promise<void>;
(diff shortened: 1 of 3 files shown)


differsthecat:
Nice solution, but just to confirm if a user loses premium one way or another, `syncProfile` will set it back to false?

gbubemismith:
> Nice solution, but just to confirm if a user loses premium one way or another, `syncProfile` will set it back to false?

Yes the state would be synced

differsthecat:
> > Nice solution, but just to confirm if a user loses premium one way or another, `syncProfile` will set it back to false?
> 
> Yes the state would be synced

Got it, just wanted to make sure!

differsthecat (approved):

PR #816, opened 2022-05-25
[EC-178] Logout on Key Connector error

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Related to https://github.com/bitwarden/key-connector/pull/16.

If there is any error with Key Connector (including contacting Key Connector or an unhandled server error), the user remains logged in. This presents them with a lock screen or their vault, depending on the flow. Because they have a token and a locally generated user key, they can even create new ciphers etc.

I propose that if there is any error with Key Connector onboarding or fetching user keys, this is most likely unrecoverable and we should just log the user out.

## Code changes

- **common/src/services/keyConnector.service.ts** - add common error handling code - log the actual error to console, then log the user out and throw a custom error message for the toast.
- **angular/src/services/jslib-services.module.ts** - update dependencies
- **angular/src/components/sso.component.ts** - minor change to add TODO and update the error string. Instead of "Unable to reach Key Connector", we'll use a more generic "Key Connector error. Make sure Key Connector is available and working correctly." so that we're not misleading the user about the nature of the error.

Video of the error flow:

https://user-images.githubusercontent.com/31796059/170176356-ecb44e92-90c0-4d11-b591-6f181299ab2b.mov

## Before you submit

### DIFF
--- a/angular/src/components/sso.component.ts
+++ b/angular/src/components/sso.component.ts
@@ -218,11 +218,13 @@ export class SsoComponent {
       }
     } catch (e) {
       this.logService.error(e);
-      if (e.message === "Unable to reach key connector") {
+
+      // TODO: Key Connector Service should pass this error message to the logout callback instead of displaying here
+      if (e.message === "Key Connector error") {
         this.platformUtilsService.showToast(
           "error",
           null,
-          this.i18nService.t("ssoKeyConnectorUnavailable")
+          this.i18nService.t("ssoKeyConnectorError")
         );
       }
     }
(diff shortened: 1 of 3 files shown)


angular/src/components/sso.component.ts
```diff
@@ -218,11 +218,13 @@ export class SsoComponent {
       }
     } catch (e) {
       this.logService.error(e);
-      if (e.message === "Unable to reach key connector") {
+
+      // TODO: Key Connector Service should pass this error message to the logout callback instead of displaying here
```
kspearrin, line 222: Todo still?

kspearrin (approved):

PR #817, opened 2022-05-25
[ps-136] Moved normalizeSearchQuery method to Utils

gbubemismith:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Code changes
- **common/src/services/search.service.ts:** referenced utils class and changed normalizeSearchQuery call to use the utils method.

- **common/src/misc/utils.ts:** moved `normalizeSearchQuery` here so it can be accessed by `lunr` pipeline function.

 

## Before you submit

### DIFF
--- a/angular/src/components/ciphers.component.ts
+++ b/angular/src/components/ciphers.component.ts
@@ -77,14 +77,19 @@ export class CiphersComponent {
   }
 
   isSearching() {
-    return !this.searchPending && this.searchService.isSearchable(this.searchText);
+    return (
+      !this.searchPending &&
+      this.searchService.isSearchable(
+        this.searchText?.normalize("NFD").replace(/[\u0300-\u036f]/g, "")
+      )
+    );
   }
 
   protected deletedFilter: (cipher: CipherView) => boolean = (c) => c.isDeleted === this.deleted;
 
   protected async doSearch(indexedCiphers?: CipherView[]) {
     this.ciphers = await this.searchService.searchCiphers(
-      this.searchText,
+      this.searchText?.normalize("NFD").replace(/[\u0300-\u036f]/g, ""),
       [this.filter, this.deletedFilter],
       indexedCiphers
     );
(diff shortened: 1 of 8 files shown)


Hinton:
Wasn't this already merged in #812?

Hinton:
We could just a private internal function in the file and use it?

gbubemismith:
> 

do you mean like a private normalizeQuery function and then call the function with `this.normalizeQuery` in the pipeline function???

Hinton:
Yes, just moving `normalizeQuery` to be a top level function. That should keep it accessible in the lunr pipeline (or so I assume).

gbubemismith:
> Yes, just moving `normalizeQuery` to be a top level function. That should keep it accessible in the lunr pipeline (or so I assume).

Moved it to the utils class

eliykat (commented):
`Utils` is a bit of a miscellaneous class, we shouldn't add to it if we can avoid it.

I would recommend just making `normalizeSearchQuery` a public method on `SearchService`, will that work for lunr? Or if it needs to be static, make it static on `SearchService`. You can do either without moving it to `Utils`.

@Hinton I don't quite understand your feedback, is this what you were suggesting?

gbubemismith:
> `Utils` is a bit of a miscellaneous class, we shouldn't add to it if we can avoid it.
> 
> I would recommend just making `normalizeSearchQuery` a public method on `SearchService`, will that work for lunr? Or if it needs to be static, make it static on `SearchService`. You can do either without moving it to `Utils`.
> 
> @Hinton I don't quite understand your feedback, is this what you were suggesting?

made it a static method in `SearchService`

Hinton:
@eliykat I'm fine with static on SearchService. My initial thought was to create a function outside the class and call it. Javascript isn't locked to OOO and we can diverge if needed.

differsthecat (approved):
This looks good 👍

